PR #21, opened 2017-07-23
Improve Perlin noise implementation

SilentSpike:
- Optimise the gradient function
- Perform a bitwise AND operation instead of a modulo
- Comment code more clearly
- Store all the functions in the perlin table


Valansch:
Working and looking great 👍

PR #37, opened 2017-08-12
Grilledham map gen rewrite

grilledham:
Moved my map gen code into gen_combined.

Can now choose which map is used by uncommenting line in map_layout.lua.
 
Patterns and shapes now behave the same which means the following:
  - A pattern doesn’t have to be used if you don’t want one. 
  - Patterns can be transformed like shapes.
  - Patterns can contain other patterns.
  - Opens up the possibility to add more pattern types.

Added support for colour images.


Valansch:
Next time use the develop as a base please. Idk if i could have changed it afterwards, its already merged. Doesnt matter for now^^

PR #39, opened 2017-08-14
Uncommented line in map_layout

grilledham:


Valansch:
Why are the files called test? Are they actually just tests? If not why are they called test?

grilledham:
They are just maps I use to try out new features. I've removed them from the master branch.

PR #40, opened 2017-08-18
Develop grilledham

grilledham:
Redlabel asked if it was possible to teleport to a ‘blurry’ part of the map without using coordinates. 

I added a command called tpmode. You call the command to switch on tpmode. When on if you place a ghost entity the player is teleported there and the ghost is destroyed.


grilledham:
Thanks for the git rebase tip.

I didn't know about desync related to changes to variables outside the global table. That's good to know and I've made the change in the code.

I uncommented the code for adding and removing the event handler at run time, with the intention that it would get added back when events adding/ removing becomes desync safe.

Valansch (approved):
I will merge into develop now. Did you test it in MP for desyncs yet? Once its tested i merge it into master

grilledham:
I have not tested in MP yet.

PR #52, opened 2017-09-08
Moved pokemessages into a seperate file and added more.

grilledham:


locale/resources/regulars.lua
```diff
@@ -0,0 +1,65 @@
...
+    zacman0510 = "",
+    TheOrangeAngle = "",
+    williambellwisdo = "",
+    Extraxyz = "",
+    markupolionCZ = "",
+    BJHunter = "",
+    ZTX = "",
+    bigglesthepirate = "",
+    ["-slug-"] = "",
+    ["g-h-o-s-t"] = "",
```
Valansch: Did you try this? I'm pretty sure that this will not work. But I could be wrong.

grilledham: image

Valansch: TIL

Valansch:
Thank you very much

Issue #19, opened 2017-07-23
implement sprinkled ores


Valansch:
Have impurities in ores

Valansch:
@TWLTriston Great!

PR #58, opened 2017-09-20
Glitter Ores Map Generation

TWLTriston:
Randomizes the ore patches outside of spawn area to density settings

### DIFF
--- a/locale/gen_ores/glitter_ores.lua
+++ b/locale/gen_ores/glitter_ores.lua
@@ -25,7 +25,6 @@ function run_ores_module_setup()
          ore_mix[ore_mix_max] = a
       end
    end
-    
 end
 
 run_ores_module_setup()
--- a/locale/gen_ores/glitter_ores.lua
+++ b/locale/gen_ores/glitter_ores.lua
@@ -10,6 +10,8 @@ local ore_ratios = {
    ["copper-ore"] = 1,
    ["stone"] = 0.25
 }
+-- 1-100% chance of sprinkling any individual ore
+local sprinkle_factor = 100
 
 -- Sets the buffer distance before ores are scrambled
 local starting_buffer = 100
@@ -51,20 +53,28 @@ function run_ores_module(event)
 
        for _, entity in pairs(entities) do
          if ore_ratios[entity.name] > 0 then 
-            
-            local new_name = nil
-            
-            --- Use the ratios to randomly select an ore
-            new_ore_random = math.random(1,mix_current)
-            new_name = ore_mix[new_ore_random]
-   --         game.print(new_name)
+            -- Roll to sprinkle_factor
+            if sprinkle_factor < 100 then
+               sprinkle_random = math.random(1,100)
+               should_sprinkle = sprinkle_random <= sprinkle_factor
+            else 
+               should_sprinkle = true
+            end
+            if should_sprinkle then
+               local new_name = nil
+               
+               --- Use the ratios to randomly select an ore
+               new_ore_random = math.random(1,mix_current)
+               new_name = ore_mix[new_ore_random]
+      --         game.print(new_name)
 
-            local position_old = entity.position
-            local amount_old = entity.amount
+               local position_old = entity.position
+               local amount_old = entity.amount
 
-            local surface = entity.surface
-            entity.destroy()
-            local new_entity = surface.create_entity{name = new_name, position = position_old, force="neutral", amount=amount_old}
+               local surface = entity.surface
+               entity.destroy()
+               local new_entity = surface.create_entity{name = new_name, position = position_old, force="neutral", amount=amount_old}
+            end
          end   
       end
     end
--- a/locale/gen_ores/glitter_ores.lua
+++ b/locale/gen_ores/glitter_ores.lua
@@ -2,7 +2,7 @@
 glitter_debug = false
 
 function run_ores_module_setup()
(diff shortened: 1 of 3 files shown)


TWLTriston:
+1 sprinkle, addresses issue #19

locale/gen_ores/glitter_ores.lua
```diff
@@ -0,0 +1,81 @@
...
+         mix_current = mix_current + 1
+         ore_mix[mix_current] = a
+      end
+   end
+      
+end
+
+--generate ores for entire chunk
+function run_ores_module(event)
+   run_ores_module_setup()
```
Valansch: Why do you call the setup on every chunk load? This means doing the loop line 23 to 28 5000 times :O on every chunk generation.

TWLTriston: I found that without the setup the savedscenarios didn't trigger the array initialization
i had a check in place for global.glitter_setup = false

But that gets set to true, and upon reloading the saved scenario, the ores weren't generating

Maybe another way to check for initialization, as globals persist but locals do not?

locale/gen_ores/glitter_ores.lua
```diff
@@ -0,0 +1,81 @@
...
+-- Sets the buffer distance before ores are scrambled
+local starting_buffer = 100
+
+function run_ores_module_setup()
+	if glitter_debug then game.print("Glitter Ore: debug is enabled; module setup;") end
+
+   -- Prime the array
+   for a,b in pairs(ore_ratios) do
+      for i=1,(b*1000) do
+         mix_current = mix_current + 1
```
Valansch: Mix_current will cause desyncs here. 
The following scenario:
- Server starts
- Server generates first chunk on tick 0 and set mix_current to some high number n.
- User joins game on tick 50. He loads the script and sets mix_current to 1 on line 6.
- 2nd chunk is generated on tick 64, both run this line.
- The clients result is 1+1 = 2, while the servers result is n + 1. The two have conflicting states and the client desyncs from the game.

locale/gen_ores/glitter_ores.lua
Valansch: Just run it once and store the data in a global table. That will solve both the desync and init problem.

locale/gen_ores/glitter_ores.lua
TWLTriston: Ok --- suggestion on how to build the array without desync?

Goal is to have an upper and lower bound to compare the random number against -- depending on outcome, was looking for an O[1] solution, but given we are only 4 elements really max, doing it O[n] may not be so bad.

Thoughts?

TWLTriston:
Additional fixes published, please review again

map_layout.lua
```diff
@@ -34,13 +34,13 @@ in this file and your run_*type*_module(event) function will be called.
 
 --ores--
 --require "locale.gen_ores.neko_crazy_ores"
+--require "locale.gen_ores.fluffy_rainbows"
+require "locale.gen_ores.glitter_ores"
```
Valansch: uncomment please :P

locale/gen_ores/glitter_ores.lua
```diff
@@ -0,0 +1,76 @@
...
+      ["iron-ore"] = 1.0,
+      ["coal"] = 0.5,
+      ["copper-ore"] = 1,
+      ["stone"] = 0.25
+   }
+   -- 1-100% chance of sprinkling any individual ore
+   sprinkle_factor = 100
+
+   -- Sets the buffer distance before ores are scrambled
+   starting_buffer = 125
```
Valansch: *noob buffer xD

locale/gen_ores/glitter_ores.lua
```diff
@@ -0,0 +1,76 @@
...
+               local position_old = entity.position
+               local amount_old = entity.amount
+
+               local surface = entity.surface
+               entity.destroy()
+               local new_entity = surface.create_entity{name = new_name, position = position_old, force="neutral", amount=amount_old}
+            end
+         else
+            -- game.print("Skipping")
+            -- game.print(entity.name)
```
Valansch: Can you remove any commented out code?

locale/gen_ores/glitter_ores.lua
```diff
@@ -0,0 +1,76 @@
+-- Glittery ores, provide a mix value, and all patches outside uranium will be a full mix.
+glitter_debug = false
+
+function run_ores_module_setup()
+	if glitter_debug then game.print("Glitter Ore: debug is enabled; module setup;") end
```
Valansch: One too many tabs spaces :O

locale/gen_ores/glitter_ores.lua
```diff
@@ -0,0 +1,76 @@
...
+   
+   ore_mix_max = 0
+   -- Prime the array
+   for a,b in pairs(ore_ratios) do
+      for i=1,(b*1000) do
+         ore_mix_max = ore_mix_max + 1
+         ore_mix[ore_mix_max] = a
+      end
+   end
+
```
Valansch: Extra line

locale/gen_ores/glitter_ores.lua
TWLTriston: Configurable!

Valansch:
Looking good, ill merge. Thank you for your work :+1:

PR #61, opened 2017-09-23
Added chunk grid terrain generation

TWLTriston:
Also added game play comments to the glitter ore for future reference

### DIFF
--- a/locale/gen_terrain/tris_chunk_grid.lua
+++ b/locale/gen_terrain/tris_chunk_grid.lua
@@ -1,0 +1,151 @@
+-- Drops a grid of concrete, hazard and brick along the chunk edges
+
+local function run_terrain_module_setup()
+   grid = {}
+   grid_widths = {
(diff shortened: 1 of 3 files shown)


locale/gen_terrain/tris_chunk_grid.lua
```diff
@@ -0,0 +1,152 @@
+-- Drops a grid of concrete, hazard and brick along the chunk edges
+
+function run_terrain_module_setup()
+   grid = {}
```
Valansch: This can be made local

Valansch:
Code looks good and really clean. Any calls to surface.set_tiles, should only be called once per tick tho. 
Tile or entity access takes a long time, but accessing n tiles is optimized and can make use of prefetching.
So setting one tile n times is going to be alot slower than setting n tiles once.

TWLTriston:
Adjusted your concerns regarding whitespace

Issue #60, opened 2017-09-22
TODO list


gentlecastaway:
So. Explosive has a TODO list that can be edited in-game. It works to direct people do take care of a certain priority. I know you do not like copying other people's code, but maybe you could check theirs just to get an idea. Maybe we can think of some feature to do our own thing.

TWLTriston:
-- Addressed with #62 PR

PR #62, opened 2017-09-24
Task List functionality

TWLTriston:
Addresses #60, based on Poll code. Needs peer review, has been tested briefly in MP


tasklist.lua
```diff
@@ -0,0 +1,243 @@
...
+
+		x = x + 1
+	end
+end
+
+
+local function tasklist_refresh()
+
+	local x = 1
+
```
Valansch: What does this x do?

tasklist.lua
```diff
@@ -0,0 +1,243 @@
...
+	global.tasklist_author = player.name
+
+	local msg = player.name
+	msg = msg .. " has created an updated tasklist!"
+
+	local frame = player.gui.left["tasklist-assembler"]
+	frame.destroy()
+
+	local x = 1
```
Valansch: And this one

tasklist.lua
```diff
@@ -0,0 +1,243 @@
...
+
+		if (frame) then
+				frame.destroy()
+		end
+
+		if (global.autoshow_tasklist_for_player[player.name] == true) then
+			tasklist_show(player)
+		end
+
+		player.print(msg)
```
Valansch: Dont you want to use game.print here?

tasklist.lua
TWLTriston: Was used in a while loop, updating to a foreach loop

tasklist.lua
TWLTriston: Yes, moving out of the loop

tasklist.lua
TWLTriston: Cleaning up the same while vs for logic

tasklist.lua
```diff
@@ -0,0 +1,213 @@
...
+	tasklist_panel_button_table.tasklist_hide_button.style.font = "default-bold"
+	tasklist_panel_button_table.tasklist_hide_button.style.minimal_height = 38
+	-- Possible for later preferences tasklist_panel_button_table.add { type = "checkbox", caption = "Show Tasklist", state = global.autoshow_tasklist_for_player[player.name], name = "auto_show_tasklist_checkbox"	}
+end
+
+local function tasklist(player)
+	local frame = player.gui.left["tasklist-assembler"]
+	frame = frame.table_tasklist_assembler
+
+	global.tasklist_items = {"","","","",""}
```
Valansch: You dont need to initalize those :D lua tables dont have a fixed size

TWLTriston: The table needs to be primed, else later references to draw the table blows up

We can either check for ~nil or prime it.  This method avoids the stop off for a check, and instead knows a value will be there

Valansch: Oh okay, yes i was assuming the texts are not nil.

TWLTriston:
Should be all set, corrected the nil reference that was being thrown

PR #63, opened 2017-09-24
Develop forcetoggle

Valansch:
Implemented /forcetoggle which makes it easier to see the enemy ghosts to spot griefers.

Waiting for #62 before merge


custom_commands.lua
```diff
@@ -365,6 +365,32 @@ local function toggle_tp_mode(cmd)
   end
 end
 
+global.old_force = {}
```
grilledham: I’m not sure if player.name is guranteed to be unique or will reference the same player given that you can change your name. I’d suggest using player.index as the key in global.old_force.

nuke_control.lua
```diff
@@ -27,5 +27,14 @@ local function on_player_deconstructed_area(event)
     end
 end
 
+
+local function on_player_mined_item(event)
```
grilledham: -I’d use the surface that the player is on, in case the map has multiple surfaces.
-I’d set ghost.last_user to the player not the player_index. according to the factorio api for [LuaEntity.last_user](http://lua-api.factorio.com/latest/LuaEntity.html#LuaEntity.last_user) needs a LuaPlayer.

I'd suggest this 
```lua
local function on_player_mined_item(event)
	if event.entity.force.name ~= "enemy" then
		local player = game.players[event.player_index]
		local ghost = player.surface.create_entity{name = "entity-ghost", position = event.entity.position, inner_name = event.entity.name, expires = false, force = "enemy"}
		ghost.last_user = player
	end
end
```

custom_commands.lua
```diff
@@ -365,6 +365,32 @@ local function toggle_tp_mode(cmd)
...
+        game.player.force = "player"
+      end
+    end
+  else
+    global.old_force[game.player.name] = game.player.force.name
+    game.player.force = "enemy"
+  end
+  game.player.print("You are now on the " .. game.player.force.name .. " force.")
+end
+
```
grilledham: If we toggle to the enemy force will we not be shot and killed by turrets, perhaps toggle to neutral force instead?

Valansch: Yes we do, i didnt want to create a new force, but also didnt think about the neutral force. Good idea! Will change.

nuke_control.lua
Valansch: Any player index, player name and lua player are unique and valid player identifiers. It's getter returns lua player but the setter accepts any player identifier.

grilledham: Fair enough, the api didn't make that clear, so I thought I'd check

Valansch: You have a point, player names are not unique and we should restrain from using them. (Even tho we have probably used them on 100 different places.)
I only know that last_user accepts everything because i tried it and all other api setters that want a user identifier can use any of them.

grilledham: Well at some point we will have to clean up the old code, but at last for now we can stop using them for future code.

custom_commands.lua
Valansch: Since a player can see the neutral forces ghosts, we are going to use the enemy force.

PR #68, opened 2017-09-28
Develop rso ores

TWLTriston:


Valansch:
Looks good. If you tested it we can merge.

PR #84, opened 2017-10-06
Dev worms

Valansch:
Waiting for #83

### DIFF
--- a/locale/gen_terrain/worms.lua
+++ b/locale/gen_terrain/worms.lua
@@ -1,0 +1,36 @@
+local Thread = require "locale.utils.Thread"
+
+local worms_per_chunk = 50
+local small_worm_spawn_distance = 100
+local medium_worm_spawn_distance = 150
+local big_worm_spawn_distance = 200
+
+
+local worm_names = {"small-worm-turret","medium-worm-turret","big-worm-turret"}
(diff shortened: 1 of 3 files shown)


locale/gen_terrain/worms.lua
```diff
@@ -0,0 +1,36 @@
...
+  if game.surfaces[1].can_place_entity{name = worm_names[worm_id], position = {x,y}} then
+    if math.sqrt(x*x+y*y) > small_worm_spawn_distance then
+      game.surfaces[1].create_entity{name = worm_names[worm_id], position = {x,y}}
+    end
+  end
+end
+
+
+function run_terrain_module(event)
+  game.print(game.tick)
```
TWLTriston: In game debug

locale/gen_terrain/worms.lua
```diff
@@ -0,0 +1,36 @@
+local Thread = require "locale.utils.Thread"
+
+local worms_per_chunk = 50
```
TWLTriston: 50 may be too many --- also the spawn area is heavily pushed in, consider tying mass worm spawn into the ore fields, such that it's not as frustrating?

Valansch: I like frustrating :D

PR #93, opened 2017-10-15
Additional fish market items, and add ons to the player gui

TWLTriston:


player_list.lua
```diff
@@ -368,5 +398,16 @@ function player_list_on_12_seconds()
 		end
 	end
 end
+
+function on_player_died( event_player, cause )
+	player = game.players[event_player.player_index]
```
Valansch: Local please. Thats a name that might come up somewhere else. (I actually called mine "player_died" just now)

PR #94, opened 2017-10-18
Updated regulars, added mona lisa gen

TWLTriston:


map_layout.lua
```diff
@@ -73,6 +74,9 @@ local on_chunk_generated = function(event)
 		end
 	else
 		run_combined_module(event)
+		if run_ores_module ~= nil then
+			run_ores_module(event)
+		end
```
Valansch: The combined modules are supposed to be the only modules running if they are enabled.

map_layout.lua
```diff
@@ -64,7 +64,8 @@ local on_chunk_generated = function(event)
 		else
 			if run_terrain_module ~= nil then
 				run_terrain_module(event)
-			elseif run_ores_module ~= nil then
+			end
+			if run_ores_module ~= nil then
```
Valansch: Can you do the same change in line 60 too please?

PR #99, opened 2017-10-24
Dev dimentions

Valansch:


locale/gen_combined/dimentions.lua
```diff
@@ -0,0 +1,53 @@
+--Author: Valansch
+
+local resource_types = {"copper-ore", "iron-ore", "coal", "stone", "uranium-ore", "crude-oil"}
+
+global.portals = {}
+--Exmaple Portal:
```
Valansch: Exmaple huehuehue

locale/gen_combined/dimentions.lua
```diff
@@ -0,0 +1,146 @@
...
+      if distance(portal.position, player.position) < global.portal_radius then
+        if not global.last_tp[player.name] or global.last_tp[player.name] + global.teleport_cooldown * 60 < game.tick then
+          player.teleport(portal.target, portal.target_surface)
+          global.last_tp[player.name] = game.tick
+          player.print("Wooosh! You are now in the " .. get_nice_surface_name(portal.target_surface) .. " dimention.")
+        end
+      end
+    end
+  end
+end
```
grilledham: This looks fine, I'm just curious if you thought about using surface.find_entities_filtered.

locale/gen_combined/dimentions.lua
```diff
@@ -0,0 +1,146 @@
...
+        for item, count in pairs(inv.get_contents()) do
+          local n_inserted = target_inv.insert{name = item, count = count}
+          if n_inserted > 0 then
+            inv.remove{name = item, count = n_inserted}
+          end
+        end
+      end
+    end
+    global.current_magic_chest_index = (global.current_magic_chest_index) % num_chests + 1 --Next portal
+  end
```
grilledham: The comment should probably read Next magic chest.

Valansch:
We have tested an older version of this in MP, its live on the "test bench" server. But not this one. I will test it before merging

locale/gen_combined/dimentions.lua
Valansch: True

locale/gen_combined/dimentions.lua
Valansch: What do you mean? What would i use it for?

Valansch: In fact i didn't. It's probably faster. I wil change it. Good point.

PR #121, opened 2017-11-19
Implemented save add_command version

Valansch:


Valansch:
Problem: Commands work, but no prints

Valansch:
@grilledham  did a thing! 9b6f8b60ccbfc484f4d5639c93a11bd16b7f2b59

PR #128, opened 2017-12-13
Fixes for .16 compat

Valansch:


Valansch:
Waiting for more playtesting

Valansch:
Waiting for #129 and #130

PR #138, opened 2017-12-21
implemented reactor meltdown

Valansch:


Valansch:
The current implementation causes stutter
also the warning should be at 600°C+

PR #139, opened 2017-12-24
Updated img to map converter for 0.16

grilledham:
Also added a template for an example / testing.


Valansch (approved):
Looks good, thanks

PR #142, opened 2017-12-25
implemented auto kick/ban

Valansch:


Valansch:
Added 2 other small fixes

nuke_control.lua
```diff
@@ -1,7 +1,11 @@
 
 
 function allowed_to_nuke(player)
+  if type(player) == "table" then
   return player.admin or is_mod(player.name) or is_regular(player.name) or ((player.online_time / 216000) > global.scenario.config.nuke_min_time_hours)
+  elseif type(player) == "number" then
+    return allowed_to_nuke(game.players[player])
```
TWLTriston: Should we return false when all else fails? Otherwise you could have no return value here

Valansch: a function without a return value returns nil, and nil is always false, therefore this function returns false in the else case.

PR #143, opened 2017-12-28
Dev anti grief tools

Valansch:


nuke_control.lua
```diff
@@ -1,7 +1,11 @@
 
 
 function allowed_to_nuke(player)
+  if type(player) == "table" then
   return player.admin or is_mod(player.name) or is_regular(player.name) or ((player.online_time / 216000) > global.scenario.config.nuke_min_time_hours)
+  elseif type(player) == "number" or type(player) == string then
+    return allowed_to_nuke(game.players[player])
+  end
```
TWLTriston: Same comment as other PR -- need to include a final return false here

PR #149, opened 2018-01-15
Action queue changes

grilledham:
I think it would be useful if actions in the action_queue stay in the queue if the action returns true.

This would help us to not fill the queue with lots of small actions. It would also allow actions to encapsulate their own data, which I find tidier.

I designed the changes so that no old code should have to be changed. I'm also not 100% sure that the changes I made will be desync safe, so I would appreciate your review.

As an example of how to use actions that can be reused, I made changes to map_gen.lua, and have tested it to the best of my abilities in single player.


locale/utils/Thread.lua
```diff
@@ -21,12 +21,25 @@ global.next_async_callback_time = -1
...
         global.actions_queue_n = global.actions_queue_n - 1
-       local callback = global.actions_queue[1]
-       local _, err = pcall(_G[callback.action], callback.params)
-       if err then log(err) end
-       table.remove(global.actions_queue, 1)
+        table.remove(global.actions_queue, 1)
+
+      -- I'm not sure how and when global values are serialized it may be necessary to re-insert the callback into the global table to update it.
+      --else
+        --global.callbacks[1] = callback
```
Valansch: if you dont remove it you wont need to reinsert it.

Valansch:
I like the idea alot. Once we have it tested we can merge it.

locale/utils/Thread.lua
Valansch: Can you removed these commented out lines? I'll merge then.

Valansch:
We had many desnycs that the  start of lines and balls, but they died down. Maybe the full queue caused the desyncs?

grilledham:
The version of lines and balls that we used on the server didn't use the action queue. Does any other part of the code base that would be active use the action queue?
I have no idea what caused the desyncs. I compared the two saves and found nothing different except the odd memory address for some tables and functions, most of which seemed part of the core game, like math.random.

Valansch:
custom_commands uses Thread.set_timeout. But since you didnt touch that i guess it is unrelated

PR #150, opened 2018-01-17
Map lines and balls

grilledham:
Final version of lines and balls.

I also added a map preview folder, I think that it might be a convenient way to keep track of what the maps look like.

This pull request is built on top of #149


grilledham:
Your right, I thought by some black magic folders that start with an underscore weren't included. I guess we have the same problem for all the data files in _locale as well.

TWLTriston:
They used to only include the locale folder and ignored the rest. Changed in .16 :(

Valansch:
No sadly not, like tris said. I'll merge this and fix the folder structure. I already started the script to clean the save. Will finish when i get to it.

PR #154, opened 2018-01-19
Restructure

Valansch:
See for yourself :)


grilledham:
all the presets (that i looked at) have:
require "map_gen.combined.grilledham_map_gen.builders"
but it should be
require "map_gen.shared.builders.lua"

grilledham:
https://github.com/Valansch/RedMew/tree/restructure/map_gen/shape/gen_islands_data 
and
https://github.com/Valansch/RedMew/tree/restructure/map_gen/shape/grid_islands_data
can be deleted, it should have been deleted when it was moved to grilledham_map_gen/data months ago.

Valansch:
Fixed

Valansch:
Fixed

grilledham:
the comment on line 4 in presets/template.lua should be changed to
-- Recommend to use generate, but generate_not_threaded may be useful for testing / debugging.

PR #155, opened 2018-01-24
threaded rainbow

Valansch:


grilledham:
Looks good, we tested it and it works, so I'm happy with it.
The only thing you might want to consider is putting a loop around do_row in fluffy_rainbows_task, so that maybe 4 rows are done per call to the task.

Valansch:
I think 32 tasks per chunk isnt excessive. If i make it even fewer tasks per chunk the speedup is gonna be even less.

Before we can implement very heavy tasks we have to implement a task weight first.

PR #156, opened 2018-01-24
Updated preset maps + added map previews + added source images

grilledham:
This pull request includes unfinished maps that I'm still working on. After the restructure I wanted to bring all my work under the new regime. Unfinished maps are marked as such in map_layout.lua.

There are also some maps I couldn't update, because I'm missing the source image, these are marked as such in map_layout.lua.

Due to 0.16 changing how water land boundaries works, we no longer get strange artefacts on chunk boundaries, so I removed the code that used to deal with that in generate.lua and generate_not_threaded.lua.

Added a binary_search function to list_utils.lua as part of an unfinished map.

Removed all preset files for unused / test maps.

I for one am impressed we collectively have 52 map previews we can show off to anyone who asks 'What maps you got.' Having said that, I hope I haven't missed any maps.


Valansch:
This looks very good.
The previews look great. 
Now we only need a readme and we might even be end user friendly :D

52 Maps donw. 48 more to go.

map_gen/combined/meteor_strike.lua
```diff
@@ -0,0 +1,174 @@
...
+            table.insert(tiles, {name = "dry-dirt", position = {x, y}})
+        end
+    end
+    local surface = game.surfaces[1]
+    surface.set_tiles(tiles, false)
+
+    game.forces.player.chart(surface, {{bx, by}, {bx+block_size, by+block_size}})
+end
+
+local function on_tick()
```
Valansch: You can use Thread.set_timeout for this. No need to have this overhead twice.

map_gen/combined/meteor_strike_data.lua
```diff
@@ -0,0 +1,23 @@
...
+            {value = 400, name = nil},
+            {value = 500, name = "iron-ore"},
+            {value = 600, name = "copper-ore"},
+            {value = 700, name = "coal"},
+            {value = 800, name = "stone"},
+        }
+    }
+}
+
+
```
Valansch: Not much happending here.

map_gen/combined/meteor_strike.lua
grilledham: That's not a bad idea.

map_gen/combined/meteor_strike_data.lua
grilledham: It's unfinished. The plan is to generate a perlin noise for each tile that is normalised between 0-1000, then these would be the thresholds for what entities spawn. I would then have multiple meteor types with different thresholds and weights.

Valansch: Okay :D

map_gen/combined/meteor_strike.lua
Valansch: Hopefully it works :D 
Do you want me to wait with the merge?

grilledham: No you should merge now. I included this because it was easier to put all the maps I wanted to keep into one branch.

Issue #157, opened 2018-01-25
Change thread implmentation


grilledham:
- [x] Change the action queue to a O(1) queue implementation.
- [x]  Add weights to the actions.
- [x]  Remove the concepts in the around threading.
- [ ]  Updates old threads.

I opened this issue so we can have a proper discussion on what we want to do with threads/ actions / tasks / whatever we are calling them.

The hash queue implementation is fairly straight forward, so I'm willing to sort that out / test for desyncs.

Coming from a mainly c# background, I'd rename the whole threading concept to tasks.
- Thread.lua -> Task.lua
- action_queue -> task_queue
- actions / callbacks -> tasks
- Thread.set_timeout_in_ticks -> Task.set_timeout_in_ticks

I am of course open to other naming suggestions.

Regarding action weights, I see there as being 2 approaches:
1. We normalise all actions so that they take roughly the same amount of time per call, thus eliminating the need for weights.
2. We add a cost field to each action's args object that represents the approximate time cost to call that action on an arbitrary scale say 1 - 100. We then have the action queue set to keep running actions until it reaches a cost of say 100. 
I would make this a fixed value and not dynamic as I see the point of the action_queue to reduce execution cost per tick. Having more actions in the queue doesn't mean we should sacrifice performance just that we have to wait longer. And with a better queue implementation, big queue will have a minimal performance impact.
I would also allow the action when run to change it's cost value, because each call to the action might have a different expected time cost.

Let me know what you think. I am also interested in any alternative suggestions you might have.

grilledham:
> Change the action queue to a O(1) queue implementation.

I've had some success with #158 on the test server. 
The test was the island_resort map and chart 2048 tiles around spawn. I did this 4 times and desynced 1 out of 4 times.
Clearly more testing is needed.

Valansch:
It's kinda funny and sad that we consider desyncing 25% of the time as a success :D

Valansch:
So if #158 works we can surely use your queue implementation.

Still open is the question about tasks per second/weight/cost.

grilledham:
I think i still need more time to think about weight/costs.

But I guess a good way to see if it's a good idea is to implement it and see what happens.

grilledham:
I implemented weights. Let me know what you think.

The more I think about it, implementing costs make less sense. If we suspect a task is going to take a long time we should just split it up into smaller tasks.

I was looking a the code for async callbacks and noticed it could probably use some attention.
[...]

PR #158, opened 2018-01-25
Action queue redo

grilledham:
I'm making this pull request so that I can reference it in #157

### DIFF
--- a/map_layout.lua
+++ b/map_layout.lua
@@ -6,7 +6,7 @@ in this file and your run_*type*_module(event) function will be called.
 --]]
 
 --combined--
-require "map_gen.combined.island_resort"
+--require "map_gen.combined.island_resort"
 --require "map_gen.combined.red_planet_v2"
 --require "map_gen.combined.borg_planet_v2"
 --require "map_gen.combined.dimensions"
--- a/map_layout.lua
+++ b/map_layout.lua
@@ -6,7 +6,7 @@ in this file and your run_*type*_module(event) function will be called.
 --]]
 
 --combined--
---require "map_gen.combined.island_resort"
+require "map_gen.combined.island_resort"
 --require "map_gen.combined.red_planet_v2"
 --require "map_gen.combined.borg_planet_v2"
 --require "map_gen.combined.dimensions"
--- a/custom_commands.lua
+++ b/custom_commands.lua
@@ -1,4 +1,4 @@
-local Thread = require "utils.Thread"
+local Task = require "utils.Task"
 require "utils.event"
 
 function player_print(str)
@@ -108,7 +108,7 @@ local function walkabout(cmd)
   
   if non_colliding_pos then
     game.print(player_name .. " went on a walkabout, to find himself.")
-    Thread.set_timeout(duration, return_player, {player = player, force = player.force, position = {x = player.position.x, y = player.position.y}})
+    Task.set_timeout(duration, return_player, {player = player, force = player.force, position = {x = player.position.x, y = player.position.y}})
     player.character = nil
     player.create_character()
     player.teleport(non_colliding_pos)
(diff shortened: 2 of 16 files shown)


map_layout.lua
```diff
@@ -6,7 +6,7 @@ in this file and your run_*type*_module(event) function will be called.
 --]]
 
 --combined--
---require "map_gen.combined.island_resort"
+require "map_gen.combined.island_resort"
```
Valansch: :D

Valansch:
Why did you rename the callbacks to actions? Didn't we want to name both "task"?

grilledham:
I wrote this before I saw that you agreed with my renaming suggestion. I didn't want to presume so I kept the action name. 

I'll make the renaming changes tomorrow.

grilledham:
Change the task for island_resort to reusable tasks.

I used to run the map at 90 - 110 ups, now I can run it at 160 - 175 ups.

Tested on the test server for desyncs.

grilledham:
90 - 110 is with the new queue, 160-175 is with reusable tasks. I think it was around 20 - 30 with the old queue.
My guess is the extra performance comes from less data being serialized, whenever the global table gets serialized. Could also be less pressure on the garbage collector.

grilledham:
Ok I ran the test again, this time multiplying the queue size by 34 in the get_task_per_tick function.
I get around 110 - 120 ups now.

Valansch:
>My guess is the extra performance comes from less data being serialized, whenever the global table gets serialized. Could also be less pressure on the garbage collector.

Agree thats most likely all contributing to better UPS.

>Or it could be that with a smaller queue, it is running less tasks per tick. In which case it's hard to measure the performance impact.

That shouldn't matter, if game speed is at some very high value. As the game is already running as fast as possible.

Valansch:
>Ok, I ran the test again, this time multiplying the queue size by 34 in the get_task_per_tick function.
I get around 110 - 120 ups now.

Do you get different performance simply by multiplying the queue size by a value?

grilledham:
Yes, because it's running more tasks per tick

grilledham:
2 major changes:
1. Changed callbacks to use a priority queue, this simplifies the code in Task.lua.
2. Callbacks now store a string for the function name rather than the function itself, this makes them consistent with how tasks work.

1 minor change - Added global.task_queue_speed.

All tested for desyncs on the test server.

After your review/ approval I think this branch is ready to be merged.

Valansch:
I think we can optimize get_task_per_tick a bit. We dont need it to change too often. Since its running every tick i dont want to caclulate log10 every tick.

```lua
global.tpt = global.task_queue_speed
function get_task_per_tick()  
  if game.tick % 300 == 0 then
    local size = global.total_task_weight
    local tpt = math.floor(math.log10(size + 1)) * global.task_queue_speed
    if apt < 1 then
      global.tpt = tpt
    else
      global.tpt = tpt
    end
  end
  return global.tpt
end
```

Valansch:
I dont understand what the purpose of the PriorityQueue is. Whats the difference and why couldn't you use the normal Queue?

grilledham:
A Priority Queue is like a normal Queue except the 'highest' priority element is removed each time pop is called. In this case the 'highest' priority element is the callback with the lowest time value.

A normal queue can't be used as there is no guarantee that the order in which the callbacks are inserted is the same as the order in which they should be removed. For example this code.
```
Task.set_timeout(20, "some_function")
Task.set_timeout(10, "some_function")
```

Valansch:
Okay that makes sense. This looks pretty good. I see nothing more than my comment earlier that i would change. Want me to merge, or do you want to change that?

Valansch:
I'll merge now and test.

grilledham:
Sorry I was just taking a long time to think about your optimisation proposal for get_task_per_tick().
Without benchmarking the performance impact, I'm wary of making any micro optimisations that make the code more complicated. 

Optimising code is still fun though, so perhaps this is an issues we can revisit in the future.

Valansch:
Nothing to be sorry about. I only waited an hour :D I wanted to launch the map, so i merged. I also applied the fix. Now its too late :P Unless you want to reverse it. I don't think it makes it too complicated.

grilledham:
Merging was the right thing to do. The code that is in this branch was ready and that's all that matters. Regarding your fix, I don't think it does any harm, so I'm happy with it.

PR #160, opened 2018-01-31
Changes to Control.lua, New file (server-api.lua)

Toledin:
added to control.lua : require "server_api"
added new file "server_api.lua"

### DIFF
--- a/server_api.lua
+++ b/server_api.lua
@@ -1,0 +1,7 @@
+function get_online_players()
+    game.write_file("api-output", "", false,0)   --delete 
+    local online_players = game.connected_players
+    for _,p in pairs(online_players) do 
+      game.write_file("api-output", p.name, true,0) 
+    end
+end
--- a/server_api.lua
+++ b/server_api.lua
@@ -5,3 +5,7 @@ function bot_command_players(filename)
       game.write_file(filename, p.name, true,0)
     end
 end
+
+function bot_command_time()
+  game.print(math.floor(game.tick / 60))
+end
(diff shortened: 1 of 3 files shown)


server_api.lua
```diff
@@ -0,0 +1,7 @@
+function get_online_players()
+    game.write_file("api-output", "", false,0)   --delete
```
Valansch: One think i forgot to mention was: Maybe we want a headline, something like "Online players (X):"
In this case we can just replace line 2 with:
game.write_file("api-output", string.format("Online players (%d):",#game.connected_players), false,0)

server_api.lua
```diff
@@ -0,0 +1,7 @@
+function get_online_players()
```
Valansch: What i just thought of is also: have a parameter called filename, that will be used instead of "api-output". This will prevent collision if two commands are called. The bot can just choose whatever name it wants.

server_api.lua
Toledin: Yeah I’ll try it, but it writes the players into the file or files with a loop so if we make a headline it wouldn’t print the headline more times?

Valansch: It only prints the first one once. It used to be "" just to delete the file, now i put the headline in.

Toledin: k and for the other part if we want it to say the amount of players like (x) we will have to concatenate the value of the quantity of the players into the string?

Valansch: # counts the number of table entries. The line above will do what we want.

server_api.lua
```diff
@@ -0,0 +1,7 @@
+function get_online_players(filename)
+    game.write_file(filename, "Online players: (1)", false,0)
```
Valansch: Now it will always print 1. You need #game.connected_players

PR #165, opened 2018-02-25
fix to not remove uranium mines

grilledham:
Wont work for mixed ore patches, but this is probably good enough.

I also noticed in testing that if you mark the mine for deconstruction on the same tick that it depletes, the mine doesn't output the last mined ore. So I added a 5 tick delay to fix this.

### DIFF
--- a/autodeconstruct.lua
+++ b/autodeconstruct.lua
@@ -1,4 +1,7 @@
 --Author: Valansch
+
+local Task = require "utils.Task"
+
 local function is_depleted(drill)
   local position = drill.position
   local area = {}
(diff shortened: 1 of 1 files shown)


Valansch:
I think we have a fix already for that. 14a6dae5e31187b61f3309fde720d9d03959b5f1
I fixed it 1 day after you did because i didnt see the PR

autodeconstruct.lua
```diff
@@ -16,13 +19,20 @@ local function is_depleted(drill)
...
 end
 
+function autodeconstruct_order_deconstruction(drill)
+  drill.order_deconstruction(drill.force)
+end
+
 local function mark_if_depleted(drill)
   if is_depleted(drill) then
-    drill.order_deconstruction(drill.force)
+    Task.set_timeout_in_ticks(5, "autodeconstruct_order_deconstruction", drill)
```
Valansch: Just out of curiosity, why the 5 ticks delay?

grilledham: If you mark a mine for deconstruction on the same tick it runs out of ore, it doesn't output the last one. From my testing 5 ticks seems to be enough time to make sure (unless the belt is backed up).

PR #167, opened 2018-03-24
added contraSpiral map, made by Zijkhal

grilledham:


grilledham:
added cookies map in this pull reuqest

PR #168, opened 2018-04-02
Player spy

grilledham:
image

I'm working on adding gui that lets remotely view another player. More testing is required, but I wanted to make this pull request, in case you have any suggestions.

Currently any player can use it. We might want to make it admin only or at least make the tp button admin only.

We might also consider merging this into the player list gui somehow.

### DIFF
--- a/control.lua
+++ b/control.lua
@@ -23,6 +23,7 @@ require "tasklist"
 require "autodeconstruct"
 require "corpse_util"
 require "reactor_meltdown"
+require "player_spy"
 
 local function player_joined(event)
 	local player = game.players[event.player_index]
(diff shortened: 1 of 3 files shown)


Valansch:
Good idea. I always wanted something like this.

Is it possible to move the players main camera to a player?

grilledham:
I'm not aware of any api that let's you move the player's main camera or minimap. However what you could do is, disconnect the player from their character and tp the player to the target position. Then on demand reattach the player back with their character.

Valansch:
Thats true, that could work. I can get to implement that this weekend. Can you implement a button that says something like "focus player" and executes a function focus_player(player_id). I will implement the function then.

grilledham:
If I get time this evening I might try and implement it. I think it's fairly straight forward, unless I'm overlooking something. The problem is that it's hard to test this code without volunteers. Unless there is a way I can make dummy accounts/ players that I can control?

Valansch:
Just ask in discord, people will volunteer

PR #170, opened 2018-04-06
events rewrite

grilledham:
So I've wanted to change how the event module works for a while, and seeing all the script.on_nth_tick in #169 gave me enough reasons to do it.

There are two parts to this rewrite.
Part 1 includes:
 - added support for registering on_nth_tick events (Event.on_nth_tick(tick, function)).
- removed Event.remove because it would cause desyncs.
- the on_init and on_load events are now called through Event.on_init and Event.on_load rather than using magic numbers as the event name.
- added a debug mode that doesn't call the events with pcall, so for testing you can see errors in events as they happen, rather than looking through the logs. (We might want to make a global debug_mode variable.)
- made Event local as to not pollute the global namespace, this is inline with how Task works.
- In my opinion made the code simpler to understand, but that's subjective.
- Renamed Event.register to Event.add, because I was planning to use register in part 2.

Part 2 is everything from line 88 (in event.lua) onwards and is an experimental idea I had for allowing event handlers that can be safely added and removed at run time. Even if the event handlers are closures. An example of how to use it would be.
```lua
local Token = require "utils.global_token"

local capture = "some variable"
local function closure(event)
    if event.tick % 60 == 0 then
            game.print(capture)
    end
end

-- this should only be called at the control stage, calling this inside events is not safe.
local token = Token.register(closure)

-- this can be called anywhere.
Event.add_removable(defines.events.on_tick, token)

local function some_other_event(event)
	Event.remove_removable(defines.events.on_tick, token)
end
```
One quirk of the system is that when the event is raised, removable events are always called after non-removable events regardless of the order they are added.

Anyway this is just my take on one way we could do events. If there are any parts you don't like or think should be changed, I'd be glad to hear it.


Valansch:
The register/add change will mean its not backwards compatible. In case we want to use our code together with another scenario like crash site. Maybe we can swap the names register and add?

I was planning to have a "submodule" folder that only has a dummy control.lua. This is tho make it easy for guys like red to just paste in another scenario and override the dummy.
Idk if thats a big issue but i think i would have to break your closure again for that.

Something like:
```lua
require "submodule.control"

if submodule_not_active then
    Event  = require "utils.event"
end
```

Also maybe token isn't the best name. Maybe something like reserved_functions/predefined_functions something like that?

utils/event.lua
```diff
@@ -1,131 +1,152 @@
...
+local function on_nth_tick_event(event)
+    local handlers = on_nth_tick_event_handlers[event.nth_tick]
+    call_handlers(handlers, event)
 end
 
+function Event.add(event_name, handler)
+    local handlers = event_handlers[event_name]
+    if not handlers then
+        event_handlers[event_name] = {handler}
+        script.on_event(event_name, on_event)
```
Valansch: Is this desync save as long as ``on_event`` was defined on load?

grilledham:
I agree on both of the naming points you raise. I'll change add back to register. And I'll think of a better name of tokens/ global_tokens.lua

Valansch:
Merging as discussed in discord

PR #171, opened 2018-04-12
Added plus map

Rickeyhb:
Added "my" first map, discussed with @grilledham about it on Discord.

Everything should be like you want it :)


grilledham:
Thanks for making the pull request and for making the source image for the map.

You are missing plus.lua in map_gen/data/presets. The processed data file.

Also some small requests. Can we change: 
 - `--map = scale(map, 2, 2)` to `map = scale(map, 1.5, 1.5)` as that is what we used for the map we played.
 - can you generate the map in single player using this command
/c game.forces.player.chart(game.player.surface, {{-512, -512},{512, 512}})
and take a screenshoot and put it in map_previews.
- can you include the source image for this map and put it in source_images

Rickeyhb:
Thanks for the comment, now it should all be fixed :) 

And now I know what I need to prepare when making other maps.

Valansch:
@grilledham not trying to put this on you, im happy to do it. But you can also merge the pull requests if you find them to be acceptable

@Rickeyhb thanks for your contribution!

PR #173, opened 2018-04-17
Added Honeycomb map and updated "map_gen/presets/template.lua"

Rickeyhb:
****


Valansch:
Can you maybe add a preview?

PR #178, opened 2018-06-02
Util cleanup

Valansch:
waiting for #177


user_groups.lua
```diff
@@ -1,32 +1,29 @@
...
-	if game.player then return game.player.name end
-	return "<server>"
-end
 
-local is_regular = function(player_name)
-	return cast_bool(global.regulars[player_name] or global.regulars[player_name:lower()]) --to make it backwards compatible
+
+Module.is_regular = function(player_name)
+	return false or global.regulars[player_name] or global.regulars[player_name:lower()] --to make it backwards compatible
 end
```
grilledham: Am I missing something?
```
false or ...
```
doesn't do anything.

did you perhaps mean to put `or false` at the end of the line to return `false` rather than `nil`?

utils/utils.lua
```diff
@@ -1,498 +1,36 @@
--- utils.lua by binbinhfr, v1.0.16
--- A 3Ra Gaming revision
--- define debug_status to 1 or nil in the control.lua, before statement require("utils")
--- define also debug_file and debug_mod_name
-
 local Event = require "utils.event"
```
grilledham: This line can be removed.

utils/utils.lua
```diff
@@ -1,498 +1,36 @@
...
 		if p.connected and p.admin then
 			p.print(msg)
 		end
 	end
 end
+
+Module.get_actor = function()
+	if game.player then return game.player.name end
+	return "<server>"
+end
```
grilledham: Need a `return Module` at the end of file.

user_groups.lua
Valansch: actually i wanted to cast it to a bool. But both methods for a positive answer. I thought (false or "") would return true. My bad. Will change.

utils/utils.lua
Valansch: I thought i did O.o

PR #180, opened 2018-06-04
Implemented antigrief tools

Valansch:
includes:
- /undo player
  undoes everything a player has done (Admins only)
- /antigrief_surface
moves you to the antigrief surface or back (Admins only)

### DIFF
--- a/antigrief.lua
+++ b/antigrief.lua
@@ -1,13 +1,13 @@
---local Event = require "utils.event"
-local surface = {}
+local Event = require "utils.event"
+
 
 Event.on_init(function()
-  surface = game.create_surface("antigrief", {  autoplace_controls = {    coal = {      frequency = "normal",      richness = "normal",      size = "none"    },    ["copper-ore"] = {      frequency = "normal",      richness = "normal",      size = "none"    },    ["crude-oil"] = {      frequency = "normal",      richness = "normal",      size = "none"    },    desert = {      frequency = "normal",      richness = "normal",      size = "none"    },    dirt = {      frequency = "normal",      richness = "normal",      size = "none"    },    ["enemy-base"] = {      frequency = "normal",      richness = "normal",      size = "none"    },    grass = {      frequency = "normal",      richness = "normal",      size = "very-high"    },    ["iron-ore"] = {      frequency = "normal",      richness = "normal",      size = "none"    },    sand = {      frequency = "normal",      richness = "normal",      size = "none"    },    stone = {      frequency = "normal",      richness = "normal",      size = "none"    },    trees = {      frequency = "normal",      richness = "normal",      size = "none"    },    ["uranium-ore"] = {      frequency = "normal",      richness = "normal",      size = "none"    }  },  cliff_settings = {    cliff_elevation_0 = 1024,    cliff_elevation_interval = 10,    name = "cliff"  },  height = 2000000,  peaceful_mode = false,  seed = 3461559752,  starting_area = "very-low",  starting_points = {    {      x = 0,      y = 0    }  },  terrain_segmentation = "normal",  water = "none",  width = 2000000})
-  surface.always_day = true
+  global.ag_surface = game.create_surface("antigrief", {  autoplace_controls = {    coal = {      frequency = "normal",      richness = "normal",      size = "none"    },    ["copper-ore"] = {      frequency = "normal",      richness = "normal",      size = "none"    },    ["crude-oil"] = {      frequency = "normal",      richness = "normal",      size = "none"    },    desert = {      frequency = "normal",      richness = "normal",      size = "none"    },    dirt = {      frequency = "normal",      richness = "normal",      size = "none"    },    ["enemy-base"] = {      frequency = "normal",      richness = "normal",      size = "none"    },    grass = {      frequency = "normal",      richness = "normal",      size = "very-high"    },    ["iron-ore"] = {      frequency = "normal",      richness = "normal",      size = "none"    },    sand = {      frequency = "normal",      richness = "normal",      size = "none"    },    stone = {      frequency = "normal",      richness = "normal",      size = "none"    },    trees = {      frequency = "normal",      richness = "normal",      size = "none"    },    ["uranium-ore"] = {      frequency = "normal",      richness = "normal",      size = "none"    }  },  cliff_settings = {    cliff_elevation_0 = 1024,    cliff_elevation_interval = 10,    name = "cliff"  },  height = 2000000,  peaceful_mode = false,  seed = 3461559752,  starting_area = "very-low",  starting_points = {    {      x = 0,      y = 0    }  },  terrain_segmentation = "normal",  water = "none",  width = 2000000})
+  global.ag_surface.always_day = true
+ 
 end)
 
 local function place_entity_on_surface(entity, surface, replace, player)
-  if entity.surface == surface then return end
   if replace then
     for _,e in pairs(surface.find_entities_filtered{position = entity.position}) do
       e.destroy()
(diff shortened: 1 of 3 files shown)


Valansch:
Still left do to:
- [ ] Support entity settings
- [ ] Catch on_player_rotated_entity 
Save original orientation on the antigrief surface and store original last_user in a global table
- [ ] Catch on_entity_settings_pasted
Similar to the previous
- [ ] Do not store entity on antigrief surface if the remover is the last_user
- [ ] Exclude admins

Valansch:
Okay so it should be able to handle: Bots, rotation, settings pasting, destroying and picking up.

antigrief.lua
```diff
@@ -0,0 +1,185 @@
+local Event = require "utils.event"
+local Utils = require "utils.utils"
+
+global.original_last_users_by_ent_pos = {}
+
+Event.on_init(function()
+  global.ag_surface=game.create_surface("antigrief",{autoplace_controls={coal={frequency="normal",richness="normal",size="none"},["copper-ore"]={frequency="normal",richness="normal",size="none"},["crude-oil"]={frequency="normal",richness="normal",size="none"},desert={frequency="normal",richness="normal",size="none"},dirt={frequency="normal",richness="normal",size="none"},["enemy-base"]={frequency="normal",richness="normal",size="none"},grass={frequency="normal",richness="normal",size="very-high"},["iron-ore"]={frequency="normal",richness="normal",size="none"},sand={frequency="normal",richness="normal",size="none"},stone={frequency="normal",richness="normal",size="none"},trees={frequency="normal",richness="normal",size="none"},["uranium-ore"]={frequency="normal",richness="normal",size="none"}},cliff_settings={cliff_elevation_0=1024,cliff_elevation_interval=10,name="cliff"},height=2000000,peaceful_mode=false,seed=3461559752,starting_area="very-low",starting_points={{x=0,y=0}},terrain_segmentation="normal",water="none",width=2000000})
```
grilledham: Just a thought but if you change it so the anti grief surface only generates water rather than grass then you don't need to remove entities in the on chunk generated event.

antigrief.lua
```diff
@@ -0,0 +1,185 @@
...
+Event.add(defines.events.on_chunk_generated, function(event)
+  if event.surface.name == "antigrief" then
+    for _,e in pairs(event.surface.find_entities_filtered{area = event.area, force = "neutral"}) do
+      if e.type ~= "player" then
+        e.destroy()
+      end
+    end
+    local tiles = {}
+    for x = event.area.left_top.x, event.area.right_bottom.x do
+      for y = event.area.left_top.y, event.area.right_bottom.y do
```
grilledham: this has an off by one error it should be
```lua
for x = event.area.left_top.x, event.area.right_bottom.x - 1 do
      for y = event.area.left_top.y, event.area.right_bottom.y - 1 do
```

antigrief.lua
```diff
@@ -0,0 +1,185 @@
...
+
+local function get_position_str(pos)
+  return string.format("%d|%d", pos.x, pos.y)
+end
+
+local function on_entity_changed(event)
+  local entity = event.entity or event.destination
+  local player = game.players[event.player_index]
+  if player.admin then return end --Freebees for admins
+  if entity.last_user ~= player and entity.force == player.force then --commented out to be able to debug
```
grilledham: You should check if the entity is valid before accessing it, another event handler might destroy the entity before this event handler is called.

antigrief.lua
```diff
@@ -0,0 +1,185 @@
...
+        end
+      end
+  end
+end)
+
+Event.add(defines.events.on_player_mined_entity, on_entity_changed)
+
+Event.add(defines.events.on_marked_for_deconstruction, function(event)
+  global.original_last_users_by_ent_pos[get_position_str(event.entity.position)] =
+    event.entity.last_user.index
```
grilledham: Not all entities have a `last_user`. For example if you deconstruction planner a tree this line crashes.

antigrief.lua
```diff
@@ -0,0 +1,185 @@
...
+end)
+
+
+local Module = {}
+
+Module.undo = function(player)
+  local player = player
```
grilledham: This line does nothing and can be removed

antigrief.lua
```diff
@@ -0,0 +1,185 @@
...
+end
+
+Event.add(defines.events.on_robot_pre_mined, function(event)
+  --The bot isnt the culprit! The last user is! They marked it for deconstruction!
+  event.player_index = Utils.ternary(event.entity.last_user, event.entity.last_user.index)
+  on_entity_changed(event)
+end)
+
+local function get_pre_rotate_direction(entity)
+  --Some entities have 8 rotation steps and some have 4. So a mathmatical reverse is not possible
```
grilledham: I haven't fully read the code, but just in case it's relevant rocket silos only have 2 rotations and chests 1.

antigrief.lua
Valansch: That's true. I tried to check if last_user is nil everywhere. Missed this one.

antigrief.lua
Valansch: :roll_eyes:

antigrief.lua
Valansch: What i meant to say with this line is: I cant just do:
`direction_pre_rotation =( direction - n) % 8`, because i dont know how many possible orientations an entity has. This will work for any number of orientations. So 2 is also covered. If an entity only has one possible orientation, then then the event wont trigger.

antigrief.lua
Valansch: Good idea. Consider it changed.

antigrief.lua
```diff
@@ -0,0 +1,183 @@
...
+  global.ag_surface=game.create_surface("antigrief",{autoplace_controls={coal={frequency="normal",richness="normal",size="none"},["copper-ore"]={frequency="normal",richness="normal",size="none"},["crude-oil"]={frequency="normal",richness="normal",size="none"},desert={frequency="normal",richness="normal",size="none"},dirt={frequency="normal",richness="normal",size="none"},["enemy-base"]={frequency="normal",richness="normal",size="none"},grass={frequency="normal",richness="normal",size="none"},["iron-ore"]={frequency="normal",richness="normal",size="none"},sand={frequency="normal",richness="normal",size="none"},stone={frequency="normal",richness="normal",size="none"},trees={frequency="normal",richness="normal",size="none"},["uranium-ore"]={frequency="normal",richness="normal",size="none"}},cliff_settings={cliff_elevation_0=1024,cliff_elevation_interval=10,name="cliff"},height=2000000,peaceful_mode=false,seed=3461559752,starting_area="very-low",starting_points={{x=0,y=0}},terrain_segmentation="normal",water="normal",width=2000000})
+  global.ag_surface.always_day = true
+
+end)
+
+local function is_mocked(entity)
+   return pcall(function() return entity.mock end)
```
grilledham: In my testing this line can be replaced with
```lua
return rawget(entity, 'mock')
```
which would probably perform better than throwing and catching an error

antigrief.lua
```diff
@@ -0,0 +1,183 @@
...
+end
+
+local function place_entity_on_surface(entity, surface, replace, player)
+  local new_entity = nil
+  for _,e in pairs(surface.find_entities_filtered{position = entity.position}) do
```
grilledham: `surface.find_entities_filtered` returns an array so you can use `ipairs` rather than `pairs` which supposedly performs better.
Same is true for line 136 and 144.

custom_commands.lua
```diff
@@ -472,6 +473,38 @@ local function pool()
...
+      game.player.print(
+        string.format("Warning! You are about to remove %s entities and restore %s entities.",
+        #Utils.find_entities_by_last_user(game.players[cmd.parameter], game.surfaces.nauvis),
+        Antigrief.count_removed_entities(game.players[cmd.parameter]))
+      )
+      game.player.print("To execute the command please run it again.")
+      return
+    end
+    Antigrief.undo(game.players[cmd.parameter])
+    game.print(string.format("Undoing everything %s did...", cmd.parameter))
```
grilledham: When the command successfully runs do we want to remove the player from the `global.undo_warned_players` table so that the next time they run the command they get the nice warning message?

Valansch: My thinking was the player will remember the warning so we dont need a second. Now that you are saying it, i think its better to warn them every time. Maybe someone will expect it to always behave like this and run it just to check for the number of entities or something...

Valansch:
About the kill switch. The only way to do this would be to have a check in every event handler:
`if global.undo_killswitch then return end`

Can you just add this when creating the map without it in the codebase?

grilledham:
Yeah I can do that

PR #182, opened 2018-06-04
Patch 1

JCA122204:


grilledham:
So we don't really have any code style guidelines and we haven't discuss it before. I think we all sort of just tolerate each other. While I agree that the code base is messy at times and inconsistent, some of which (probably most) is my fault. I have a habit of turning 2 spaces into 4 and changing double quotes into single quotes, it's just what the formatter on my IDE does and I haven't changed it. 

But we haven't done PR on just code style before and before we start we might want to discus what code style we want. I will also say that I'm personally not a fan of single line `if then end` blocks

JCA122204:
Personally, if it is a large block of code, or a single long line, I would do the non inline.

If is is like
`if (frame) then
    frame.destroy()
  end`

Then I would shorten it down

PR #188, opened 2018-06-15
Update info.lua

theorangeangle:
take 2

removed the BTC link and replaced with patreon, and replaced a couple of words with synonyms.

### DIFF
--- a/info.lua
+++ b/info.lua
@@ -31,7 +31,7 @@ And this is what you ought to know:
 - No political, racist, or hateful content.
 - If you suspect you desync, restart the game.
 - Join our growing community on redmew.com/discord
-- Contribute to our Hookers and Blow fund at:
+- Contribute to our Hookers & Blow fund at:
   Patreon: patreon.com/redmew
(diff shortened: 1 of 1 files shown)


grilledham:
I approve the changes.

Just quickly before this is merged:
 - who does `paypal.me/jsuesse` go to?
 - the `redmew.com/patreon`  link appears to be broken?

Valansch:
jsuesse is diffie. Also why do hookers have to go? First you take the title puns then you take the hookers? o.O

theorangeangle:
I want the title puns!

Just a little worried about the hookers because a guy PMed me once asking if we legit "got together once a year and had a party"

Valansch:
>Just a little worried about the hookers part because a guy PMed me once asking if we legit "got together once a year and had a party"

Thats hilarious. But i dont see why that would motivate to remove the line :D

theorangeangle:
got it - changed it back to hookers
(definitely didn't take 3 tries because i'm still new to this github thing)

PR #192, opened 2018-08-04
Add train saviour softmod

plague006:
This makes players invulnerable to train-related deaths so long as they hold the required number of tokens. Tokens are expended as they save lives.

Currently if a player is in a vehicle their life will *not* be spared. And I quite frankly have *no* idea how to resolve that.

### DIFF
--- a/train_saviour.lua
+++ b/train_saviour.lua
@@ -1,12 +1,12 @@
 local Event = require 'utils.event'
 
-search_range = 1 -- distance to search for the killer train
-saviour_token_name = "coin" -- item name for what saves players
-saviour_timeout = 60 -- number of ticks players are train immune after getting hit (roughly)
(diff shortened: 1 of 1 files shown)


theorangeangle:
We might want to increase the cost a little from 1 coin, as train deaths can be more frequent than expected and 1 almost feels too cheap (I speak from experience with train deaths :D )

plague006:
The idea was to have coins for purchase in the fish market which would dictate the actual cost. But it seems grilled wants to use coins as the new currency in the market so that kinda throws things in the air. 😄

If coins become the new currency I would propose probably go for a buff system similar to mining+walk speed. Much harder for people to keep track of though. We'll see which way currency goes I guess.

Valansch:
waiting for @grilledham

grilledham:
I re-wrote this in #209 which supersedes this PR, so I'm closing

PR #193, opened 2018-08-05
Add plague's colors

plague006:
Pure vanity..


Valansch:
I'd rather not have anything player specific in the codebase. Instead of this we could just save all players colors. This should be easily doable on the diffie server. Not so much on the prime server.

plague006:
That's cool, just saw grilled's color and followed suit.

Valansch:
Didnt see that. Sneaky :D

Valansch:
Cant blame you then, so i may as well merge it.

plague006:
Would/could happily split it out into its own resources file so that it's not edits to control.lua if others want to add their own colours.

theorangeangle:
Possible patreon perk?

(Assuming we can get someone to donate on patreon first)

plague006:
Opened https://github.com/Valansch/RedMew/issues/195 to follow up.

PR #196, opened 2018-08-06
Add jail and unjail commands

plague006:
This could potentially be a replacement for walkabout, but for that to occur it would likely need to account for a time parameter.

Closes https://github.com/Valansch/RedMew/issues/6

Edit: I know I should break the group checking and creation into its own function since it's repeated, but I'm not sure how to do that properly.

### DIFF
--- a/custom_commands.lua
+++ b/custom_commands.lua
@@ -682,8 +682,6 @@ local function unjail_player(cmd)
   end
 end
 
-
-_DEBUG = true
 if not _DEBUG then
     local old_add_command = commands.add_command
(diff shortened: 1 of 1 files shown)


plague006:
Oh, and I know I should break the group checking and creation into its own function since it's repeated, but I'm not sure how to do that properly.

grilledham:
Good work, thanks.
I'm going to make some minor changes then merge this into the master branch on my Github so it's ready for today's map.

grilledham:
This is now part of #194

PR #197, opened 2018-08-07
Add a check before granting all techs to player force

plague006:
Because I might have been the first to do it, but I would also like to be the last.


Valansch:
Good idea! (y)

plague006:
I can look into ignoring for the first 30 mins when I get home.

PR #202, opened 2018-08-09
Make the language we use a little more mature/professional

plague006:
We don't want to lose the quirk/silliness we have, but if we want patrons/donations it might be better to not call it a hooker+blow fund.

Instructing admins not to cheat or in-fight implies a problem so big that it needs to be addressed publicly. (Which might be the case, but our users don't need to know that)

Make admins immune to the patreon/discord spam. More just quality of life for admins since we talk about discord a lot without actually needing the link ourselves.

### DIFF
--- a/control.lua
+++ b/control.lua
@@ -38,7 +38,7 @@ local function player_created(event)
     player.insert {name = 'coin', count = 10}
     player.insert {name = 'iron-gear-wheel', count = 8}
     player.insert {name = 'iron-plate', count = 16}
-    player.print('Welcome to our Server. You can join our Discord at: redmew.com/discord')
+    player.print('Welcome to our Server. You can join our Discord at: redmew.com/discord or check out our patreon at patron.com/redmew')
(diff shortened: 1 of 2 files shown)


plague006:
Ack, going to have to remember some git to get rid of https://github.com/Valansch/RedMew/pull/202/commits/e3174f3e91bdd858edc7430f314bae6e9f7bdfd3

grilledham:
I agree with the first two points.

I also agree with the patreon/discord spam point, but we need to communicate that change to the admin team else they will think it's broken.

As an aside I am currently working on redoing the info screen #204.
I will need help with the content / text. When It's closer to complete I'll post samples to the admin channel to try and get contributions from the admin team.

plague006:
Could educate on 2 fronts: an advisory in admin chat on discord (which will probably get missed by most) and a heads up in-game which I'll add to this PR shortly.

Valansch:
the info_rewrite merge fixed this

PR #203, opened 2018-08-09
Add a check before granting all techs to player force

plague006:
Because I might have been the first to do it, but I would also like to be the last.

Supercedes and closes https://github.com/Valansch/RedMew/pull/197

### DIFF
--- a/custom_commands.lua
+++ b/custom_commands.lua
@@ -604,12 +604,9 @@ local function unjail_player(cmd)
 end
 
 local function all_tech()
-    if game.player and _DEBUG then
-        local player = game.player
-        player.force.research_all_technologies()
+    if game.player
+        game.player.force.research_all_technologies()
         player_print('Your force has been granted all technologies')
-    else
-        player_print('This command is only available in debug mode')
     end
 end
 
@@ -654,7 +651,9 @@ commands.add_command(
 
 commands.add_command('tempban', '<player> <minutes> Temporarily bans a player (Admins only)', tempban)
 commands.add_command('zoom', '<number> Sets your zoom.', zoom)
-commands.add_command('all-tech', 'researches all technologies (debug only)', all_tech)
+if _DEBUG then
+    commands.add_command('all-tech', 'researches all technologies (debug only)', all_tech)
+end
 commands.add_command(
     'hax',
     'Toggles your hax (makes recipes cost nothing)',
--- a/custom_commands.lua
+++ b/custom_commands.lua
@@ -605,7 +605,6 @@ end
 
 local function all_tech(cmd)
     if game.player and game.player.admin then
-game.print('entering 1')
         local confirm = cmd.parameter
         local player = game.player
         if player.force ~= game.forces['player'] then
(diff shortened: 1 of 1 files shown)


plague006:
Not WIP, just flat out forgot to remove those debugging prints

Valansch:
Not a fan of having to type "true". Do you think running the command twice isnt enough?

plague006:
Running it twice would still leave it prone to accidental use. Because the
use-case for granting the player force all techs is so niche, I want it to
be difficult to run. If you're not a fan of the command that tells me that
I'm doing it right.

Valansch:
No its okay. Actually the fact that it checks for debug mode is perfect. IMO we could even disable it outright outside of debug modus. If you don't want that than i will merge it after your response.

custom_commands.lua
```diff
@@ -603,6 +603,25 @@ local function unjail_player(cmd)
...
 end
 
+local function all_tech(cmd)
+    if game.player and game.player.admin then
+        local confirm = cmd.parameter
+        local player = game.player
+        if player.force ~= game.forces['player'] then
+            player.force.research_all_technologies()
+            game.player.print('Your force has been granted all technologies')
+        elseif player.force == game.forces['player'] and (confirm == 'true' or game.tick < 108000 or _DEBUG == true) then  -- tick 108,000=30 mins
```
Valansch: `_DEBUG == true` can just be `_DEBUG`

plague006:
Disabling this for all but debug makes a lot of sense to me. Will update
the PR later this week when I'm at home.

Valansch:
Sorry to be super picky about this PR, i don't mean to discourage you.
But ideally that command wont even be registered if not in debug mode.
This will mean that switching to _DEBUG will not make all-tech available.

But since the _DEBUG variable isn't desync save anyways, we can just implement a function to register the function when debug is set. We could do that with a metatable but im not super confident with that and don't know if its desync save.

So i propose:

```lua

_DEBUG = false
function set_debug(value) --global function
   _DEBUG = value
   if _DEBUG then
      commands.add_command('all-tech', 'researches all technologies (debug only)', all_tech)
   else
      commands.remove_command('all-tech')
   end
end
```

grilledham:
Or we could just add
```lua
if _DEBUG then
    commands.add_command('all-tech', 'researches all technologies (debug only)', all_tech)
end
```
So the command is only registered when in debug mode

Valansch:
@grilledham then we cant enable it while testing if we forgot. But if we dont mind that then obviously your solution is easier

grilledham:
True, but toggling _DEBUG mode on/off at run time opens up a whole can of worms*. Probably best to just restart the map with it enabled /disabled depending on your needs.

* For example in the same file this code would have to be re thought
```lua
if not _DEBUG then
    local old_add_command = commands.add_command
    commands.add_command =
        function(name, desc, func)
        old_add_command(
            name,
            desc,
            function(cmd)
                local success, error = pcall(func, cmd)
                if not success then
                    log(error)
                    --player_print(error) -- This casues desyncs
                    player_print('Sorry there was an error running ' .. cmd.name)
                end
            end
        )
    end
end
```

PR #204, opened 2018-08-10
info_rewrite

grilledham:
~**Not ready to merge, PR only so I can reference this**~
This is now ready for review / merge


Valansch:
I suggest we keep the fish! :/

grilledham:
We can if you want. My thinking is a `?` is a more logical button to click if you are looking for help / info

PR #205, opened 2018-08-10
Implemented /a and /report

Valansch:
Ready from my side


grilledham:
We should consider providing a way to let admins view all reports for the current map.

Also we should add a report button to the player list. Clicking the button would open a dialog where you could fill out the reason and choose to submit. That might provide a more user friendly experience / easier to discover to new users. Also some players have names that are hard to type and not all players know about tab completing names.

Valansch:
I will address what you said in another PR on a branch based on this.
Merging this as a first working version

PR #211, opened 2018-08-14
bug fixes

grilledham:
Not a bug fix but
```lua
for _,child in pairs(parent.children) do
        Gui.remove_data_recursivly(child)
        child.destroy() 
end
```
is the same functionality as `Gui.Clear(parent)`

line 64 - 66
Missing check for when /showreports is called but there are no reports to show

Removed lines 42 - 45
```lua
if player.opened then --Destroy whatever is open
    Gui.remove_data_recursivly(player.opened)
    player.opened.destroy()
end
```
This throws an error if you have an entity opened (e.g looking inside a chest).
I get what you are trying to do, close an open guiElement if one is already opened. But `player.opened` can return an entity, equipment or element, so you you would have to check the type first using `player.opened_gui_type`. 
However, you don't need to explicitly close opened elements like this, when `player.opened` is changed to something else the `on_gui_closed` event is called with the old element. For native game elements they are obviously closed correctly, for custom elements you have to close them explicitly. As long as all custom elements do that, which I find best done in `Gui.on_custom_close` (like the reports elements do) then all is good. 
So short version you should just remove this code. When `player.opened = report_frame` is called a few lines down whatever element used to be opened will be closed correctly.

Removed lines 129-132
Same reason as above

### DIFF
--- a/report.lua
+++ b/report.lua
@@ -20,10 +20,7 @@ local function draw_report(parent, report_id)
     local time_ago = Utils.format_time(game.tick - report.tick)
 
     local message = report.message
-    for _,child in pairs(parent.children) do
-        Gui.remove_data_recursivly(child)
-        child.destroy() 
-    end
+    Gui.clear(parent)
 
     parent.add {type="label", caption="Offender: " .. reported_player_name} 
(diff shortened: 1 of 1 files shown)


Valansch:
Thank you for your review, fixes and explanation. I appreciate it.

Valansch:
I'm merging into a new feature branch as i want to change the behavior of /showreports when there a no reports

PR #212, opened 2018-08-15
Report fixes

Valansch:
Few changes:

- moved show reports into custom commands
- printed that there are no reports when /showreports is run with #reports = 0.
- My old implementation would  throw errors when entities where opend, but after the fix it would throw errors when the frame itself was already open. So i fixed that.

### DIFF
--- a/report.lua
+++ b/report.lua
@@ -36,7 +36,7 @@ local function draw_report(parent, report_id)
     parent.add {type="label", caption="Reported by: " .. reporting_player_name}
 end
 
-local function show_reports(player)
+Module.show_reports = function(player)
     local reports = global.reports or {}
 
     local center = player.gui.center    
@@ -74,14 +74,14 @@ local function report(reporting_player, reported_player, message)
     local notified = false
     for _,p in pairs(game.players) do
         if p.admin and p.connected then
-            show_reports(p)
+            Module.show_reports(p)
             if p.afk_time < 3600 then notified = true end
         end
     end
     if not notified then
         for _,p in pairs(game.players) do
             if p.admin then
-                show_reports(p) 
+                Module.show_reports(p) 
             end
         end
     end
@@ -177,14 +177,6 @@ Gui.on_click(
         report(event.player, game.players[reported_player_index], msg)
         
         event.player.print("Sucessfully reported " .. game.players[reported_player_index].name)
-    end
-)
-
-commands.add_command('showreports', 'Shows user reports (Admins only)', 
-	function(event) 
-		if game.player and game.player.admin then 
-            show_reports(game.players[event.player_index]) 
-        end
     end
 )
 
--- a/report.lua
+++ b/report.lua
@@ -41,6 +41,9 @@ Module.show_reports = function(player)
 
     local center = player.gui.center    
   
+    if center[report_frame_name] then 
+        Gui.clear(center)
+    end
     local report_frame = center.add {
         type = 'frame',
         name = report_frame_name,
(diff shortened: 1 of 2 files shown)


report.lua
```diff
@@ -35,15 +36,14 @@ local function draw_report(parent, report_id)
...
-    local center = player.gui.center
-    if player.opened then --Destroy whatever is open
-        Gui.remove_data_recursivly(player.opened)
-        player.opened.destroy()
-    end
+    local center = player.gui.center    
   
+    if center[report_frame_name] then 
+        Gui.clear(center)
+    end
```
grilledham:
```lua
if center[report_frame_name] then 
    Gui.clear(center)
end
```

I don't like this. This code touches guiElements that it doesn't own. That could cause problems if the other guiElements have their own closing logic. (For example the Blueprint helper and poll GUIs on close store part of their state in global so that the player and continue where they left off when they re-open them.)

But the code needs to solve the problem of a guiElement can't have two children with the same name. So if a report_frame is already opened it needs closing.

This should do the trick.
```lua
local frame = center[report_frame_name]
if frame and frame.valid then 
    Gui.destroy(frame)
end
```

report.lua
```diff
@@ -125,12 +125,12 @@ local reporting_input_name = Gui.uid_name()
...
-    if player.opened then --Destroy whatever is open
-        Gui.remove_data_recursivly(player.opened)
-        player.opened.destroy()
-    end
+    local center = player.gui.center    
   
+
+    if center[reporting_popup_name] then 
+        Gui.clear(center)
+    end
```
grilledham: Same as above. I'd advocate replacing with
```lua
local frame = center[reporting_popup_name]
if frame and frame.valid then 
    Gui.destroy(frame)
end
```

Valansch:
You're right. Will change

Valansch:
What does it mean when a frame is invalid? 
The fact that you are checking for validity suggests it is possible for frame to not be nil and invalid.
What happends when you override an invalid frame? It may still throw this error

grilledham:
Did we remove the /report command because we added the report button to the player list? 
I personally see no harm in having both a /report command and a report button. But if we decide to not have the report command we should remove the reference to it in info.lua and tell the admins that the command is obsolete

grilledham:
Checking a LuaThing is valid before performing any action on it is just a habit I've gotten into. I probably do more valid checks than I need to, but I like to stay on the side of caution.

In this case it should impossible for a frame to be invalid then accessing directly from the parent. If you were storing your own reference to a guiElement say in global and you didn't trust that other code hasn't tampered with it then you would need to check it was valid first.

report.lua
```diff
@@ -39,12 +39,15 @@ end
...
-    local center = player.gui.center    
-  
-    if center[report_frame_name] then 
-        Gui.clear(center)
+    local center = player.gui.center
+
+    local report_frame = center[report_frame_name]
+    if report_frame and report_frame.valid then 
+        Gui.remove_data_recursivly(report_frame)
+        Gui.destroy(report_frame)
```
grilledham: `Gui.destroy(element)` is just
```lua
function Gui.destroy(element)
    Gui.remove_data_recursivly(element)
    element.destroy()
end
``` 
I wrote it so I didn't have to keep writing
```lua
 Gui.remove_data_recursivly(element)
element.destory()
```
But some of my earlier code still does that because I never updated it

Valansch:
How do i mark your review as accepted and applied?

PR #213, opened 2018-08-18
Tasklist rewrite + fruit loops changes + /report command

grilledham:
You might want to review the changes I made to bring back the /report command, just to make sure I didn't mess anything up.

Otherwise this is the version that is live on the server, so I think it's stable


Valansch:
Looking good from my end

PR #216, opened 2018-08-28
added donator to list

theorangeangle:
title.

in the future, we should consider implementing a `/donator add [username]` command to add new patreons in real time, rather than inbetween map resets. Not sure how this would affect the other perks we currently advertise on patreon (extra coins, welcome message and train invincibility) that are given ingame.

### DIFF
--- a/resources/donators.lua
+++ b/resources/donators.lua
@@ -1,4 +1,5 @@
 return {
     ['robertkruijt'] = true,
     ['aldldl'] = true
+    ['Geostyx'] = true
 }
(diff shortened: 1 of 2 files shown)


plague006:
Should open a separate issue for adding the donator command.

grilledham:
You can add donators in real time use `/silent-command global.donators['username'] = true`
However they still need to be manually added to the donators.lua file for it to be persistent, much like regulars, a limitation of prime's server.
Also donators need to be added to the regulars list as well.

theorangeangle:
ok thanks @plague006 - will do that

we just received another patreon so I have added them to the list as well - have double checked and that **is** the correct username.

PR #220, opened 2018-09-10
Implemented cave miner

Valansch:
This is a working version of the cave miner map.

Keeping this branch as is will make it possible to merge updates into the branch. (Altough there would be some merge conflicts)

But i would like to merge it into develop aswell, this way we do not need to have a seperate branch for the cave miner map. The release versions could be had in another seperate release branch.

Merging into develop will require changes to the codebase. The following changes that are required for cave miner to work, but would break the rest of the scenario:

- resources/market_items.lua
   - Replaced some of the market items
- control.lua
  - disabled some modules
  - changed some on_player_spawned items
  - altered welcome message (I think this can simply be reverted as it does not seem important for the cave miner map to not have those two messages)
  - disabled no cheat
 - nuke.control.lua
   - disabled autoban
   - disabled autokick
 - player_list.lua
   - removed coin column

All these conflicts can be resolved by creating config entries in config and loading config.lua before any other file. The map gen files could then manipulate the config as needed.
This works without conflicts under the assumtions: No conflicting map_gen modules are loaded (Those are conflicting anyways so we do not really need to worry about that) and only map_gen modules are alowed to modify the config entries.

Another thing is: I would like to move cave_miner*.lua into the map_gen/combined/cave_miner/ and move the import into map_layout.

Please share your thoughts.
@M3wM3w @grilledham 

(I will create the biter battles PR after this PR is finished)


M3wM3w:
resources/market_items.lua is not used by cave_miner anymore, you can overwrite it
the coin column does not work for any custom market.

Actually the only really necessary file change for cave_miner to run is the change of control.lua,
to add it in, fish_market.lua is not advised to add in tough because it just spawns in too many fish extra.

Also i could just add a clear inventory at the start of the game so other files don´t add in items on accident. So even if control lua adds items you still start with only the cave miner gear.

Valansch:
Perfect.

Valansch:
I applied the changes @M3wM3w @grilledham.
Ready for merge from my side.

paint.lua
```diff
@@ -40,6 +40,9 @@ local filter_table_close_button_name = Gui.uid_name()
 
 global.paint_brushes_by_player = {}
 local function player_build_tile(event)
+    if not global.scenario.config.paint.enable then 
+        return 
+    end
```
grilledham: Wouldn't a simpler solution be to stick this code at the top of the file?
Or is your thinking that we should be able to enable / disable the paint module at run time? But I don't think the current implementation would support that. If this module started of enabled, disabling it would only disabled the button at the top for new joining players.

fish_market.lua
```diff
@@ -30,6 +30,11 @@ local function spawn_market(cmd)
         cant_run(cmd.name)
         return
     end
+
+    if not global.scenario.config.fish_market.enable then 
+        game.player.print("Cannot spawn market. Market is disabled in the scenario config.")
+        return
+    end
```
grilledham: Fish market does more than provide a market. My assumption would be if `fish_market.enable` was set to false there also wouldn't be market items (fish / coins) for crafting killing biters etc, and also no fish pun messages.

Valansch:
You are right.
A better way to do it would be to unregister the event on init.

paint.lua
Valansch: If we return the require early, than the feature cannot be disabled by modules that are required after that.

However i didnt mean for the features to be disabled on runtime. Maybe i shouldn't have put them in global...

Valansch:
Fixed the fish_market still doing stuff if disabled.

PR #221, opened 2018-09-14
Initial setup for Diggy

linaori:
Features digging + cave collapses using the pressure system. It's fully integrated in RedMew now. Map can be enabled by removing the comment `--require "map_gen.combined.diggy"` in `map_layout.lua`

Help is welcome!

MVP:
- [X] Allow digging
- [X] Have cave collapsing and support beams (stone wall, concrete, stone bricks)
- [X] Spawn random ore when mining
- [x] Increase manual mining speed with mining research
- [x] Use perlin noise to open rooms
- [x] Spawn water in rooms
- [x] ~Spawn biters + spawners in rooms~ will use natural spawning mechanism for now
- [x] Spawn biters randomly when mining
- [x] Scenario introduction + tips
- [ ] Basic anti-griefing

Fixes:
- [x] Can send deconstruct orders for supportbeams
- [x] Players should not spawn with coins
- [x] Spawn rocks only once per location, collapse + mine will trigger multiple rocks at the same position
- [x] Collapses cause a corrupted stress map, could be related to not properly triggering entities it crushes?
- [x] Proper mining speed management when external sources modify it
- [x] Collapse doesn't spawn rocks 
- [x] ~Starting point sometimes contains ore~ happens rarely, can recreate the map if this happens
- [x] Rooms collapse after spawning
- [x] Cave collapsing doesn't stop when placing a support beam before the collapse is triggered
- [x] Replace `math.sqrt()` usages
- [x] players spawn at 0,0 where 2 players at once will cause them to get stuck
- [x] split warning label to 2 rows
- [ ] find a mechanic to prevent performance issues with spilled loot
- [ ] refined concrete gives infinite support, not intended but should be looked at
- [ ] disable landfil? Need to find a solution to prevent griefing for that
- [ ] slightly reduce copper spawn chances for balance?
- [ ] use normal evolution settings but adjust evolution based on bitters killed/rocks mined?
- [ ] greatly reduce oil spawns, perhaps by a factor of 10? perhaps noise based?
- [ ] slightly reduce the ore spawn chance
- [ ] slightly reduce the coal spawn
- [ ] improve market buffs
- [ ] remove void from collapses? this would prevent water from being collapsed, keeps the original tile floor and resources and makes it possible to spawn non-infinite veins of specific resources

Nice to have:
- [x] Popup when mining the first wall to alert of collapses
- [x] Use the donate button to unlock new stuff in the market
- [x] Use the donate button to unlock minor quality of life improvements (such as mining increase)
- [x] Implement a GUI element/button to show the above feature progress
- [x] Count total stone sent to the surface
- [x] Have a market where we can use stone to buy stuff
- [x] Find some way to make water spawn sooner or find a way to detect where it would spawn and hint this
- [x] Update debug grid entity colors to show -1 to 1 with 3 colors instead of 2
- [x] ~Find out how to clear the inventory of a rock so it doesn't drop on death, enables destroying rocks~ Not possible yet: [Add the ability to influence item loot in lua](https://forums.factorio.com/viewtopic.php?f=28&t=62708&p=380505#p380505)
- [ ] Void chest to increase stone sent to the surface
- [ ] A clickable alert when something collapses, paired with the sound CRACK is not enough according to players
- [x] ~have "rooms" where there is no cave (don't spawn based on noise during chunk generation?)~
- [ ] A chat message when something new is unlocked
- [ ] Scoreboard with who donated what


grilledham:
Copy of message sent to iltar in discord.

Maybe it's a bit late to say this, but I'm looking at the metatables in PressureMap.lua and I'm thinking wouldn't a hash be simpler. As an example. Should be good up to 4,194,303 tiles in +/- 
```lua
local band = bit32.band

local function new_pressure_map()
    return {}
end

local function set_value(pressure_map, x, y, value)
    local i = band(x, 0x7fffff)
    i = i + 0x800000 * band(y, 0x7fffff)
    pressure_map[i] = value
end

local function get_value(pressure_map, x, y)
    local i = band(x, 0x7fffff)
    i = i + 0x800000 * band(y, 0x7fffff)
    return pressure_map[i]
end

local function add_fraction(pressure_map, x, y, fraction)
    local i = band(x, 0x7fffff)
    i = i + 0x800000 * band(y, 0x7fffff)

    local old_value = pressure_map[i]
    local value = old_value + fraction

    pressure_map[i] = value
end
``` 
Just a suggestion

Valansch:
A hash would be simpler. However a hash key would be stored in the hash table part of the lua table. Which has an access time far worse than an array lookup. 
If we use hashes we could also just use a 2 dimensional lua table with negative values.

Unless i am missing something here?

grilledham:
With the exception of x and y = 0 where i is 0, i is positive for all values of x and y in the range [- 4,194,303, 4,194,303]. So I'd assume lua would store the values in a sparse array.

However I don't have any data measuring the performance, so I can't say which way is quicker.

Valansch:
Lua stores only spares arrays in the array part. Everything after the first gap is in the hash table part.

Valansch:
I couldn't find a reliable source so i tested it myself. I removed all key conversion code from the test by just running the same code for both the array test and the hash_table test.

Here are my result for the same size hash table and array: (4 million keys read and written 50 times)


```
Running hash_table test
Time passed (ms): 32011
Running array test
Time passed (ms): 25646
```

Thats a 25% speed increase for using the hash table. I honestly thought it would be worse, but i think 25% is enough to justfiy not using hash_tables when we dont have to, but not enough to bother for other cases where performance is less relevant.

I also tested 40k keys in 5k runs (same number of accesses) and the numbers where the same so table size really does not matter, it is only about hashing time.

Here is my test code:

```lua



local n = 4000000
local runs = 50
local array = {}
local hash_table = {}

local function set_value_hash_table(key, value)
    local i = n + key -- I dont have bit32 :/
    hash_table[i] = value
end

local function get_value_hash_table(key)
    local i = n + key
    return hash_table[i]
end

local function set_value_array(key, value)
    local i = n + key --Do the same decode calculation as the hash table functions (To reduce any other side effects)
    array[key] = value
end

local function get_value_array(key)
    local i = n + key
    return array[key]
end

local function fill()
  for i=1,n do
    array[i] = math.random(0, 1000) / 1000 --This should also get the cpu boost to ramp to max for a fair test
    set_value_hash_table(i, array[i])
  end
end

local function run_hash_table_test()
  print("Running hash_table test")
  start_t = os.clock()
  for _=1,runs do
    for i=2,n do
      local v = get_value_hash_table(i)
      set_value_hash_table(i - 1, v)
    end
  end
  end_t = os.clock()
  print("Time passed (ms): " .. (end_t - start_t) * 1000)
end


local function run_array_test()
  print("Running array test")
  start_t = os.clock()
  for _=1,runs do
    for i=2,n do
      local v = get_value_array(i)
      set_value_array(i - 1, v)
    end
  end
  end_t = os.clock()
  print("Time passed (ms): " .. (end_t - start_t) * 1000)
end

--initialize
fill()

--run tests
run_hash_table_test()
run_array_test()

```

Valansch:
For clarification: The previous test was run on lua 5.1.

Additional tests:

Lua5.2:
```
Running hash_table test
Time passed (ms): 22589
Running array test
Time passed (ms): 17152
```

Lua5.3:
```
Running hash_table test
Time passed (ms): 15292.0
Running array test
Time passed (ms): 14813.0
```

Valansch:
after changing the key to this: (It didnt make a difference)

```lua
local function set_value_hash_table(key, value)
    hash_table[-key] = value
end

local function get_value_hash_table(key)
    return hash_table[-key]
end
```

Lua5.1:
```
Running hash_table test
Time passed (ms): 31566
Running array test
Time passed (ms): 25744
```

Lua5.2:
```
Running hash_table test
Time passed (ms): 22325
Running array test
Time passed (ms): 17404
```

Lua5.3:
```
Running hash_table test
Time passed (ms): 14818.0
Running array test
Time passed (ms): 14482.0
```

Valansch:
New test results.
I tested the array implementation with metatables vs the hash table implementation using bit32.band.

4 million values written and read 10 times. (This is a whole lot slower than the previous test)

Lua5.2 (125%):
```
Running hash_table test
Time passed (ms): 18098
Running array test
Time passed (ms): 8025
```


Code:


```lua
local band = bit32.band

--CODE EXTRACTED FROM PressureMap.lua


local defaultValue = 0
local _mt_y = { __index=function(tbl,key) tbl[key] = defaultValue return tbl[key] end}
local _mt_x = {__index=function(tbl,key) tbl[key] = setmetatable({},_mt_y) return rawget(tbl,key) end}

map_a = {}
map_a[1] = setmetatable({},_mt_x) --quadrant 1
map_a[2] = setmetatable({},_mt_x) --quadrant 2
map_a[3] = setmetatable({},_mt_x) --quadrant 3
map_a[4] = setmetatable({},_mt_x) --quadrant 4


local function set_value_array(x, y, value)
  local quad = 1
  if x < 0 then
    quad = quad + 1
    x = - x
  end
  if y < 0 then
    quad = quad + 2
    y = - y
  end
  map_a[quad][x][y] = value
end

local function get_value_array(x, y, value)
  local quad = 1
  if x < 0 then
    quad = quad + 1
    x = - x
  end
  if y < 0 then
    quad = quad + 2
    y = - y
  end

  return map_a[quad][x][y]
end

----------------------------------

--Suggested change from grilled starts here


local map_h = {}

local function set_value_hash_table(x, y, value)
  local i = band(x, 0x7fffff)
  i = i + 0x800000 * band(y, 0x7fffff)
  map_h[i] = value
end

local function get_value_hash_table(x, y, key)
  local i = band(x, 0x7fffff)
  i = i + 0x800000 * band(y, 0x7fffff)
  return map_h[i]
end


-------------------------------

--Test Starts here


local min = - 1000
local max = 1000
local runs = 10

local function fill()
  for x = min + 1, max do
    for y = min + 1, max do
      local value = math.random(0, 1000) / 1000
      set_value_array(x,y, value) --This should also get the cpu boost to ramp to max for a fair test
      set_value_hash_table(x, y, value)
    end
  end
end

local function run_hash_table_test()
  print("Running hash_table test")
  start_t = os.clock()
  for _=1,runs do
    for x = min + 1, max do
      for y = min + 1, max do
        local v = get_value_hash_table(x, y)
        set_value_hash_table(x - 1, y - 1, v)
      end
    end
  end
  end_t = os.clock()
  print("Time passed (ms): " .. (end_t - start_t) * 1000)
end


local function run_array_test()
  print("Running array test")
  start_t = os.clock()
  for _=1,runs do
    for x = min + 1, max do
      for y = min + 1, max do
        local v = get_value_array(x, y)
        set_value_array(x - 1, y - 1, v)
      end
    end
  end
  end_t = os.clock()
  print("Time passed (ms): " .. (end_t - start_t) * 1000)
end

--initialize
fill()
run_hash_table_test()
run_array_test()
```

Valansch:
Ingame difference:
http://puu.sh/BxcXJ/05cf29e557.jpg

grilledham:
So on reflection I realised I made a mistake in my implementation of the direct array, so I repeated the tests.
Code used:
```lua
--metatable array

local defaultValue = 0
local _mt_y = {
    __index = function(tbl, key)
        tbl[key] = defaultValue
        return tbl[key]
    end
}
local _mt_x = {
    __index = function(tbl, key)
        tbl[key] = setmetatable({}, _mt_y)
        return rawget(tbl, key)
    end
}

map_a = {}
map_a[1] = setmetatable({}, _mt_x) --quadrant 1
map_a[2] = setmetatable({}, _mt_x) --quadrant 2
map_a[3] = setmetatable({}, _mt_x) --quadrant 3
map_a[4] = setmetatable({}, _mt_x) --quadrant 4

local function set_value_array(x, y, value)
    local quad = 1
    if x < 0 then
        quad = quad + 1
        x = -x
    end
    if y < 0 then
        quad = quad + 2
        y = -y
    end
    map_a[quad][x][y] = value
end

local function get_value_array(x, y, value)
    local quad = 1
    if x < 0 then
        quad = quad + 1
        x = -x
    end
    if y < 0 then
        quad = quad + 2
        y = -y
    end

    return map_a[quad][x][y]
end

----------------------------------
-- direct array

local map_d = {{}, {}, {}, {}}

local function set_value_array_d(x, y, value)
    local quad = 1
    if x < 0 then
        quad = quad + 1
        x = -x
    end
    if y < 0 then
        quad = quad + 2
        y = -y
    end

    local quad_t = map_d[quad]

    local x_t = quad_t[x]
    if not x_t then
        x_t = {}
        quad_t[x] = x_t
    end

    x_t[y] = value
end

local function get_value_array_d(x, y, value)
    local quad = 1
    if x < 0 then
        quad = quad + 1
        x = -x
    end
    if y < 0 then
        quad = quad + 2
        y = -y
    end

    local quad_t = map_d[quad]

    local x_t = quad_t[x]
    if not x_t then
        x_t = {}
        quad_t[x] = x_t
    end

    return x_t[y] or defaultValue
end

-------------------------------

--Test Starts here

local min = -1000
local max = 1000
local runs = 10

function fill()
    for x = min + 1, max do
        for y = min + 1, max do
            local value = math.random(0, 1000) / 1000
            set_value_array(x, y, value) --This should also get the cpu boost to ramp to max for a fair test            
            set_value_array_d(x, y, value)
        end
    end
end

function run_array_test()
    for _ = 1, runs do
        for x = min + 1, max do
            for y = min + 1, max do
                local v = get_value_array(x, y)
                set_value_array(x - 1, y - 1, v)
            end
        end
    end
end

function run_array_d_test()
    for _ = 1, runs do
        for x = min + 1, max do
            for y = min + 1, max do
                local v = get_value_array_d(x, y)
                set_value_array_d(x - 1, y - 1, v)
            end
        end
    end
end

--initialize
/c fill()
--run test:
/measured-command run_array_test()
/measured-command run_array_d_test()
```
| Test            | Time (ms) | 
|-----------------|---------|
| metatable array | 6949    |
|  direct array   | 6256    |

@Valansch If you could repeat the test on your hardware, that would make for a more fair comparison between the old tests

iltar:
Which would be better for a save game?

iltar:
Not having metatables has my vote, as it makes creating a re-usable map that we can persist between states, a lot easier.

Valansch:
Metatables are still faster and we should really take any performance bonus we can.
Whats the problem with loading the metatables in on_load?

iltar:
I hate fiddling directly with metatables, I want them hidden from usage. They are nothing but an implementation detail. I would like this detail to stay hidden, which is hard due to on_load.

grilledham:
My vote will be for whatever is faster. When I ran my latest test direct arrays were faster then metatables. Can one of you confirm my results?

Valansch:
Interesting.
@grilledham i read your results wrong i thought metatables were faster. I will try to replicate your results. But I may not get to it until tomorrow.

Valansch:
@iltar Unless you call this function itself on load or before init it is going to cause desyncs.
Why not simply put the bind() outside of the function body?
Cooldude simply has a module that hides all this. Something like this could be done:

```lua
local Metatables = 'utils.metatables'
local table_with_persistent_metatables = Metatables.create_global(table_without_persistent_metatables, metatable)
```

iltar:
I've added a feature that spawns random resources. The only things not being spawned yet are biters and water.

Valansch:
Okay. Feel free to keep adding to this PR, i will merge it as soon as i tested the meta table performance.

iltar:
Edge case to take into account when collapsing the cave, every tile below a rock should also become dirt: 
image

Valansch:
>@Valansch If you could repeat the test on your hardware, that would make for a more fair comparison between the old tests

@grilledham sorry for the long wait. I finally ran the tests:
| Test            | Time (ms) | 
|-----------------|---------|
| metatable array | 9355    |
|  direct array   | 8342 |

About 10%, which is consistent with your results.

Valansch:
New test:

```lua
--metatable array

local defaultValue = 0
local _mt_y = {
    __index = function(tbl, key)
        tbl[key] = defaultValue
        return tbl[key]
    end
}


array = setmetatable({}, _mt_y)

local function set_value_array(x, value)
  array[x] = value
end

local function get_value_array(x)
    return array[x]
end

----------------------------------
-- direct array

local array_direct = {}

local function set_value_array_d(x, value)
    array_direct[x] = value
end

local function get_value_array_d(x, value)
    return array_direct[x] or defaultValue
end

-------------------------------

--Test Starts here

local max = 1500000
local runs = 20

function fill()
    for x = 0, max do
      local value = math.random(0, 1000) / 1000
      set_value_array(x, y, value) --This should also get the cpu boost to ramp to max for a fair test
      set_value_array_d(x, y, value)
    end
end

function run_array_test()
    for _ = 1, runs do
        for x = 0 + 1, max do
          local v = get_value_array(x)
          set_value_array(x - 1, v)
        end
    end
end

function run_array_d_test()
    for _ = 1, runs do
        for x = 0 + 1, max do
          local v = get_value_array_d(x)
          set_value_array_d(x - 1, v)
        end
    end
end

--initialize
--/c fill()
--run test:
--/measured-command run_array_test()
--/measured-command run_array_d_test()

```

Valansch:
Okay so as cooldude said on discord the tests were faulty because the direct version didnt set the default value.

Valansch:
after making this change:

```lua
    local v = x_t[y]
    if v then return v end

    x_t[y] = defaultValue
    return defaultValue
```

The results are:
| Test            | Time (Fully initialized) | Time (Half initialized) 
|-----------------|---------|------|
| metatable array | 9165  |     10050      |
|  direct array       | 8314    |    8567     |

grilledham:
True, the tests weren't equivalent code, however the part where the code wasn't comparable doesn't run doing the test as we only lookup indexes that are already set. So it probably didn't make a different.

Valansch:
```lua

--metatable array

local defaultValue = 0
local _mt_y = {
    __index = function(tbl, key)
        tbl[key] = defaultValue
        return defaultValue
    end
}
local _mt_x = {
    __index = function(tbl, key)
        tbl[key] = setmetatable({}, _mt_y)
        return rawget(tbl, key)
    end
}

map_a = {}
map_a[1] = setmetatable({}, _mt_x) --quadrant 1
map_a[2] = setmetatable({}, _mt_x) --quadrant 2
map_a[3] = setmetatable({}, _mt_x) --quadrant 3
map_a[4] = setmetatable({}, _mt_x) --quadrant 4

local function set_value_array(x, y, value)
    local quad = 1
    if x < 0 then
        quad = quad + 1
        x = -x
    end
    if y < 0 then
        quad = quad + 2
        y = -y
    end
    map_a[quad][x][y] = value
end

local function get_value_array(x, y, value)
    local quad = 1
    if x < 0 then
        quad = quad + 1
        x = -x
    end
    if y < 0 then
        quad = quad + 2
        y = -y
    end

    return map_a[quad][x][y]
end

----------------------------------
-- direct array

local map_d = {{}, {}, {}, {}}

local function set_value_array_d(x, y, value)
    local quad = 1
    if x < 0 then
        quad = quad + 1
        x = -x
    end
    if y < 0 then
        quad = quad + 2
        y = -y
    end

    local quad_t = map_d[quad]

    local x_t = quad_t[x]
    if not x_t then
        x_t = {}
        quad_t[x] = x_t
    end

    x_t[y] = value
end

local function get_value_array_d(x, y, value)
    local quad = 1
    if x < 0 then
        quad = quad + 1
        x = -x
    end
    if y < 0 then
        quad = quad + 2
        y = -y
    end

    local quad_t = map_d[quad]

    local x_t = quad_t[x]
    if not x_t then
        x_t = {}
        quad_t[x] = x_t
    end

    local v = x_t[y]
    if v then return v end

    x_t[y] = defaultValue
    return defaultValue
end

-------------------------------

--Test Starts here

local min = -1000
local max = 1000
local runs = 10

function fill()
    for x = min + 1, max do
        for y = min + 1, max do
            local value = math.random(0, 1000) / 1000
            set_value_array(x, y, value) --This should also get the cpu boost to ramp to max for a fair test
            set_value_array_d(x, y, value)
        end
    end
end

function run_array_test()
    for _ = 1, runs do
        for x = min + 1, max do
            for y = min + 1, max do
                local v = get_value_array(x, y)
                set_value_array(x - 1, y - 1, v)
            end
        end
    end
end

function run_array_d_test()
    for _ = 1, runs do
        for x = min + 1, max do
            for y = min + 1, max do
                local v = get_value_array_d(x, y)
                set_value_array_d(x - 1, y - 1, v)
            end
        end
    end
end

--initialize
--/c fill()
--run test:
--/measured-command run_array_test()
--/measured-command run_array_d_test()

```

PR #222, opened 2018-09-22
Game table

grilledham:
So it turns out that replacing the game.players table is a bad idea. I propose making a Game table and using Game.players in place of game.players to fix the player names that are numbers issue.

**I have only tested this in single player and have not tried changing my name to a number. I would need a second player to properly test this.**


Valansch:
Why the global Game? why not just local Game = require "utils.Game"?

Valansch:
I like this.

PR #226, opened 2018-10-02
Removed several math.sqrt in codebase

Valansch:
Since i was about to remove math.sqrt from the diggy code i took the liberty to do the same to the rest of the codebase.

I couldn't remove all occurrences, of it. When the sqrt is needed in a sum or the result is passed into another function i didnt touch it.

I hope i didnt break anything.
If you feel that readabily is more important in some instances feel free to reset the files in question.

### DIFF
--- a/map_gen/ores/tiny_ores.lua
+++ b/map_gen/ores/tiny_ores.lua
@@ -20,8 +20,8 @@ Global.register_init(
 
 local function value(base, mult, pow)
     return function(x, y)
-        local d = math.sqrt(x * x + y * y)
-        return base + mult * d ^ pow
+        local d_sq = x * x + y * y
+        return base + mult * d ^ (pow / 2)
     end
 end
 
--- a/map_gen/ores/tiny_ores.lua
+++ b/map_gen/ores/tiny_ores.lua
@@ -21,7 +21,7 @@ Global.register_init(
 local function value(base, mult, pow)
     return function(x, y)
         local d_sq = x * x + y * y
-        return base + mult * d ^ (pow / 2)
+        return base + mult * d_sq ^ (pow / 2)
     end
 end
 
--- a/map_gen/presets/lines_and_balls.lua
+++ b/map_gen/presets/lines_and_balls.lua
@@ -1,5 +1,6 @@
 local b = require 'map_gen.shared.builders'
 local Random = require 'map_gen.shared.random'
+local math = require "utils.math"
 
 local seed1 = 17000
 local seed2 = seed1 * 2
@@ -97,8 +98,6 @@ end
 
 local map_ores = do_resources()
 
-local root2 = math.sqrt(2)
-
 local big_circle = b.circle(150)
 local small_circle = b.circle(140)
 local crop = b.rectangle(300, 150)
@@ -121,8 +120,9 @@ local function h_arc()
     return b.any {arc, ball1, ball2, ball3, arm1, arm2, arm3}
 end
 
+local div_100_sqrt2 = 100 / math.sqrt2
 local function v_arc()
-    local circle = b.circle(100 / root2)
+    local circle = b.circle(div_100_sqrt2)
     circle = b.apply_entity(circle, map_ores)
     local ball1 = b.translate(circle, -0, 385)
     local ball2 = b.translate(circle, -460, 345)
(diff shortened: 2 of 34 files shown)


map_gen/ores/tiny_ores.lua
```diff
@@ -20,8 +20,8 @@ Global.register_init(
 
 local function value(base, mult, pow)
     return function(x, y)
-        local d = math.sqrt(x * x + y * y)
-        return base + mult * d ^ pow
+        local d_sq = x * x + y * y
+        return base + mult * d ^ (pow / 2)
     end
```
grilledham: This change is really nice, I hadn't thought of that.
Also it's funny how many times I declare this exact same function because I normally copy code from previous presets.
I later on realised that and switched to using `local value = b.exponential_value` but couldn't be bothered to change old code

utils/math.lua
```diff
@@ -5,7 +5,10 @@ math.sin = function(x)
      return math.floor(_sin(x) * 10000000 + 0.5) / 10000000
 end
 
-
 math.cos = function(x)
      return math.floor(_cos(x) * 10000000 + 0.5) / 10000000
 end
+
+math.sqrt2 = math.sqrt(2)
+
```
grilledham: if we are adding constants to math I would like `math.inv_sqrt2 = 1 / math.sqrt(2)`

map_gen/presets/lines_and_balls.lua
```diff
@@ -140,7 +140,7 @@ arc1 = b.single_pattern(arc1, 1380, 1380)
 local arc2 = v_arc()
 arc2 = b.single_pattern(arc2, 1380, 1380)
 arc2 = b.rotate(arc2, degrees(45))
-arc2 = b.scale(arc2, root2, root2)
+arc2 = b.scale(arc2, math.sqrt2, math.sqrt2)
 arc2 = b.translate(arc2, -165, -688)
```
grilledham: I don't think these code changes make any difference as the value was only calculated when the script is loaded, then at run-time the cached value is used. I don't think it affects readability though, so there is no need to change it back.

Valansch: I wanted it to consistent, I just forgot to ask:
Do you prefere root2 or sqrt2?

map_gen/ores/tiny_ores.lua
Valansch: Fixed the typo.

map_gen/presets/lines_and_balls.lua
Valansch: Also i would not rely on math being required by any other module.
Although i overwrite math itself, it should always be required by the module that uses the new function.
Unless we plan on just leaving it in control.lua. It's currently only there to fix math.sin.

grilledham: sqrt2 works for me.

We currently only require utils.list_utils in control.lua despite it being used in multiple places.
If utils.math only makes changes to the math table my vote would be to just leave it in control.

But this would be an exception to our rule about explicit requires, so maybe we should think about this.

grilledham:
There are two function in `utils.utils` that might make more sense in `utils.math`
```lua
-- rounds number (num) to certain number of decimal places (idp)
math.round = function(num, idp)
    local mult = 10 ^ (idp or 0)
    return math.floor(num * mult + 0.5) / mult
end

function math.clamp(num, min, max)
    if num < min then
        return min
    elseif num > max then
        return max
    else
        return num
    end
end
```

Valansch:
I did not realize there are other math functions in utils. Yes they should be moved.

About the require thing: Well tell me when you made a decision :D I don't mind either way but it should be consistent.

grilledham:
OK I've thought about it. For code sharing I really like the idea that each module explicitly states it's dependencies, so an implementer know what other modules they need. So we should explicitly require `utils.utils.` `utils.list_utils` and `utils.math` in all places that need them. 

It looks like the only reason `utils.utils` is required in control.lua is because it makes changes to the math table. Any module that uses other function already explicitly requires it, so we just need to remove `require utils.utils` from control.lua.

`utils.math` needs to be required in any module that uses `math.sin` or `math.cos`

For `utils.list_utils` I think the only function i use is `table.binary_search` which probably about half of the presets use.

Maybe this is a topic for different conversation, but I don't really like the implementation of `utils.list_utils`. But the issues that is most relevant for this to this PR is the name, list makes me think the function only operate on the array part of the table not the dictionary part, but they are mixed. I would suggest renaming it `utils.table`.

Valansch:
I changed all the points you addressed.
About list_utils.lua. I also think this is another conversation. Feel free to open an issue :P
But if we touch list_utils we should redo it complete. Because i think a metatable implementation would be better here.

PR #228, opened 2018-10-08
Only add the market item when market is enabled

linaori:
This ensures that coins are not added to the inventory unless the fish market is actually enabled.


grilledham:
This is simple, so I'll merge it.

Issue #229, opened 2018-10-14
Diggy Scenario - TODO list after first release


iltar:
Here's a list of proposed features/fixes that are not implemented yet.

#### Basic anti-griefing
There's currently no anti-griefing and it's easy to grief. Suggestions are highly welcome!

- [x] done

####  Void chest to increase stone sent to the surface

- [x] done

#### Find a proper biter evolution curve
Our public run had x2 time and x4 pollution, which resulted in about 94% in 18h playtime. 

_Proposed solution:_ leave the evolution settings default, but base it on biters liberated + void tiles removed, if we can modify evolution dynamically in-game. Otherwise, start with x5 time and x2 pollution (or 1.5x if possible).

- [x] done

#### Scoreboard with who donated what, biters killed and void removed

- [x] done

#### Make all non void tiles only affect themselves.
- [x] done

#### (Refined) concrete gives infinite support, not intended but should be looked at

- [x] done

#### A clickable alert when something collapses

- [x] done

#### Find a mechanic to prevent performance issues with spilled loot
There's a serious performance drop when loot is spilled but there's either no place or no place nearby to spawn the loot. The game will grind to a halt.

_Proposed solution_: Check for 20 (average) nearby stones on the floor and delete them when a rock dies, to ensure space nearby. This should greatly improve performance.

- [x] done

#### Disable landfill
There's no good reason to have it. Water is scarce and landfill causes easy griefing

- [x] done

#### Rebalance resources on the map
Current map has far too much copper and coal compared to the rest.

_Proposed solutions_:
 - Change percentages to increase iron spawn by a lot, reduce copper by a lot and reduce coal by some
 - Reduce oil spawn chance by 10 fold
 - Reduce max spawn chance slightly
 - Reduce the minimum spawn chance slightly

- [x] done 

#### Improve market buffs
Currently electric poles are a bit expensive, and we might have to look into adding more utility in here (repair packs for example). We should also add new buffs beyond 500k

- [x] done

#### Remove void tiles from collapses
This would prevent water from being collapsed (less griefing), keeps the original tile floor and resources and makes it possible to spawn non-infinite veins of specific resources. 

- [x] done

#### A chat message when something new is unlocked from the market

 - [x] done

#### To improve serialization time reduce stress map resolution to 2x2 tiles per pixel
 - [x] done

#### Fix player starting position, players get teleported on each other.
 
 - [x] done

iltar:
Closing this issue as most points are complete and the last one is in the works

PR #230, opened 2018-10-15
Collapse changes

Valansch:
- Void will no longer spawn on collapse
- Nerfed tiles significantly
- Implemented /test-tile-support-range
- removed on_void_added event
- made tiles not support neighbors


Valansch:
#229

map_gen/Diggy/Feature/DiggyCaveCollapse.lua
```diff
@@ -79,60 +79,32 @@ DiggyCaveCollapse.events = {
...
+    for _, position in pairs(positions) do
+        local x = position.x
+        local y = position.y
+        insert(entities, {position = {x = x, y = y}, name = 'sand-rock-big'})
+
+        for _, entity in pairs(find_entities_filtered({position = position})) do
+            pcall(function()
+                local strength = support_beam_entities[entity.name]
+                local position = entity.position
+                entity.die()
```
iltar: Wasn't there a plan to damage instead of destroy? Can also do this at a later stage.

Valansch: There was but i ditched it. The only entity that would not be destroyed by damage() was walls because those are the only ones that have support value. 
Which makes it kinda silly imo.
I can change this at any time but i wasn't really sure on what would be best so i just kept die().

Also using damage() means we have to check if the entity actually died or not (in case it had strength) and fire the event. (I think?)

I'm not opposed to the idea of damage() but I think we need to discuss the how first. So yea lets do it at a later stage

PR #239, opened 2018-10-16
Changed stress map resolution

Valansch:
Can probably need more testing but it seemed to work fine.

It can also be optimized because the conversion happends at a very low level. But i had trouble that so this is a working solution for now.

### DIFF
--- a/map_gen/Diggy/Debug.lua
+++ b/map_gen/Diggy/Debug.lua
@@ -77,10 +77,13 @@ end
     @param surface LuaSurface
     @param position Position {x, y}
 ]]
-function Debug.print_grid_value(value, surface, position)
-    local r = max(1, value) / 4
-    local g = 1 - abs(value) / 4
-    local b = min(1, value) / 4
+function Debug.print_grid_value(value, surface, position, scale, offset)
+    local scale = scale or 1
+    local offset = offset or 0
+    local position = {x = position.x + offset, y = position.y + offset}
+    local r = max(1, value) / scale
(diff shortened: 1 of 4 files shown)


map_gen/Diggy/Debug.lua
```diff
@@ -78,9 +78,9 @@ end
     @param position Position {x, y}
 ]]
 function Debug.print_grid_value(value, surface, position)
-    local r = max(1, value)
-    local g = 1 - abs(value)
-    local b = min(1, value)
+    local r = max(1, value) / 4
+    local g = 1 - abs(value) / 4
+    local b = min(1, value) / 4
```
iltar: This will break noise values when debugging, can you move the `/ 4` to the calling location instead?

map_gen/Diggy/Feature/DiggyCaveCollapse.lua
```diff
@@ -85,18 +85,22 @@ local function create_collapse_template(positions, surface)
...
-                    stress_map_add(surface, position, strength)
+                    do_insert = false
+                else
+                    local position = entity.position
+                    entity.die()
                 end
             end)
         end
+        if do_insert then
+        insert(entities, {position = {x = x, y = y}, name = 'sand-rock-big'})
```
iltar: indent

map_gen/Diggy/Debug.lua
Valansch: Thats true. Didnt think of noise.

PR #241, opened 2018-10-16
Reduced biter spawn cache footprint and digging increases evo

linaori:
Opening 100x100 tiles would mean 0.8% evolution increase. Is this a correct value? I feel like this would scale a lot less aggressive than the settings we had for the first open run.

I've also fixed the cache to only store the last evolution factor, but this only works if evolution increases, never decreases. This should fix a "memory leak" when spawning biters as evolution values were a float but never rounded. This meant that it was stored uniquely anyway.

See #229

### DIFF
--- a/map_gen/Diggy/AlienEvolutionProgress.lua
+++ b/map_gen/Diggy/AlienEvolutionProgress.lua
@@ -6,13 +6,20 @@
 -- dependencies
 local Global = require 'utils.global'
 local random = math.random
+local round = math.round
 
 -- this
 local AlienEvolutionProgress = {}
 
 local alien_cache = {
-    biters = {},
-    spitters = {},
+    biters = {
+        evolution = -1,
+        cache = {},
+    },
+    spitters = {
+        evolution = -1,
+        cache = {},
+    },
 }
 
 Global.register({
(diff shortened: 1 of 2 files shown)


iltar:
Instead of creating a giant table containing all data ever used, it will now only keep track of the highest evolution reached, rounded to a full percentage, as more precision is not going to change enough.

Previously the cache would look like this:

```lua
{
    42.87323432 = {
        "small-biter": 0.6,
        "medium-biter": 0.4,
    },
    42.8735234 = {
        "small-biter": 0.6,
        "medium-biter": 0.4,
    },
    42.88256234 = {
        "small-biter": 0.6,
        "medium-biter": 0.4,
    },
    42.9127367341 = {
        "small-biter": 0.6,
        "medium-biter": 0.4,
    },
    43.0000023172 = {
        "small-biter": 0.55,
        "medium-biter": 0.45,
    },
    43.001001423172 = {
        "small-biter": 0.55,
        "medium-biter": 0.45,
    },
}
```

Now it will look like this:
```lua
biters = {
    evolution = 42,
    cache = {
        "small-biter": 0.6,
        "medium-biter": 0.4,
    },
}
```

When updated, it will look like this:
when evo changes to a certain range, it's updated to
```lua
biters = {
    evolution = 43,
    cache = {
        "small-biter": 0.55,
        "medium-biter": 0.45,
    },
}
```

Valansch (approved):

Valansch:
Great :+1:

PR #242, opened 2018-10-17
Added stone void chest that sends stone to surface

linaori:
See #229

### DIFF
--- a/map_gen/Diggy/Config.lua
+++ b/map_gen/Diggy/Config.lua
@@ -179,7 +179,7 @@ local Config = {
             },
 
             -- every x ticks it will clear y currency_item
-            void_chest_frequency = 300,
+            void_chest_frequency = 307,
 
             unlockables = {
                 {stone = 50, type = 'buff', prototype = {name = 'mining_speed', value = 10}},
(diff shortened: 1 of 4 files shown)


map_gen/Diggy/Config.lua
```diff
@@ -173,114 +173,173 @@ local Config = {
             stone_to_surface_amount = 50,
             currency_item = 'stone',
 
+            -- locations where chests will be automatically cleared from currency_item
+            void_chest_tiles = {
+                {x = -1, y = 5}, {x = 0, y = 5}, {x = 1, y = 5},
+            },
+
+            -- every x ticks it will clear y currency_item
+            void_chest_frequency = 300,
```
Valansch: I suggest using 307 instead. Or am i missing something and its important that its exactly 300 ticks?
307 is the closest prime to 300, which will minimize the collisions with other on_nth tick handlers. Helping to keep the workload across ticks constant.

Very minor detail, just a thing i like to do.

300 is going to collide with the following on_nth_tick handlers: 2, 3, 4, 5, 6, 10, 12, 15, 20, 25, 30, 50, 60, 75, 100, 150

Valansch: Those all have the same factors. You want "round" numbers i want "not so round" numbers.
Something like 307 would be close enough to 5 seconds for me to just say "its 5 seconds". But if you disagree then you can just merge it.

iltar: if they all suffer from the same problem, I'll change it to 307, performance is a thing I care about

PR #248, opened 2018-10-21
Balance changes and fixes based on the second play

linaori:
This contains a list of small fixes based on the previous run. The bigger things I will create a new issue for.

Changelog
 - Speed modifier in cheats
 - `/clear-void` command to open a big portion of void
 - Fixed a bug causing 70% instead of intended max resource density
 - Greatly reduced min and max ore density
 - Overall lowered resource spawn
 - Added a shotgun in the market
 - Added more market rewards
 - Fixed a bug causing the log to fill with errors, removed the offending dead code
 - Mining speed from the market is now 5 instead of 2 for the majority
 - Mining speed from mining productivity is now 5 per level instead of 10
 - Reduced pollution evolution factor by x15
 - Fixed the bug causing only 1/4th of the starting position to be cleared of resources

### DIFF
--- a/map_gen/Diggy/Config.lua
+++ b/map_gen/Diggy/Config.lua
@@ -26,10 +26,14 @@ local Config = {
             cheats = {
                 manual_mining_speed_modifier = 1000,
                 character_inventory_slots_bonus = 1000,
+                character_running_speed_modifier = 2,
             },
         },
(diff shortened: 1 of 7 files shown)


Valansch (approved):
Looking very good

PR #250, opened 2018-10-22
Add the ability to lock players into cars

plague006:
Just a dumb thought I had, no idea if it'll ever be used. :smile:


iltar:
Like an actual Jail? Should make a parking lot 😂

Valansch:
No. I didint get it at first either. Until i tried it. You no longer have a player character, your body is the car. Every player that is. Nobody can leave the cars. Imagine 30 cars building a factory. :D

Valansch:
Tested it in my rocket league map and it seems to work as intended :)

PR #252, opened 2018-10-22
Initial merge of biter battles by mewmew

plague006:
Credit to @M3wM3w for the code.

This PR the first step toward integrating the scenario: getting the files themselves into the codebase so that it can be worked on collaboratively and begin a commit history.


Valansch:
I changed the base to develop as develop should only have clean, working versions.
When you are done with the merge i suggest rebasing to an earlier commit, so the history doesnt contain so many "Added new file" --> "Removed file" changes.

PR #253, opened 2018-10-22
Updated readme to contain more information

linaori:
This should explain most of the things people wanted to know from various sources. Feedback welcome!


Akotsan:
Помогите! Как включить осыпание! А то рою оно не осыпается в одиночке.

plague006:
Are we going back to in-repo docs? I thought we were doing wiki for easier maintenance and lower barriers for contributions?

PR #255, opened 2018-10-23
Working import for Biter Battles

linaori:
It's rather untested, but most basic stuff seems to work. Fixed a lot of performance related stuff.


map_gen/combined/Biter_Battles/biter_battles.lua
```diff
@@ -0,0 +1,1250 @@
...
+        for _, t in pairs (global.biter_building_inhabitants[e]) do
+            for x = 1, random(t[2],t[3]), 1 do
+                local p = global.battle_surface.find_non_colliding_position(t[1] , event.entity.position, 6, 1)
+                if p then global.battle_surface.create_entity {name=t[1], position=p} end
+            end
+        end
+    end
+end
+
+local function get_valid_biters(requested_amount, y_modifier, pos_x, pos_y, radius_inc)
```
Klonan: This function seems excessively heavy,

It might be better to do more controlled `find_entities_filtered` on chunks and spiral outward from the starting chunk position,
That way you can avoid checking `entity.position > 0` by making sure the area is searches is correct,
Also then you will only need to add biters to the `valid_biters` array once

map_gen/combined/Biter_Battles/biter_battles.lua
```diff
@@ -0,0 +1,1250 @@
...
+        if global.biter_battles_debug then
+            game.players[1].print(counted_biters .. " valid biters found.")
+            game.players[1].print(#biters_selected_for_attack .. " biter going for a spread attack")
+        end
+    end
+
+    if mode == "line" then
+        local valid_biters = get_valid_biters(requested_amount, y_modifier, 0, 150*y_modifier, 500)
+        if #valid_biters < requested_amount then
+            valid_biters = get_valid_biters(requested_amount, y_modifier, 0, 1500*y_modifier, 500)
```
Klonan: You could optimize out this extra check if `get_valid_biters` would always return at least the requested amount

map_gen/combined/Biter_Battles/biter_battles.lua
```diff
@@ -0,0 +1,1250 @@
...
+        local f = floor(counted_biters/requested_amount,0)
+        if f >= 2 then
+            array_start = requested_amount * random(1,f-1)
+            if random(1,f) == 1 then array_start = 1 end
+        end
+        local x = 0
+        for y = array_start, counted_biters, 1 do
+            x = x + 1
+            if not valid_biters[y] then break end
+            if #biters_selected_for_attack >= requested_amount then break end
```
Klonan: Checking length every iteration is quite expensive, especially on larger arrays,

This whole loop could be removed if the `get_valid_biters` returned what you need in the first place

map_gen/combined/Biter_Battles/biter_battles.lua
iltar: Would it make sense to search 1 chunk per tick and aggregate?

mheguy:
Pretty sure we've all more or less agreed to drop BB outright since mew is working on it. If anyone objects, feel free to reopen.

PR #264, opened 2018-11-04
Update donators.lua

mheguy:
I'll try to avoid taking it personally that nobody added me as a donator even though I've been donating since September 13. -_-


grilledham:
Adding yourself is an initiative test.

PR #267, opened 2018-11-05
Added noise based cluster resources

linaori:
Adds noise based clustered resource spawning

image

### DIFF
--- a/map_gen/Diggy/Feature/SimpleRoomGenerator.lua
+++ b/map_gen/Diggy/Feature/SimpleRoomGenerator.lua
@@ -6,12 +6,21 @@
 local Template = require 'map_gen.Diggy.Template'
 local Perlin = require 'map_gen.shared.perlin_noise'
 local Event = require 'utils.event'
+local Global = require 'utils.global'
 local Debug = require'map_gen.Diggy.Debug'
 local Task = require 'utils.Task'
 local Token = require 'utils.global_token'
 
 -- this
 local SimpleRoomGenerator = {}
+
+local seed
+
+Global.register({
+    seed = seed,
+}, function(tbl)
+    seed = tbl.seed
+end)
 
 local do_spawn_tile = Token.register(function(params)
     Template.insert(params.surface, {params.tile}, {})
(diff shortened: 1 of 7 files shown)


plague006:
Comments! Yes!

map_gen/Diggy/Feature/SimpleRoomGenerator.lua
```diff
@@ -55,7 +54,7 @@ function SimpleRoomGenerator.register(config)
     local room_noise_minimum_distance_sq = config.room_noise_minimum_distance * config.room_noise_minimum_distance
 
     local function get_noise(surface, x, y)
-        local seed = surface.map_gen_settings.seed + surface.index
+        local seed = surface.map_gen_settings.seed + surface.index + 200
         return Perlin.noise(x * config.noise_variance, y * config.noise_variance, seed)
```
grilledham: I have found `surface.map_gen_settings` to be an expensive call in the past.
I recommend caching the seed. Below is something I've used in the past for that.
```lua
local seed = nil
Global.register_init(
    {},
    function(tbl)
        tbl.seed = game.surfaces[1].map_gen_settings.seed
    end,
    function(tbl)
        seed = tbl.seed
    end
)
```

iltar: This will work, but only because the seed is the same across surfaces, I'll see if I can make this work based on the surface index (which should be less heavy). Luckily this part isn't a big issue, but as I use this function elsewhere, I will have to do this in 2 places, in 2 globals :(

iltar: Actually I want to store the entire seed in a global for the noise function, meaning I can do a bit more math on it to make it less likely to collide (see current +100 and +200).

iltar:
@grilledham I've fixed the caching of that seed, which should increase performance a bit at the cost of another global. Verified against `/toggle-heavy-mode` that it will not cause a desync.

grilledham:
For SimpleRoomGenerator.lua and ScatteredResources.lua I'm pretty sure the code
```lua
Global.register({
    seed = seed,
}, function(tbl)
    seed = tbl.seed
end)
```
does nothing useful. The line
`seed = seed or surface.map_gen_settings.seed + surface.index + 100`
Is sufficient for seed caching, so I'd remove the Global code.

The comment I gave earlier cached the seed at init instead, but this way is fine too.

iltar:
You're correct, it doesn't desync if I remove the global. I've moved the local to the register function as there's no need to have it available out of that scope

PR #268, opened 2018-11-06
removed some _G entries

Valansch:


config.lua
```diff
@@ -1,4 +1,4 @@
-_DEBUG = false
+_DEBUG = true
```
iltar: I don't think this should be committed

utils/game.lua
```diff
@@ -39,4 +39,12 @@ function Game.get_player_by_index(index)
     end
 end
 
+function Game.player_print(str)
+    if game.player then
```
iltar: > The player typing at the console - nil in all other instances. See `LuaGameScript::players` for accessing all players. 

https://lua-api.factorio.com/latest/LuaGameScript.html#LuaGameScript.player

Valansch: This is the desired behavior.

grilledham:
I haven't tested it, but assuming nothing breaks I approve.

PR #272, opened 2018-11-09
Made the documentation clear for single player/LAN

iltar:
Who needs documentation right? I just want to play this awesome redmew map!

This small change should hopefully guide players in the right direction to look for a self hosted or single player scenario.


plague006:
👍

PR #278, opened 2018-11-10
Added bot ore islands, use bool flag to turn on.

Jayefuu:


map_gen/presets/fractal_balls.lua
```diff
@@ -19,6 +19,9 @@ local function no_resources(x, y, world, tile)
     return tile
 end
 
+-- bot_islands_flag true if you want to add islands of ores only reachable by robots
+local bot_islands_flag = false;
```
grilledham: I like the bot islands, let's remove the flag and have them on by default. If someone doesn't want them they can just comment out the code below.

Jayefuu: Can do, but I thought a single value to change at the top made it clearer for someone who wants to generate their own map. Otherwise it's not clear which lines to comment to remove the bot islands. Let me know either way and I'll make the change.

grilledham: OK we can keep the flag. Do you want to change it to true by default?

PR #288, opened 2018-11-10
Split colors resources from colors code & catchup on donators

plague006:
Also a catchup on donators, their messages, and colors (which I know should be a distinct branch and PR, I'm just really lazy today)


Valansch:
Nice

PR #289, opened 2018-11-10
Added a "comeptitive" system, which lets Diggy "matches" compete

linaori:
Added features:
 - A scoreboard, still needs a GUI (ping @der-dave 😄)
 - Biters drop coins now
 - Mining rocks can drop coins now
 - Chests spawn with randomized contents, but always coins
 - Launching a rocket with coins will now add it to the high score

As a miner you've been sent underground to collect artefacts (coins) and send them into space. (Lore = Lore, don't hate me).

### DIFF
--- a/map_gen/Diggy/Config.lua
+++ b/map_gen/Diggy/Config.lua
@@ -31,6 +31,11 @@ local Config = {
                 manual_mining_speed_modifier = 1000,
                 character_inventory_slots_bonus = 1000,
                 character_running_speed_modifier = 2,
+                starting_items = {
+                    {name = 'modular-armor', count = 1},
+                    {name = 'submachine-gun', count = 1},
+                    {name = 'uranium-rounds-magazine', count = 200},
+                },
             },
         },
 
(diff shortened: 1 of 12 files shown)


der-dave:
:ok_hand: I will have a look and contact you

grilledham:
I haven't tested it but it looks good.

Only one minor suggestion.
In some place like AlienSpanwer.lua line 43 and ScatteredResources.lua line 85 you have this
```lua
local x = event.position.x
local y = event.position.y
```
Where as in ArtefactHunting.lua line 41, DiggyCaveCollapse.lua line 170 and SimpleRoomGenerator.lua line 62 you have this
```lua
local position = event.position
local x = position.x
local y = position.y
```
I prefer the second way of doing it, as each time you call `event.position` you get a copy of the position table, so the second way is slightly more performant.

iltar:
Ah yes, fair point, must've missed that

map_gen/Diggy/Config.lua
```diff
@@ -30,6 +30,11 @@ local Config = {
                 manual_mining_speed_modifier = 1000,
                 character_inventory_slots_bonus = 1000,
                 character_running_speed_modifier = 2,
+                starting_items = {
```
plague006: I love this idea. I was trying to think of where to put the start items in the base scenario instead of control: config is the right call.

iltar: This is specifically for the cheat mode though, scroll a bit down in this file to see the default items for normal mode (done in SetupPlayer).

iltar:
Merging this so I can continue with some follow-up fixes that require some changes from where

PR #293, opened 2018-11-11
Split chat triggers from control

plague006:
This also introduces the concept of the `features` folder where we store our default/generic scenario code.


features/chat_triggers.lua
```diff
@@ -0,0 +1,121 @@
...
+    {'Hodor! Hodor! Hodor! Hodor!', 4},
+    {'Hodor :(', 4},
+    {'Hodor :)', 4},
+    {'HOOOODOOOR!', 4},
+    {'( ͡° ͜ʖ ͡°)', 1},
+    {'☉ ‿ ⚆', 1}
+}
+
+local auto_responses = {
+    {trigger = 'discord', response1 = 'Did you ask about our discord server?', response2 = 'You can find it here: redmew.com/discord'},
```
iltar: perhaps not completely within scope of this change, but could change the responses to be a table and iterate over them instead

features/chat_triggers.lua
```diff
@@ -0,0 +1,121 @@
...
+        for word in message:gmatch('%S+') do
+            if naughty_words[word] then
+                game.print(player.name .. ' this is a Christian Factorio server, no swearing please!')
+                break
+            end
+        end
+    end
+end
+
+Event.add(defines.events.on_console_chat, hodor)
```
iltar: Missing final line break at the end of the file:
image

features/chat_triggers.lua
```diff
@@ -0,0 +1,121 @@
...
+                end
+            end
+        end
+    end
+
+    if global.naughty_words_enabled then
+        local naughty_words = global.naughty_words
+        for word in message:gmatch('%S+') do
+            if naughty_words[word] then
+                game.print(player.name .. ' this is a Christian Factorio server, no swearing please!')
```
iltar: Is it really a Christian server? Last time I checked, hardware wasn't religious unless sprinkled by holy water 😆 

Perhaps something for another time: randomized messages? Religious preferences in games are always a good way to fuel a flame war.

features/chat_triggers.lua
```diff
@@ -0,0 +1,121 @@
...
+    end
+
+    -- player_index is nil if the message came from the server,
+    -- and indexing Game.players with nil is apparently an error.
+    local player_index = event.player_index
+    if not player_index then
+        return
+    end
+
+    local player = Game.get_player_by_index(event.player_index)
```
iltar: Wouldn't it make more sense to have the nil check integrated inside the `get_player_by_index`?
```diff
function Game.get_player_by_index(index)
+    if not index then
+        return
+    end
+
    local p = game.players[index]
```

features/chat_triggers.lua
plague006: Done

features/chat_triggers.lua
plague006: In this case it would be within the scope since I made the change from individual `if`s to iterating over the `auto_responses` table.

features/chat_triggers.lua
plague006: Moved to #306

features/chat_triggers.lua
plague006: Resolved in fed64c4

features/chat_triggers.lua
plague006: I think this falls outside the scope of this PR.

PR #294, opened 2018-11-11
Add Diagonal Ribbon to map presets

plague006:
Creates a diagonal, wavy, ribbon world.

Makes a generic bus + blueprints base difficult. Also challenges players to think about rail layout.


grilledham:
This is really cool. Maybe we should play this map Tuesday?

plague006:
Would love that

PR #295, opened 2018-11-11
Move cheat tracking outside of control.lua

plague006:
At the same time, we should probably also remove the player-specific code since we can still track cheating.

Thoughts @grilledham @Valansch ?

### DIFF
--- a/features/free_item_logging.lua
+++ b/features/free_item_logging.lua
@@ -47,6 +47,8 @@ Event.add(
         local count = user_item_record.count
         local time = user_item_record['time'] or Utils.format_time(game.tick)
         data[name] = {count = stack.count + count, time = time}
+        local s = table.concat {'[Cheated item] ', p.name, ' - ', stack.count, ' x ', name}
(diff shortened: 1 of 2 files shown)


features/free_item_logging.lua
```diff
@@ -0,0 +1,62 @@
...
+            data = {}
+            cheat_items[pi] = data
+        end
+
+        local stack = event.item_stack
+        local name = stack.name
+        local user_item_record = data[name] or {count = 0}
+        local count = user_item_record.count
+        local time = user_item_record['time'] or Utils.format_time(game.tick)
+        data[name] = {count = stack.count + count, time = time}
```
iltar: Can we also add a special log message to detect this? Would make it easier to find

plague006:
Sample output:
1659.281 Script @C:/Users/Matthew/AppData/Roaming/Factorio/temp/currently-playing/features/free_item_logging.lua:51: [Cheated item] plague006 - 1 x iron-chest

PR #308, opened 2018-11-11
Force diggy biters to spawn, even if there's no space

### LINKED ISSUE #302: Diggy biters not spawning due to size
When digging, it often happens that beyond small biters, they refuse to spawn, this due to biter size and the game not force spawning anything. This kind of breaks the difficulty later on, meaning you can dig really far without any resistance at all. 

What about making biters "destroy" the rocks they would spawn on to force space? This would mean that if no spawn point can be found, the size of a small-biter should be located and a decent sized area auto-mined or destroyed (with special effects?), to ensure biters can spawn.

linaori:
Should increase performance a bit and ensure biters/spitters are spawned inside rocks if there's no space. It's not perfect because they don't automatically start digging their way out, but this is something we can tackle at a later stage, it works decently in most scenarios and forces player action due to high level spitters.

Fixes #302

### DIFF
--- a/map_gen/Diggy/Feature/AlienSpawner.lua
+++ b/map_gen/Diggy/Feature/AlienSpawner.lua
@@ -28,7 +28,7 @@ local function spawn_alien(surface, x, y)
         insert(units, {name = name, position = position, force = enemy_force, amount = amount})
     end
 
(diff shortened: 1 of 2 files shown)


Valansch:
Looks good

PR #313, opened 2018-11-11
Fixed some small things from feedback and issues

### LINKED ISSUE #275: Clearify sending stone to surface
It's a commonly asked in-game, especially by new players, what it does. 

Suggestion: Explain it in the message and refer to the button the left top.
Suggestion: Change the stone icon to the market icon (if possible).

### LINKED ISSUE #280: Diggy cluster Ore balancing
At some point I'd like to have the whole ore spawning system rewritten to something better, but for now it will suffice to simply re balance the numbers a bit. I think the scattering of ore everywhere is fine, it adds the mine feeling. However, the clustered ore looks like it's slightly too big with too much ore per tile. cluster size should probably be lowered a little bit while reducing the flat multiplier to 1.6 or 1.7.

linaori:
Fixes #275 
Fixes #280

 - Replaced small electric pole with raw wood
 - Some minor performance improvements
 - Renamed noise grid debug settings and made them only show what's needed
 - Made pump-able liquids have a configurable modifier so modded resources can be added

### DIFF
--- a/map_gen/Diggy/Config.lua
+++ b/map_gen/Diggy/Config.lua
@@ -127,7 +127,7 @@ local Config = {
                 ['behemoth-spitter'] = 6,
             },
 
-            -- Shows the chest locations, only use when debugging
+            -- shows the chest locations, only use when debugging
             display_chest_locations = false,
 
             treasure_chest_raffle = {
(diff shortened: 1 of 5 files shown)


Valansch:
Looking good.

PR #317, opened 2018-11-12
fixed tile corruption issus

### LINKED ISSUE #310: Corrupt values when removing/placing tiles
There have been several reports about corrupt values regarding floor tiles. Namely:
 - hazard (refined) concrete not carrying what it should
 - Sometimes a cave-in happens where it shouldn't
 - Adding/removing tiles seems to directly corrupt the stress levels in some scenarios, see gif

https://i.gyazo.com/de38739b0590ac64ef9cded09e6559dc.mp4

Valansch:


map_gen/Diggy/Feature/DiggyCaveCollapse.lua
```diff
@@ -288,6 +288,28 @@ function DiggyCaveCollapse.register(cfg)
     config = cfg
     support_beam_entities = config.support_beam_entities
 
+    if support_beam_entities['stone-path'] then
```
iltar: What if we add a config item instead with something like:
```lua
support_beam_aliases = {
    ['stone-brick'] = {'stone-path'},
    ['hazard-concrete'] = {'hazard-concrete-left', 'hazard-concrete-right'},
    ['refined-hazard-concrete'] = {'refined-hazard-concrete-left', 'refined-hazard-concrete-right'},
}
```

This will make it easier for modded configs to add aliases to walls, without having to fix any config values when we decide to change the wall strength

Valansch: This will allow for illegal config states again...
You can certainly already break the config if you try too. However I think it should be somewhat robust towards easy mistakes. That includes missing an alias. If you have new tiles you can always still add them as support_beam_entitity.
I dont see how the alias implementation makes modding easier.

iltar: > I dont see how the alias implementation makes modding easier.

`['stone-wall'] = {'bob-wall-1', 'bob-wall-2', 'wooden-wall'} -- automatically get the wall strength`

If with the above the wall strength is changed, so is that for all others, no need to copy values or cause local conflicts

Valansch: I dont think this is any easier than doing:

['stone-wall'] = 1
['bob-wall-1'] = 1
['bob-wall-2'] = 1
etc...

Especially since most modded items probably will get diffieren strangth than their vanilla counterpart.

PR #318, opened 2018-11-12
Isolated market items and buffs

### LINKED ISSUE #300: Diggy stone reward system overhaul
Currently it's a big list of "at X stone, you get reward Y". This is a lot of effort to maintain and balance. It would be easier to have an "at level X, you get reward Y", where the level requirements in stone are automatically calculated. This makes it possible to have infinite rewards, but also allows to make the scaling easier for single player/lan matches.

SimonFlapse:
~~Needs additional testing~~
Everything seems fine, no desync with toggle-heavy-mode

I have changed the GUI, and implemented ~~practical~~ true infinite stone progression. Created levels to mask the stone requirement.

~~Still needs some work, and refactoring to restore config.lua to it's original clean look.~~
**New cleaner config.lua**

~~I have also left a lot of the old code as comments. I need to remove those entirely. Same goes for some test code.~~

Closes #300


SimonFlapse:
image

map_gen/Diggy/Feature/MarketUnlockables.lua
```diff
@@ -0,0 +1,72 @@
+-- dependencise
+
+-- this
+local MarketUnlockables = {}
+
+local marked_prototype_items = {}
+
+local function market_prototype_add(self_level, self_price, self_name)
+    if (not marked_prototype_items[self_level]) then
+        table.insert(marked_prototype_items, self_level, {})
```
iltar: Can you extract this to `local insert = table.insert` and reference to it as `insert(..., ...)`? This gives a minor performance boost. Same goes for `math.floor` and `math.ceil` below.

map_gen/Diggy/Feature/MarketUnlockables.lua
```diff
@@ -0,0 +1,72 @@
...
+        
+        levelcost[i] = number
+        prev_number = number
+
+        
+    end
+    
+    -- Add new market unlockables here
+    -- market_prototype_add(unlock_level, price, prototype_name)
+    market_prototype_add(1, 50, 'raw-fish')
```
iltar: Removing this from the config will make it a lot harder for people to configure this, is there any way where we can do this without splitting the config? You, I, modders and most of the RedMew admins won't have trouble with this, but there's been quite a high demand for Diggy by non-technical players. I would like to have it easy for them to add/remove/change this config as well

map_gen/Diggy/Feature/MarketExchange.lua
```diff
@@ -387,16 +370,77 @@ local function redraw_table(data)
...
             tag_items.style.font_color = {r = 1, g = 1, b = 1 }
         else
             tag_stone.style.font_color = {r = 0.5, g = 0.5, b = 0.5 }
-            tag_buffs.style.font_color = {r = 0.5, g = 0.5, b = 0.5 }
             tag_items.style.font_color = {r = 0.5, g = 0.5, b = 0.5 }
         end
     end
 end
 
+local function redraw_buff(data) --! Almost equals to the redraw_table() function !
```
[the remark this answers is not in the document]
SimonFlapse: I could do that, i'm just to lazy. If permitted, i'd like to return later to fix it.

map_gen/Diggy/Feature/MarketExchange.lua
```diff
@@ -229,37 +247,26 @@ local tag_label_item = Gui.uid_name()
...
     Gui.clear(frame)
 
-    local heading_table = frame.add({type = 'table', column_count = 3})
-    apply_heading_style(heading_table.add({type = 'label', name = tag_label_stone, caption = 'Name'}).style, 90)
-    apply_heading_style(heading_table.add({type = 'label', name = tag_label_buff, caption = 'Buff'}).style, 200)
-    apply_heading_style(heading_table.add({type = 'label', name = tag_label_item, caption = 'Item'}).style, 200)
+    local heading_table = frame.add({type = 'table', column_count = 2})
+    apply_heading_style(heading_table.add({type = 'label', name = tag_label_stone, caption = 'Requirement'}).style, 100)
+    apply_heading_style(heading_table.add({type = 'label', name = tag_label_buff, caption = header_caption}).style, 220)
+    --apply_heading_style(heading_table.add({type = 'label', name = tag_label_item, caption = 'Item'}).style, 200)
```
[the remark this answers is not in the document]
SimonFlapse: Yeah, missed that one, thanks.

map_gen/Diggy/Feature/MarketExchange.lua
```diff
@@ -229,37 +247,26 @@ local tag_label_item = Gui.uid_name()
 
 local function apply_heading_style(style, width)
     style.font = 'default-bold'
-    style.align = 'center'
+    --style.align = 'center'
```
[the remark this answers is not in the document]
SimonFlapse: Another one I missed, thanks 👍

map_gen/Diggy/Feature/MarketUnlockables.lua
SimonFlapse: Okay got confused by it earlier so just stuck with the LUA reference, but will change from now on.

map_gen/Diggy/Feature/MarketUnlockables.lua
SimonFlapse: I'm not really sure what else to do. I'm open for ideas but the old way of declaring them was troublesome, and I need to pass it through the stone calculation. I commented that they should visit MarketUnlockables.lua to add/remove.
I could further clarify what to edit.
I'll think about it

map_gen/Diggy/Feature/MarketUnlockables.lua
iltar: 👍

map_gen/Diggy/Feature/MarketUnlockables.lua
SimonFlapse: fixed in https://github.com/Valansch/RedMew/pull/318/commits/da82a748b11710d42ca0846b38b5bdb3e915c89b

map_gen/Diggy/Feature/MarketUnlockables.lua
SimonFlapse: fixed in https://github.com/Valansch/RedMew/pull/318/commits/da82a748b11710d42ca0846b38b5bdb3e915c89b

map_gen/Diggy/Feature/MarketExchange.lua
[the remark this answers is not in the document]
SimonFlapse: fixed in https://github.com/Valansch/RedMew/pull/318/commits/da82a748b11710d42ca0846b38b5bdb3e915c89b

map_gen/Diggy/Feature/MarketExchange.lua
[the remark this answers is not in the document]
SimonFlapse: fixed in https://github.com/Valansch/RedMew/pull/318/commits/da82a748b11710d42ca0846b38b5bdb3e915c89b

SimonFlapse:
I didn't actually get around to making it infinite. It currently works by making 100 levels. Depending on the initial `b` value in `MarketUnlockables.lua` the max stone limit is `b*10^6` (20 million default)

SimonFlapse:
I think I managed to make it truely infinite.

Also added more information to the tool tip in next commit:
image

iltar:
Oh nice! This would be awesome to re-use for the force leveling system, but replace stone with XP

map_gen/Diggy/Config.lua
```diff
@@ -1,496 +1,317 @@
...
-            },
-        },
-    },
-}
-
-return Config
+-- dependencies
+
+-- this
+local Config = {
```
iltar: According to github, this is a whitespace change, any idea why?

map_gen/Diggy/Config.lua
```diff
@@ -1,496 +1,317 @@
...
+            },
+
+            -- every x ticks it will clear y currency_item
+            void_chest_frequency = 307,
+            
+            -- add or remove a table entry to add or remove a unlockable item from the mall.
+            -- format: {unlock_at_level, price, prototype_name},
+            unlockables = require('map_gen.Diggy.Feature.MarketUnlockables').initalize_unlockables(
+                {
+                    {level = 1, price = 50, name = 'raw-fish'}, -- unlocks at level 1, price is 50 and the prototype name for fish is raw-fish.
```
iltar: Is the comment still needed? It looks obvious if you replace `is` by `=` 😄

map_gen/Diggy/Feature/MarketExchange.lua
```diff
@@ -104,6 +105,8 @@ local function update_stone_collecting()
     stone_collecting.initial_value = old_modifier + stone_collecting.active_modifier
 end
 
+
+--Handles the updating of market items when unlocked, also handles the buffs
```
iltar: There's actually luadoc for this (I found out how to do this yesterday), add a 3rd dash and it becomes documentation rather than a comment

map_gen/Diggy/Feature/MarketExchange.lua
```diff
@@ -130,24 +138,34 @@ local function update_market_contents(market)
...
+                mining_efficiency.market_modifier = mining_efficiency.market_modifier + (value / 100)
+            elseif (buffs.prototype.name == 'inventory_slot') then
+                local value = buffs.prototype.value
+                print('Mining Foreman: Increased inventory slots by ' .. value .. '!')
+                should_update_inventory_slots = true
+                inventory_slots.market_modifier = inventory_slots.market_modifier + value
+            elseif (buffs.prototype.name == 'stone_automation') then
+                local value = buffs.prototype.value
+                if (stone_tracker.current_level == 1) then
+                    print('Mining Foreman: We can now automatically send stone to the surface from a chest below the market!')
```
iltar: nice!

map_gen/Diggy/Feature/MarketExchange.lua
```diff
@@ -229,37 +247,24 @@ local tag_label_item = Gui.uid_name()
...
     style.font = 'default-bold'
-    style.align = 'center'
     style.width = width
 end
 
-local function redraw_heading(data)
-    local frame = data.market_list_heading
+local function redraw_heading(data, header)
+    local frame = (header == 1) and data.market_list_heading or data.buff_list_heading
+    local header_caption = (header == 1) and 'Reward Item' or 'Reward Buff'
```
iltar: Could move the comparison to a local, not sure if it matters much. Are the parenthesis necessary for the 'ternary'?

map_gen/Diggy/Feature/MarketExchange.lua
```diff
@@ -268,11 +273,11 @@ local function redraw_progressbar(data)
 
     local range = next_stone - act_stone
     local sent = stone_tracker.stone_sent_to_surface - act_stone
-    local percentage = sent / range
+    local percentage = (math.floor((sent / range)*10^4))/10^4
```
iltar: In Lua multiplications are faster: http://lua-users.org/wiki/OptimisationCodingTips. Though for a simple static number like this, unfolding might be a better idea: `10000`. For the second scenario, I recommend to do `*0.0001` as multiplications are faster than division.

Can also do the `local floor = math.floor` (measured performance difference is ~30% for this call).

map_gen/Diggy/Feature/MarketExchange.lua
```diff
@@ -387,16 +368,77 @@ local function redraw_table(data)
...
+        -- get items and buffs for each stone value
+        buffs = config.buffs
+        
+        local result = {}
+
+        -- 1st column
+        result[1] = 'All levels'
+
+        -- 2nd column
+        if buffs[i].prototype.name == 'mining_speed' then
```
iltar: Can you do `local buff_prototype = buffs[i].prototype`? That saves about 2x table access per usage

map_gen/Diggy/Feature/MarketExchange.lua
```diff
@@ -387,16 +368,77 @@ local function redraw_table(data)
...
+        elseif buffs[i].prototype.name == 'stone_automation' then
+            result[2] = '+ '.. buffs[i].prototype.value .. ' stones automatically sent'
+        else
+            result[2] = 'Description missing: unknown buff. Please contact admin'
+        end
+        
+        -- 3rd column
+        result[3] = ''
+        -- indicator to stop print level number
+        if i > 1 then
```
iltar: This could actually be rewritten as `result[4] = i > 0` I believe

map_gen/Diggy/Feature/MarketUnlockables.lua
```diff
@@ -0,0 +1,58 @@
+-- dependencise
```
iltar: typo 😄 `dependencies`

map_gen/Diggy/Feature/MarketUnlockables.lua
```diff
@@ -0,0 +1,58 @@
...
+
+-- this
+local MarketUnlockables = {}
```
iltar: Can you move this file 1 directory lower? It's not a Diggy feature, more of a generic functionality

map_gen/Diggy/Feature/MarketUnlockables.lua
```diff
@@ -0,0 +1,58 @@
...
+end
+
+function MarketUnlockables.initalize_unlockables(items)
+    local levelcost = {}
+    local unlockables = {}
+    local prev_number = 0
+    for i = 1,100 do
+        local b = 20 -- Default 20 <-- Controls how much stone is needed.
+        local start_value = 50 -- The start value/the first level cost
+        local formula = b*(i^3)+(start_value-b)
```
iltar: In this scenario, the `b` and `start_value` seem to only be used here and always local with the same calculation. It will allocate memory and calculate the same thing 100 times here. Would it make sense to just do `local formula = b*(i^3)+30` instead? Considering the many times this is executed, it might be interesting to do `i*i*i` instead of `i^3` as well, or is my math off here? 🤔

map_gen/Diggy/Feature/MarketUnlockables.lua
```diff
@@ -0,0 +1,58 @@
...
+        
+        local precision = 2 -- Sets the precision
+        
+        --Truncates to the precision and prevents dublicates by incrementing with 5 in the third highest place
+        local number = formula
+        local numberlen = floor(log10(number)+1)
+        precision = (numberlen >= 8) and (precision+1) or precision
+        number = number/10^(numberlen-precision)
+        number = floor(number)*10^(numberlen-precision)
```
iltar: `10^(numberlen-precision)` can be put in a local here

map_gen/Diggy/Feature/MarketUnlockables.lua
```diff
@@ -0,0 +1,58 @@
...
+        levelcost[i] = number
+        prev_number = number
+    end
+    
+    -- handles the unlockables from Config.lua in map_gen.Diggy
+    for _, item in pairs(items) do
+        MarketUnlockables.add(item.level, item.price, item.name)
+    end
+    
+      for lvl, v in pairs(marked_prototype_items) do
```
iltar: minor indenting

map_gen/Diggy/Feature/MarketUnlockables.lua
```diff
@@ -0,0 +1,58 @@
...
+        while (prev_number >= number) do
+            number = (prev_number > number) and number or number + ceil(5*10^(numberlen-3))
+        end
+        
+        levelcost[i] = number
+        prev_number = number
+    end
+    
+    -- handles the unlockables from Config.lua in map_gen.Diggy
+    for _, item in pairs(items) do
```
iltar: the `pairs`  could probably be `ipairs` here

map_gen/Diggy/Feature/MarketUnlockables.lua
SimonFlapse: I have a newer version where these values are configurable in the config.lua

In that version infinite level calculation works (almost) so it will not do the for 1,100 loop anymore

iltar: If that's the case, you have to verify my other comments to see if they are still applicable

SimonFlapse: Regarding the math, I can't do +30 since that would not garentie a start requirement of 50 If you change b.

iltar: If it's configurable, my comment doesn't make sense. If anything, the start_value result can be put in a local, but that's it.

map_gen/Diggy/Feature/MarketUnlockables.lua
SimonFlapse: You like it outside of the map_gen.Diggy?

map_gen/Diggy/Feature/MarketExchange.lua
SimonFlapse: I think I'd refrain from changing much here since I'd like to combine the function with redraw_table()

map_gen/Diggy/Feature/MarketExchange.lua
SimonFlapse: Yeah. I'll keep that in mind

map_gen/Diggy/Feature/MarketExchange.lua
SimonFlapse: Oh nice you want the code to be documented with luadoc?

map_gen/Diggy/Config.lua
SimonFlapse: I'd remove it

map_gen/Diggy/Feature/MarketUnlockables.lua
iltar: `map_gen.Diggy` would be the right location

map_gen/Diggy/Feature/MarketExchange.lua
iltar: If possible, I've used a custom format before, but that doesn't work well with intellij, luadoc should be globally supported

map_gen/Diggy/Feature/MarketExchange.lua
iltar: I don't know if it is, just looks slightly odd to do the same comparison twice

map_gen/Diggy/Feature/MarketExchange.lua
SimonFlapse: I'd like to add luadoc eventually

map_gen/Diggy/Config.lua
SimonFlapse: No idea. That's weird. Could be because I moved a dependency

map_gen/Diggy/Feature/MarketUnlockables.lua
SimonFlapse: Fixed in https://github.com/Valansch/RedMew/pull/318/commits/c87511d0c0fe4162194b9a5d1e874da1201857e7

map_gen/Diggy/Feature/MarketUnlockables.lua
SimonFlapse: Fixed in https://github.com/Valansch/RedMew/pull/318/commits/c87511d0c0fe4162194b9a5d1e874da1201857e7

map_gen/Diggy/Feature/MarketUnlockables.lua
SimonFlapse: Fixed in https://github.com/Valansch/RedMew/pull/318/commits/c87511d0c0fe4162194b9a5d1e874da1201857e7

map_gen/Diggy/Feature/MarketUnlockables.lua
SimonFlapse: Fixed in https://github.com/Valansch/RedMew/pull/318/commits/c87511d0c0fe4162194b9a5d1e874da1201857e7

map_gen/Diggy/Feature/MarketExchange.lua
SimonFlapse: Fixed in https://github.com/Valansch/RedMew/pull/318/commits/c87511d0c0fe4162194b9a5d1e874da1201857e7

map_gen/Diggy/Config.lua
SimonFlapse: Fixed in https://github.com/Valansch/RedMew/pull/318/commits/c87511d0c0fe4162194b9a5d1e874da1201857e7

map_gen/Diggy/FormatMarketItems.lua
```diff
@@ -5,19 +5,27 @@ local FormatMarketItems = {}
 
 local market_prototype_items = {}
 local insert = table.insert
-
-function FormatMarketItems.add(self_level, self_price, self_name)
+--- Returns the correct format for Diggy.Feature.MarketExhange.lua to process
+-- @param self_level integer of the level the given item should be unlocked at
```
iltar: Lua only has numbers and floats as types (as far as I know) 😢

map_gen/Diggy/FormatMarketItems.lua
```diff
@@ -5,19 +5,27 @@ local FormatMarketItems = {}
...
+local function add(self_level, self_price, self_name)
     if (not market_prototype_items[self_level]) then
         insert(market_prototype_items, self_level, {})
     end
     insert(market_prototype_items[self_level], {price = self_price, name = self_name})
 end
 
+--- handles the unlockable market items from Config.lua in map_gen.Diggy
+-- serves as a handler for an array of items and passes it on to FormatMarketItems.add() that returns the correct format for Diggy.Feature.MarketExhange.lua to process.
+-- @param items array of items where each item is an table with keys: level (integer level it unlocks at), price (price in the configured currency_item) and name (has to be an entity's prototype-name)
```
iltar: Should be `table of items` when speaking in technical terms, ideally I would call it an iterator, but lua doesn't seem to support this outside if the context of (i)pairs within a loop

map_gen/Diggy/MarketUnlockables.lua
```diff
@@ -19,6 +19,11 @@ local function truncate(precision, formula)
     return number, numberlen
 end
 
+--- Handles the level requirement to stone sent. Calculates based on a forumla one number corresponding to that levels cost
+-- You can configure this in Diggy.Config.lua under features.MarketExhange
+-- @param level integer of a level
+-- @returns integer of cost corresponding to the level based on a calculation
```
iltar: Should also be "number"

map_gen/Diggy/FormatMarketItems.lua
SimonFlapse: Yeah I was thorn on that, said integer because it realy needs to be an integer (And is auto converted to an int) But I'll change to number if you'd like to to only state types available in lua

map_gen/Diggy/FormatMarketItems.lua
SimonFlapse: Oh yeah sorry, had a long day yesterday :)

map_gen/Diggy/FormatMarketItems.lua
iltar: Not sure if this is accurate for the current version, but: https://www.lua.org/pil/2.3.html

SimonFlapse: The problem with double (number) is that stone or any other entity in Factorio can't be 1/2 entity or any other part of 1. So if you make the calculation from a decimal number you'd end up creating a mess in the market and the stone sent to surface popup.

I tried to work around it, but eventually just decided to convert it to the nearest lower integer (math.floor)

If you think math.floor causes performance issues, I could isolate some of them to only make the conversion once.

map_gen/Diggy/FormatMarketItems.lua
SimonFlapse: Fixed in https://github.com/Valansch/RedMew/pull/318/commits/b32afe36c27968e49cde7a39a7346c3ea2e81fae

map_gen/Diggy/FormatMarketItems.lua
SimonFlapse: Fixed in https://github.com/Valansch/RedMew/pull/318/commits/b32afe36c27968e49cde7a39a7346c3ea2e81fae

SimonFlapse:
Alright everything should be good to go. I might add some luadoc for the part of MarketExchange.lua that I have been working with.

iltar:
Looks good to me! I'm not 100% sure why the config is complaining though, it looks odd 🤔 I'll see if I can merge this when I get home, unless someone else wants to

SimonFlapse:
I should have manually merged the config. I used https://www.diffchecker.com/ to compare line to line.

Valansch:
You are using \n line endings while the config file was \r\n before.

Which is fine by me.

iltar:
Ah so that's the problem, I made them with `\n\r` most likely, but my git is set up to ignore those

PR #319, opened 2018-11-12
implemented autojail on too many collapses

Valansch:


features/custom_commands.lua
```diff
@@ -487,8 +487,8 @@ local function jail_player(cmd)
         Game.player_print('Usage: /jail <player>')
         return
     end
-
-    Report.jail(player, cmd['parameter'])
+    local target = game.players[cmd['parameter']]
```
iltar: Shouldn't we be using `Game.get_player_by_index(cmd['parameter'])` here? Or will it always work when you have the name?

map_gen/Diggy/Config.lua
```diff
@@ -490,6 +490,15 @@ local Config = {
                 {stone = 10000000, type = 'buff', prototype = {name = 'inventory_slot', value = 1}},
             },
         },
+
+
+        --Tracks players causing collapses
+        Antigrief = {
+            enabled = true,
+            autojail = true,
+            allowed_collapses_per_hour = 3
```
iltar: That's a very low amount, If a regular joins, digs a few rocks and within 5 minutes, digging far away from the base, this will cause a jail. What if we only trigger this when the entity is a support beam other than an entity place-able by a player?

Valansch: No it will not. Within the first hour its always 3 collapses that are allowed.
So withing the first 5 minutes you have to cause 4 collapses to get sent to jail.
However if you played 3h you can cause 9 collapses.

Its a config setting, 3 was just my lowest bound. What do you suggest we set it to? 5-6 maybe?
This will mean 18 collapses for a player online for 3h. That should be very rare.

That could be a solution. But i don't know if we really need it if you just sett the allowed_collapses_per_hour high enough.

plague006: Does it have to be **per** hour? I would think 5 collapses in the **first** hour of play aught to catch anyone intentionally griefing. It's incredibly rare for someone to spend an hour or more before griefing, In fact it's so rare that I'm sure there would be more false positives than legitimate cases.

Valansch: Okay ill see what i can do :P

PR #320, opened 2018-11-12
Regulars: remove duplicates and sort alphabetically

plague006:
`.shang` was removed from the list completely because I don't believe a factorio name *can* start with a period?


Valansch:
> I don't believe a factorio name can start with a period

Condiering "1" is a valid name i wouldnt be surprised if you were wrong :D

plague006:
It wouldn't surprise me either. If .~$$shang comes back, I owe them an apology. :)

PR #328, opened 2018-11-13
Dev diggy stress colors

Rdjrjqouqcu:
Makes the stress grid much clearer to interpret, the colors go from green (at <= 0) to red (at 3.57), then turn to white when a collapse is about to occur.

### DIFF
--- a/map_gen/Diggy/Debug.lua
+++ b/map_gen/Diggy/Debug.lua
@@ -175,6 +175,16 @@ function Debug.print_colored_grid_value(value, surface, position, scale, offset,
     local u_color = under_bound or color
     local o_color = over_bound or color
     
+    if (color_value < 0) then
+        color = u_color
+    elseif (color_value > 1) then
+        color = o_color
+    else
+        color = { r = color.r + color_value * d_color.r,
+                  g = color.g + color_value * d_color.g,
+                  b = color.b + color_value * d_color.b }
+    end
+    
     text = value
 
     if type(immutable) ~= 'boolean' then
@@ -184,16 +194,6 @@ function Debug.print_colored_grid_value(value, surface, position, scale, offset,
     if not is_string then
         offset = offset or 0
         position = {x = position.x + offset, y = position.y + offset}
-        
-        if (color_value < 0) then
-            color = u_color
-        elseif (color_value > 1) then
-            color = o_color
-        else
-            color = { r = color.r + color_value * delta_color.r,
-                      g = color.g + color_value * delta_color.g,
-                      b = color.b + color_value * delta_color.b }
-        end
 
         -- round at precision of 2
         text = floor(100 * value) * 0.01
--- a/map_gen/Diggy/Debug.lua
+++ b/map_gen/Diggy/Debug.lua
@@ -97,30 +97,103 @@ function Debug.print_grid_value(value, surface, position, scale, offset, immutab
     if type(immutable) ~= 'boolean' then
         immutable = false
     end
-
+    
     if not is_string then
         scale = scale or 1
         offset = offset or 0
         position = {x = position.x + offset, y = position.y + offset}
-		local collapse_stress = 3.57
-		local collapse_amount = value / collapse_stress
+        local r = max(1, value) / scale
+        local g = 1 - abs(value) / scale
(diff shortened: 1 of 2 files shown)


Rdjrjqouqcu:
[https://i.gyazo.com/370eaccccd836a1ec3348fa1733bd755.jpg](https://i.gyazo.com/370eaccccd836a1ec3348fa1733bd755.jpg)

iltar:
While I love this change (makes it a lot better for stress!), this would also change the colors for other noise usage. This debug function is becoming a bit of a monster, it has too much functionality and should probably be split up a bit. The debug function is something generic, so it shouldn't have any specifics for other features.

Rdjrjqouqcu:
Fair enough, feel free to mangle the code so that it works only for the correct debug values or just move it around as needed.

iltar:
Here's an idea that might make it more complex (but we can always split it later), passing a min/max value and a color scheme could solve this issue in a way where we can make it work properly for multiple scenarios; 
```lua
{
    min =  {value = 0, color = {r = 0, g = 1, b = 0}, 
    max = {value = 3.57, color = {r = 1, g = 0, b = 0}},
    out_of_bounds = {r = 1, g = 1, b = 1},
}
```

Rdjrjqouqcu:
I can convert the logic so that it will take parameters: color_value, base_color, color_change, under_bound, over_bound, with the logic defined as follows:
if color_value < 0 : color = under_bound
if color_value > 1 : color = over_bound
else color = base_color + color_value * color_change
So for the example above the parameters would be:
value/3.57, {r = 0, g = 1, b = 0}, {r = 1, g = -1, b = 0}, {r = 0, g = 1, b = 0}, {r = 1, g = 1, b = 1}

map_gen/Diggy/Debug.lua
```diff
@@ -148,4 +148,77 @@ function Debug.print_grid_value(value, surface, position, scale, offset, immutab
     }.active = false
 end
 
+--[[--
```
iltar: Can you replace this multi-line comment with Luadoc? That would give IDEs a lot of extra info. Luadoc is simply 3 dashes in front of the line. Example: https://github.com/Valansch/RedMew/pull/338/files#diff-1e591d484002b3abd05bfdacb86bd7b1R34

map_gen/Diggy/Debug.lua
```diff
@@ -148,4 +148,77 @@ function Debug.print_grid_value(value, surface, position, scale, offset, immutab
...
+    @param immutable bool if immutable, only set, never do a surface lookup, values never change
+    @param color_value float How far along the range of values of colors the value is to be displayed
+    @param base_color {r,g,b} The color for the text to be if color_value is 0
+    @param delta_color {r,g,b} The amount to correct the base_color if color_value is 1
+    @param under_bound {r,g,b} The color to be used if color_value < 0
+    @param over_bound {r,g,b} The color to be used if color_value > 1
+]]
+function Debug.print_colored_grid_value(value, surface, position, scale, offset, immutable,
+        color_value, base_color, delta_color, under_bound, over_bound)
+    local is_string = type(value) == 'string'
```
iltar: Does the string invariant here make sense? I think this case can be removed for this function. It doesn't hurt to keep, but it's extra logic to maintain

map_gen/Diggy/Debug.lua
Rdjrjqouqcu: For this I tried to match the rest of the file.  Should I convert the entire file to Luadocs?  I feel like this should just be it's own issue so that the entire repository can be converted to either one syntax or the other.

map_gen/Diggy/Debug.lua
Rdjrjqouqcu: This allows string values to bypass the colors.  It's necessary in order to not cause issues when displaying 'ore' or 'chest' for some debug modes. It does point out to me that the color code can be used for that text so I'll move it around a bit so it'll work properly.

map_gen/Diggy/Debug.lua
iltar: For new code it's easier to add a luadoc already

Valansch:
I like this solution.

PR #327, opened 2018-11-13
converted chances in config to weights

Rdjrjqouqcu:
issue #309

Issue #309, opened 2018-11-11
Change Diggy resources to weights instead of percentages


iltar:
Feedback from Rdjrjqouqcu
> I'd change the 'resource_chances' field in the config.lua to be a 'resource_weights' field instead. This will make adding modded ores easier without having to recalculate the new weights for every other entry. Something like resource_chances['coal'] = resource_weights['coal'] / sum(resource_weights)

Rdjrjqouqcu:
done, will create a pull request once I figure out how to

PR #329, opened 2018-11-13
ore weights instead of chances

Rdjrjqouqcu:
issue #309 
continuation of #327 because of my horrible git skills

### DIFF
--- a/map_gen/Diggy/Feature/ScatteredResources.lua
+++ b/map_gen/Diggy/Feature/ScatteredResources.lua
@@ -15,17 +15,11 @@ local floor = math.floor
 -- this
 local ScatteredResources = {}
 
-local function get_name_by_weight(collection)
+local function get_name_by_weight(collection, sum)
     local pre_calculated = random()
     local current = 0
-	local sum = 0
-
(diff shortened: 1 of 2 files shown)


map_gen/Diggy/Feature/ScatteredResources.lua
```diff
@@ -15,18 +15,19 @@ local floor = math.floor
...
 
-local function get_name_by_random(collection)
+local function get_name_by_weight(collection, sum)
     local pre_calculated = random()
     local current = 0
-
-    for name, probability in pairs(collection) do
-        current = current + probability
-        if (current >= pre_calculated) then
+    local target = pre_calculated * sum
```
iltar: What about `random(1, sum)` instead?

Rdjrjqouqcu: This way, even if you put in floating point values the logic will still work. The weights will work if you instead put ratios, such as iron=1, copper=0.8,coal=1.25 etc.

iltar: Fair point!

PR #330, opened 2018-11-13
Added maltease_crossings.lua preset.

Jayefuu:


grilledham:
Good work thanks

map_gen/presets/maltease_crossings.lua
```diff
@@ -0,0 +1,177 @@
...
+        b.rectangle(10,8)
+    }	
+water_line = b.change_tile(water_line, true, 'water')
+
+local waters = b.single_y_pattern(water_line, 9)
+local bounds = b.rectangle(10, starting_area+1)
+waters = b.choose(bounds, waters, b.empty_shape)
+waters = b.translate(waters,34,0)
+
+water_pattern =
```
plague006: Does this need to be global in scope? Seems like it can be local

Jayefuu: No need to be global. Just a missed local declaration.

PR #333, opened 2018-11-14
Split create_player from control

plague006:


features/player_create.lua
```diff
@@ -0,0 +1,33 @@
...
+local UserGroups = require 'features.user_groups'
+
+local function player_created(event)
+    local player = Game.get_player_by_index(event.player_index)
+
+    if not player or not player.valid then
+        return
+    end
+
+    if global.scenario.config.admin_check == true then
```
iltar: Is there a scenario where you want to disable this? (testing?)

features/player_create.lua
```diff
@@ -0,0 +1,33 @@
...
+        if UserGroups.is_admin(player.name) then
+            player.admin = true
+            log(player.name .. " promoted to admin via control.")
+        end
+    end
+
+    if (global.scenario.config.fish_market.enable) then
+        player.insert {name = MARKET_ITEM, count = 10}
```
iltar: What if "MARKET_ITEM" does not exist? Where does it come from? I think we need an additional require here (correct me if I'm wrong)

Valansch:
I don't understand this either. What why do you want to replace the server admin list with a scenario resource. This seems very counter productive. It now is a redudant system.
With the new admin interface we are doing quite the opposite. Remove the resource folder entirely and store the data on server side.

Valansch:
But this isnt a backup, this is functionality.

features/player_create.lua
Valansch: MARKET_ITEM comes from the config. It has been used by other files without require so i dont blame @plague006. But it should be changed everywhere it is used.

Valansch:
Thank you. Looks good 👍

PR #335, opened 2018-11-14
Added Luadoc for NightTime.lua

SimonFlapse:
Not much else to say. Title speaks for itself

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,2 @@
 *.swp
 *.sh
-/.idea
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
 *.swp
 *.sh
+/.idea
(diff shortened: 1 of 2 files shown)


.gitignore
```diff
@@ -1,2 +1,3 @@
 *.swp
 *.sh
+/.idea
```
iltar: This one should be reverted and added to your personal git ignore instead: http://egorsmirnov.me/2015/05/04/global-gitignore-file.html

map_gen/Diggy/Feature/NightTime.lua
```diff
@@ -9,6 +11,10 @@ local Game = require 'utils.game'
 -- this
 local NightTime = {}
 
+--- Event handler for on_built_entity
+-- checks if player placed a solar-panel and displays a popup
+-- @param event the on_built_entity event
```
iltar: Missing the type here, I think it's "Table", but not 100% sure `@param event Table the on_built_entity event`. The other luadoc is also missing this

.gitignore
SimonFlapse: Thanks for the link :) I found this instead: https://help.github.com/articles/ignoring-files/#explicit-repository-excludes

PR #337, opened 2018-11-14
Moved report function to cmd_report in report.lua

### LINKED ISSUE #334: Split the report function out of custom_commands.lua
We have an entire [report.lua](https://github.com/Valansch/RedMew/blob/develop/features/report.lua). We should probably be calling that instead of custom_commands having its own [report function](https://github.com/Valansch/RedMew/blob/a366d3aaf96b5ef40f7e8581dc6eb16cbfeb38a9/features/custom_commands.lua#L655)

@SimonFlapse would you be willing to tackle this?

SimonFlapse:
Splitted the report functionality out of custom_command.lua and into report.lua

Reoccurring issue: Some output seems to have indentation (Tabs) before the game.print text, when jailing people from the show-report gui.
image

closes #334

### DIFF
--- a/features/custom_commands.lua
+++ b/features/custom_commands.lua
@@ -651,30 +651,7 @@ commands.add_command(
 )
 commands.add_command('a', 'Admin chat. Messages all other admins (Admins only)', admin_chat)
 
-
-local function report(cmd)
-    local reporting_player = game.player
-    if reporting_player then
-        local params = {}
-        for param in string.gmatch(cmd.parameter, '%S+') do
-            table.insert(params, param)
-        end
-        if #params < 2 then
-            reporting_player.print('Please enter then name of the offender and the reason for the report.')
-            return nil
-        end
-        local reported_player_name = params[1] or ''
-        local reported_player = game.players[reported_player_name]
-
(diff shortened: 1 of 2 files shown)


SimonFlapse:
I found out it happens when the console line wraps the outputted text. Not much I can do about it. My only suggestion is to make the string shorter and use multiple prints.

SimonFlapse:
image

Since i'm already working in report.lua I thought of color coding some important messages that might get lost. It looks fine without other players chatting. Not sure how it would look with or if we want it.

I also moved unjail_player out of custom_commands.lua

iltar:
One side-note, should we use a special prefix to make the message stand out more? Perhaps a dotted line before/after this message?

SimonFlapse:
That's a good idea. I'll add that

SimonFlapse:
A prefix of '## - '
image

iltar:
I like!

SimonFlapse:
I tweaked it so it's clear what is the beginning and what is the end. Otherwise the text inbetween could be viewed as an important message too.
image

I'll post a combined version of this shortly:

PR #338, opened 2018-11-14
Added force_control.lua to track force levels

linaori:
Implementation is _really, really_ basic, but MVP = MVP.

Example usage:

```lua
local ForceControl = require 'features.force_control'
local ForceControl_builder = ForceControl.register(function (level_reached)
    return 100 + level_reached * 10
end)
ForceControl_builder.register_on_every_level(function (level_reached, force)
    force.print('Leveled up to ' .. level_reached .. '!')
end)

Event.on_init(function ()
    ForceControl.register_force('player') -- recommended due to lazy loading the force
    -- or
    ForceControl.register_force(game.forces.player)
end)

function get_force_data()
    ForceControl.get_force_data('player') -- see screenshot
    -- or
    ForceControl.get_force_data(game.forces.player)
end

function add_experience()
    ForceControl.add_experience('player', 500) -- carries over to next level
    -- or
    ForceControl.add_experience(game.forces.player, 500)
end

function remove_experience()
    ForceControl.remove_experience('player', 500) -- minimum experience is 0, does not de-level
    -- or
    ForceControl.remove_experience(game.forces.player, 500)
end
```
image

### DIFF
--- a/features/force_control.lua
+++ b/features/force_control.lua
@@ -1,0 +1,238 @@
+-- dependencies
+local Global = require 'utils.global'
+local Event = require 'utils.event'
+local raise_event = script.raise_event
+local max = math.max
+local insert = table.insert
+
+-- this, things that can be done run-time
+local ForceControl = {}
+ForceControl.events = {
+    --- triggered when the force levels up
+    --- uses event = {level_reached = number, force = LuaForce}
+    on_level_up = script.generate_event_name(),
+}
+
+-- the builder, can only be accessed through ForceControl.register() and should be avoided used run-time
+local ForceControlBuilder = {}
+
+-- all force data being monitored
+local forces = {}
+
+-- the table holding the function that calculates the experience to next level
+local next_level_cap_calculator = {
+    execute = nil,
+}
+
+Global.register({
+    forces = forces,
+    next_level_cap_calculator = next_level_cap_calculator,
+}, function (tbl)
+    forces = tbl.forces
+    next_level_cap_calculator = tbl.next_level_cap_calculator
+end)
+
+---Asserts if a given variable is of the expected type using type().
+---
+---@param expected_type string
+---@param given any
+---@param variable_reference_message string displayed when the expectation is not met
+local function assert_type(expected_type, given, variable_reference_message)
+    local given_type = type(given)
+    if given_type ~= expected_type then
+        error('Argument ' .. variable_reference_message .. ' must be of type \'' .. expected_type .. '\', given \'' .. given_type .. '\'')
+    end
+end
+
+---Register a reward that checks if a reward should be given if the level
+---matches the callback. If so, it will apply the if_level_criteria_matches
+---callback.
+---
+---@param level_matches function function(number level_reached)
+---@param callback function function(number level_reached, LuaForce force)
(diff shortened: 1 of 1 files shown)


iltar:
Should I change the registration of level up to use the event system instead?

iltar:
Internally it now uses the existing event system and if desired, you can hook onto that as well

features/force_control.lua
```diff
@@ -0,0 +1,260 @@
...
+---@param lua_force_name string|nil only register for this force (optional)
+function ForceControlBuilder.register_on_every_level(callback, lua_force_name)
+    assert_type('function', callback, 'callback of function ForceControlBuilder.register_on_every_level')
+
+    ForceControlBuilder.register(always_true, callback, lua_force_name)
+end
+
+---Register the config and initialize the feature.
+---@param level_up_formula function
+function ForceControl.register(level_up_formula)
```
Valansch: Is this suppsed to be called on runtime? I'm guessing not? In this case why not use Token.lua to store level_up_formula instead of Global.lua?

features/force_control.lua
```diff
@@ -0,0 +1,260 @@
...
+---@param lua_force_or_name LuaForce|string
+local function get_valid_force(lua_force_or_name)
+    if not lua_force_or_name then
+        return
+    end
+
+    if type(lua_force_or_name) == 'string' then
+        local force = game.forces[lua_force_or_name]
+        if not force or not force.valid then
+            return
```
Valansch: Returns nil if force doesnt exist. Doc says you always return a valid force.

features/force_control.lua
iltar: This is supposed to be called before the game is initialized due to how the events have to be registered

Valansch: Thats my point. Why not use token, instead of putting the function into the global?

Valansch:
Will merge now, we can always refactor later

features/force_control.lua
```diff
@@ -0,0 +1,260 @@
...
+    forces = tbl.forces
+    next_level_cap_calculator = tbl.next_level_cap_calculator
+end)
+
+---Asserts if a given variable is of the expected type using type().
+---
+---@param expected_type string
+---@param given any
+---@param variable_reference_message string displayed when the expectation is not met
+local function assert_type(expected_type, given, variable_reference_message)
```
mheguy, line 39: Would it make sense to spin this out into a full utility?

features/force_control.lua
```diff
@@ -0,0 +1,260 @@
...
+    local given_type = type(given)
+    if given_type ~= expected_type then
+        error('Argument ' .. variable_reference_message .. ' must be of type \'' .. expected_type .. '\', given \'' .. given_type .. '\'')
+    end
+end
+
+---Returns a valid force based on the lua force or name given.
+---@param lua_force_or_name LuaForce|string
+local function get_valid_force(lua_force_or_name)
```
mheguy, line 48: Same here, should we make this a general utility?

linaori: That would work, could make a `force.lua` for that

features/force_control.lua
linaori: Might have to receive a critical look, but it would be nice

Issue #282, opened 2018-11-10
A credit list for people that have contributed to the RedMew repository


iltar:
I think it would be nice to create a list of people who contributed to the RedMew code-base. Whether it's a typo fix, a balance change, bug fix or a whole new feature. This might encourage players to contribute if they wish, even if it's just to say "Hey, I've contributed to RedMew!". Could be displayed via a GUI in-game for example.

SimonFlapse:
We could add it at the bottom of the "Welcome" tab in the info GUI.

iltar:
That's not a bad idea, or perhaps a new tab? Would prefer to keep the names limited to factorio names

SimonFlapse:
![image](https://user-images.githubusercontent.com/44922798/48314452-0b449a80-e5ca-11e8-8ef0-a3858d96104d.png)

I like the looks of the info GUI as it is. So I don't think a new tab would be a good idea.

My suggestion would still be to do it at the bottom of the "Welcome" tab, or rethink the "Other Servers" tab which hasn't been used.

Side note: The "How To Chat" seems to be placed out of sight for the players that really needs it, I'm not sure what to do about it.

plague006:
I think listing 17 names in a GUI window that is already teeming with information that already goes unread is just adding extra noise. At most we could link to the contributors list somewhere: https://github.com/Valansch/RedMew/graphs/contributors

iltar:
My github name is not the same as my in-game name, so that doesn't seem feasible to me. It should be visible somewhere in-game, with player names, even if it's not obvious to be found.

plague006:
I just fundamentally disagree. As things currently stand: questions are asked hourly about information that is provided in 'info'. We should be trying to trim down 'info' to funnel people towards the most important information.

plague006:
I just fundamentally disagree. As things currently stand: questions are asked hourly about information that is provided in 'info'. We should be trying to trim down 'info' to funnel people towards the most important information.

SimonFlapse:
We could have the info window open automatic on first join

Jayefuu:
Seems like make-work to me. Isn't that what the contributors tab on Github is for?

iltar:
You won't find me in-game under this name, which is what I'm trying to point out

plague006:
Yeah that's a tough issue... I totally get wanting recognition for the work. It's just a question of implementing it in a way that doesn't force it on people but let's them find the info if they want it.

SimonFlapse:
[...]

PR #340, opened 2018-11-14
Made info.gui display on player_created

SimonFlapse:
Removed old welcome message and replaced it with a short reminder on how to change keybinding to enable chat.

Leaving it here to see what people think

#282 discusses the problems of people asking questions available in the info.gui. This is a suggestion on how to help people notice this gui.

### DIFF
--- a/control.lua
+++ b/control.lua
@@ -37,7 +37,7 @@ require 'features.custom_commands'
 
 -- GUIs the order determines the order they appear from left to right.
 -- These can be safely disabled. Some map presets will add GUI modules themselves.
-require 'features.gui.info'
+local info = require 'features.gui.info'
 require 'features.gui.player_list'
 require 'features.gui.poll'
 require 'features.gui.tag_group'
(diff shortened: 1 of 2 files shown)


control.lua
```diff
@@ -60,13 +60,16 @@ local function player_created(event)
     end
     player.insert {name = 'iron-gear-wheel', count = 8}
     player.insert {name = 'iron-plate', count = 16}
-    player.print('Welcome to our Server. You can join our Discord at: redmew.com/discord')
-    player.print('Click the question mark in the top left corner for server infomation and map details.')
-    player.print('And remember.. Keep Calm And Spaghetti!')
+
+    player.print('Trouble chatting? Change the keybinding in:')
```
iltar: People often don't know that the backtick is the key they need to press, perhaps we should also mention this.

Bit of a brainfart, what about a "ping" system? We can parse messages and if we see `@Linaori` and the player is in the map, we could ping that player with a sound.

SimonFlapse: It's listed in the info GUI already, and you can see the Keybinder inside your keyboard settings. I haven't been able to find the backtick on my keyboard (that Factorio recognize) so I didn't include it.

I like the ping system idea.

plague006:
Conflicts #333

SimonFlapse:
I'll make the necessary changes when #333 has been merged

PR #341, opened 2018-11-15
Created a compound cluster mode for scattered resources.

Rdjrjqouqcu:
A few small changes:
-rearranged the config file to be clearer which settings go together
-Added a enable/disable config setting to toggle random single-tile resources (scatter_mode)
And the larger change is the compound_cluster mode:
It essentially bundles together the settings for the cluster mode and allows it to be repeated for different settings in order to generate multiple different ore patches.

This can be used to generate ore patches such as the following:
https://i.gyazo.com/thumb/1000/e9d9d3bfe342bc169cf8f05b85514ca0-jpg.jpg
Where one layer mixes iron and copper ores, a second layer mixes coal and stone, and a third layer adds oil and traces of uranium


map_gen/Diggy/Feature/ScatteredResources.lua
```diff
@@ -80,24 +99,89 @@ function ScatteredResources.register(config)
...
 
-        local distance = floor(sqrt(x * x + y * y))
+        local distance = sqrt(x * x + y * y)
+        
+        if cc_mode then
+            if (spawn_compound_clusters(surface, x, y, distance)) then
+                return
+            end
+        end
+        distance = floor(distance)
```
Valansch: I dont think you really need this floor 🤔

Rdjrjqouqcu:
added in 'connected_tendril' and 'fragmented_tendril' cluster patterns, for example:
https://i.gyazo.com/bf20c1387c47a76348975e484fdb0d81.png

Removed old cluster generation and cleaned up names and config settings

Fixed yield multiplier for each individual cluster

Rdjrjqouqcu:
Do we want to change how resources increase based on distance? currently it's the same as the original, which was that ores increase by 100% every 70 tiles. I feel like that's too high for the tendril spawning, and that it should be something closer to every 150 or 200 tiles, however it will affect game balance significantly.

iltar:
Merging to see if we can make it in time for the diggy run to get the tendrils working

Issue #336, opened 2018-11-14
Use consistent line endings


Valansch:
We may want to make a decision on what to use. We currently have (excluding data)  ~5000 windows style line endings while we have ~22000 unix style line endings in the project.

We should decide on changing all line endings to one of the two.

While I like \n more it is not going to be read by windows editor, which many people who use the scenario use.

iltar:
`\n` is supported by every windows editor except for notepad, but this will be added in the near future. People that use just notepad (not ++), already have issues as is due to the massive amount of `\n`. The windows LF are probably because of me. 

My vote goes for `\n`

Jayefuu:
I'm for whatever makes it easiest to maintain the regulars list. Please decide then provide instructions on how to format the list.

Valansch:
Thats not what i understood^^
I was bringing up the editor issue but @iltar said editor will be updated anyways.
I'm very much undecided, but i like \n more, and since i didnt see anyone else arguing for \r\n i think we should just use \n.

BUT @iltar brought up that since we dont have an automatic way to enforce this we should require it for commits. So we can change it lazily in the code, but not discouraging people from contributing.

Jayefuu:
Can git autoformat submissions so it's a non-issue? I used Windows, so I didn't think to check what line endings my editor was using. I just submitted. Can we tell git to ignore the endings?

plague006:
You should get where you want to be with `git config --global core.autocrlf true`

plague006:
https://github.com/Valansch/RedMew/issues/206 has a bit more on the subject of formatting overall

Valansch:
@Jayefuu git doesnt provide an autoformatter.
But git provides line ending functionallity:

  **Checkout Windows-style, commit Unix-style**
    Git will convert LF to CRLF when checking out text files. When committing text files, CRLF will be converted to LF. For cross-platform projects, this is the recommended setting on Windows ("core.autocrlf" is set to "true")

 **Checkout as-is, commit Unix-style**
    Git will not perform any conversion when checking out text files. When committing text files, CRLF will be converted to LF. For cross-platform projects this is the recommended setting on Unix ("core.autocrlf" is set to "input").

 **Checkout as-is, commit as-is**
    Git will not perform any conversions when checking out or committing text files. Choosing this option is not recommended for cross-platform projects ("core.autocrlf" is set to "false")

PR #347, opened 2018-11-15
Updated from changes to regulars on Server 1 and Server 2

Jayefuu:
Downloaded regulars.lua from Server 1 (Winding Road) and Server 2 (previously Diggy), merged them and deleted duplicated. Sorry for the mass delete and re-add, next time I will just modify, but was in a rush.


Valansch:
Missed that. You are right. I ususally wouldn't mind " vs '. I personally use " more often. 
But this file is used often by many people. And ' seems to be used more in this project by people. 
So if we merge this with " we will see a commit in the future changing it all back to ' because someone has their autoformatter on ' or whatever :)

Valansch:
Okay. I rest my case. If user_groups outputs " then i would say we leave it as is. The file endings can be fixed via #336

PR #350, opened 2018-11-15
Create player fix

plague006:
Fixes #333 overwriting https://github.com/Valansch/RedMew/pull/340


SimonFlapse:
https://github.com/Valansch/RedMew/blob/263b8fb6801527bffe53399035f1451d2a1fd437/features/user_groups.lua#L4
Is causing a error because resources.admins no longer exists

image

PR #356, opened 2018-11-16
Shift elements around, remove "other servers", add train savior info

plague006:


features/gui/info.lua
```diff
@@ -142,17 +142,28 @@ local pages = {
...
+                ]]
             )
 
+
+
+            header_label(parent, 'How To Chat')
+            centered_label(
+                parent,
+                [[
+To chat with other players, press the "grave" key on your keyboard.
```
iltar: grave key? I've never heard of this term. Only know this key as tilde or back tick

plague006: I believe you, but I'm not sure why you're saying this?
We're using factorio's nomenclature, we also give the other names it's known by, and explicitly give instructions on changing/finding the keybind explicitly to avoid the issue of the key's name.

Marking this as resolved.
capture

plague006: This is just factorio's nomenclature. I personally learned it as the tilde key before anything else. The other names the key is known by are included, and we explicitly give instructions on changing/finding the keybind in case they're on a non-standard keyboard. I'm not sure we could do much more.

capture

iltar: If that's what Factorio calls it, we should follow their naming schema 👍

PR #357, opened 2018-11-16
Extend tables library, split on_create and hodor messages from functions

plague006:


utils/list_utils.lua
```diff
@@ -78,35 +80,59 @@ table.get = function (t, index)
   error("Index out of bounds", 2)
 end
 
+-- Chooses a random entry from a table
+table.get_random = function(this_table)
+    return this_table[math.random(#this_table)]
+end
+
+-- Chooses a random entry from a weighted table
```
iltar: Can you change those to luadocs and add a usage example?

utils/list_utils.lua
```diff
@@ -78,35 +80,59 @@ table.get = function (t, index)
...
+      total_weight = total_weight + w[weight_index]
+  end
+
+  local index = math.random(total_weight)
+  local weight_sum = 0
+  for _, w in pairs(weighted_table) do
+      weight_sum = weight_sum + w[weight_index]
+      if weight_sum >= index then
+          local string = w[item_index]
+          return string
```
iltar: You can directly return this instead of putting it inside a local

utils/list_utils.lua
plague006: Done (I think?)

PR #360, opened 2018-11-16
Add day/night cycle options

plague006:
I haven't looked at this in a while and it might be a mess, but if I don't PR it now I'm likely to just forget about this branch.


iltar:
Could probably just merge this and expand on it at a later point

Valansch:
If you change the filename to README.md instead of txt github will show it if you navigate to the folder.

plague006:
Quite right

Valansch:
Alrighty :D

PR #363, opened 2018-11-16
Added grid_islands_rotated.lua to map presets.

Jayefuu:


map_gen/presets/grid_islands_rotated.lua
```diff
@@ -0,0 +1,170 @@
...
+-- /silent-command game.forces["player"].technologies["landfill"].enabled = false
+-- /silent-command game.forces.player.character_running_speed_modifier = 1.5
+-- /silent-command game.difficulty_settings.technology_price_multiplier=2
+
+local b = require 'map_gen.shared.builders'
+local Random = require 'map_gen.shared.random'
+local ore_seed1 = 1000
+local ore_seed2 = ore_seed1 * 2
+local island_separation = 350
+local math = require "utils.math"
```
iltar: `require 'utils.math'` should be sufficient as it automatically overwrites it. But to be fair, this is one of those tools that shouldn't be needed as require. Perhaps it's best to ensure these and table functions are _always_ loaded as part of the RedMew core?

Valansch: I disagree with both points.

1. 
`require 'utils.math'` does not do the same thing as `local math = math`.
Assuming utils.math was already required by another module the former does nothing.
While the latter sets a local ref to a table in _G, which halfs the access time because you are no longer accesing _G.

2. I would rather not assume the implementation of an unrelated module for one module to function properly. If you require math explicitly your only assumption about the rest of the codebase is that utils.math exists. This is especially important because we 100% want to be sure that math is ALWAYS called to stop sine related desyncs. Thats the reason why utils.math is required in control.lua AND every module that needs it. Just to be extra sure. This way n things have to go wrong, not one :)

PR #371, opened 2018-11-17
Allowing a single character after a mention

SimonFlapse:
%S is the opposite of %s (space characters) and the plus sign combines all the characters. Names like der-dave.com should work.

Tested expression using: https://regexr.com/


features/chat_triggers.lua
```diff
@@ -141,7 +142,7 @@ local function hodor(event)
                     success = true
                 end
 
-                if not admin_call and '#'..p.name == word then
+                if not admin_call and ('#'..p.name == word or '#'..p.name == trimmed_word) then
```
iltar: What if I have a player called `linaori` and `linaor`, which will get the trigger? Both 😄

Valansch: Seems like it :D
Whats the reason for this functionality anyways?

SimonFlapse: Yeah both will be called, I should have a break actually in there. That should fix it

Valansch: I think we have a player called val3ry.
So if you type val and hit tab nothing is gonna happen. Then you hit enter to send the command and its gonna notify both me and val3ry. Is that how you intended it to work?

SimonFlapse: Would it do that? I'm only trimming the last character away to allow pinging the cases: #SimonFlapse? and #SimonFlapse, and #SimonFlapse.

Valansch: Oh no it wouldnt do that. My bad. It would only do that if there was a player Valansch and a player Valansc.

Valansch: Because there is /follow. Tabs work afaik. You can use it for commands and player names. It autocompletes to the lowest common denominator

iltar: It won't work mid sentence though

Rdjrjqouqcu: The way Lua's or works, the non-trimmed name has priority over the trimmed name
a or b ==> a iff a is truthy

Valansch: > It won't work mid sentence though
What do you mean? Autocompleting names mid sentence? I use that all the time.

Valansch: > The way Lua's or works, the non-trimmed name has priority over the trimmed name
a or b ==> a iff a is truthy

But this simple evaluates the bool. It checks this for every player. And it is true for all players with the same name as another player except one fewer character. But im not saying its gonna cause any issue...

PR #374, opened 2018-11-17
Updated readme and minor perf fix

linaori:
Updated the readme to be slightly more accommodating and fixed a minor performance issue introduced in a recent change


Valansch (approved):
Looks good

PR #377, opened 2018-11-17
fixed report desyncs

Valansch:
After removing the merge errors i could not reproduce the desyncs on automatic player report after collapse.

### DIFF
--- a/features/report.lua
+++ b/features/report.lua
@@ -93,13 +93,12 @@ function Module.report(reporting_player, reported_player, message)
     if reporting_player then
         player_index = reporting_player.index
     end
-    table.insert(global.reports, {reporting_player_index = reporting_player, reported_player_index = reported_player.index, message = message, tick = game.tick})
+    table.insert(global.reports, {reporting_player_index = player_index, reported_player_index = reported_player.index, message = message, tick = game.tick})
 
     local notified = false
     for _,p in pairs(game.players) do
         if p.admin and p.connected then
             p.play_sound{path='utility/tutorial_notice', volume_modifier = 1}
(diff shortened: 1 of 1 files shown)


plague006:
I feel dumb since someone else approved the changes.. but I'm unable to test this. After checking out 1bb491dc6259d0593fbc1aa6912a414651d4d369 Diggy doesn't load for me since antigrief isn't present.

Is this just #319 all over again? :D

iltar:
The problem is a missing file that wasn't committed along, once we have that commit back in there, it should work.

Valansch:
TODO: Add antigrief to config.

map_gen/Diggy/Feature/Antigrief.lua
```diff
@@ -0,0 +1,57 @@
...
+
+Event.add(CaveCollapse.events.on_collapse, function(event)
+    local player_index = event.player_index
+    if player_index and global.Antigrief.last_collapse ~= game.tick then
+        global.Antigrief.last_collapse = game.tick
+        local count = player_collapses[player_index] or 0
+        count = count + 1
+        player_collapses[player_index] = count
+        local player = Game.get_player_by_index(player_index)
+        if global.Antigrief.autojail and count > allowed_collapses_first_hour and player.online_time < 216000 and not global.Antigrief.jailed_players[player_index] then
```
iltar: Considering we _only_ do anything when this is set to true-ish, wouldn't it make more sense to add it as very first statement?
```lua
Event.add(CaveCollapse.events.on_collapse, function(event)
    if not global.Antigrief.autojail then
        return
    end
```

Valansch: No. This this does not disable Antigrief. This only disables the autojail. It should still count the collapses this way we can see who caused a grief even without autojail enabled.

PR #379, opened 2018-11-17
Fix wrong variable in message

plague006:
Fixes #319 breaking /jail


Valansch:
This is fixed in #377

PR #380, opened 2018-11-17
Move level up logic outside of unlock check loop

damium:
This allows leveling up when there isn't an unlockable defined for that level.


SimonFlapse:
Nice! I never really put much thought into moving it outside, when I redid the leveling up. This also avoids my hideous while loop.

iltar:
Thanks @damium!

PR #381, opened 2018-11-17
Increase cost of landmines in the market

plague006:
As it turns out, landmines in the early game are supremely powerful. One player picked up 10k coins in the first couple hours of maltease cross using them.


SimonFlapse:
You think 3 is enough? I haven't used landmines at all.

plague006:
I don't know if 3 is high enough, but I'd rather under-nerf them than run them into the ground.

PR #387, opened 2018-11-18
Added @ to mention triggers also allowed the prefix as a postfix

SimonFlapse:
Avaliable triggers now:
admin moderator @SimonFlapse #SimonFlapse SimonFlapse@ SimonFlapse#

(@SimonFlapse might trigger discord)


mheguy (approved):
Tested with Simon on server and all went as expected.

PR #389, opened 2018-11-18
Fish Market update. Better description for boosts and train savior item

SimonFlapse:
image

image

image


features/fish_market.lua
```diff
@@ -322,6 +322,12 @@ local function market_item_purchased(event)
         boost_player_miningspeed(player, market)
     end
 
+    if event.offer_index == 3 then -- train saviour item
+        local player = Game.get_player_by_index(player_index)
+        local train_savior_item = Market_items[offer_index].item
+        player.insert{name=train_savior_item, count=1}
```
iltar: This doesn't work for shift-click, `event.count` is the amount of items purchased

SimonFlapse: Yeah 👍 Forgot to change that, thanks for catching. Just copied the command i used to give it ingame

SimonFlapse: Fixed in https://github.com/Refactorio/RedMew/pull/389/commits/4dbf3b251ab62ff659510be43d2c66fcc7079699

PR #391, opened 2018-11-18
Mining related improvements

linaori:
Contains 3 adjustments:
 - Smaller memory footprint for the collapse message in `DiggyCaveCollapse`
 - Coal is now filtered
 - Added a message when the player is digging with a full inventory. Shows every first and then every 5th mining. Is designed to be annoying because it causes performance issues. Is a temporary fix until the stone system is overhauled. Can be turned off with a config setting. Message: _"Miner, you have a full inventory!\n\nMake sure to empty it before you continue digging."_

### DIFF
--- a/map_gen/Diggy/Config.lua
+++ b/map_gen/Diggy/Config.lua
@@ -47,6 +47,10 @@ local Config = {
         DiggyHole = {
             enabled = true,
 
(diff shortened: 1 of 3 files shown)


Valansch:
Lookin good

PR #393, opened 2018-11-18
Changed windows style new lines to unix

Valansch:


Valansch:
Approved :O Here we go! :D

PR #395, opened 2018-11-18
Fix stone drop on death

linaori:
I've added an event listener for `on_damage_taken`, when a rock reaches 0 damage, it will use the same trick as the room generator. It will fire the event where it simulates the death of the entity and then destroy it to prevent loot spill.


mheguy (approved):
Launched in debug with cheats, ran around spawning a whole bunch of biters and let them go to town on the rocks.
It's a shame there's no `on_pre_entity_died` event to hook.

PR #394, opened 2018-11-18
Reformatted entire codebase

Valansch:
Style settings are 4 spaces as tabs, unix line endings, single quotes and 120 character lines.

Please dont hold back on any criticism.

PR #396, opened 2018-11-18
Finalize styleguide

### LINKED ISSUE #206: Fix tabs in entire project
I think we should finally agree on a standard.
@grilledham uses 4 spaces. I have been using a bit of everything so i guess most of the inconsistency a due to me.

So can we agree to 4 spaces?

### LINKED ISSUE #336: Use consistent line endings

mheguy:
I recommend we come to a consensus on a style before changing the style of everything in https://github.com/Refactorio/RedMew/pull/394.

Included is what the consensus has generally skewed towards and some fairly standard rules.

Closes #336
Closes #206


docs/STYLEGUIDE.md
```diff
@@ -0,0 +1,51 @@
+
+# RedMew style guide
+
+Not strictly enforced, but we appreciate if you try to remain consistent with our standards.
+
+* Include at least a brief one-line summary in a luadoc (a comment with 3 dashes `---`) before each function
```
SimonFlapse: What about the tags? If your IDE supports it it should generate `@param` tags when starting luadoc with 3 dashes

mheguy: I think having a full luadoc would be nice, how should we word it so that it's clear a full luadoc would be nice but is unnecessary while a one-line luadoc summary is expected?

SimonFlapse: I think the at least part is perfect. Maybe encourage a `@param` description and `@return` if the function returns for non-local functions. That'll help a lot. So much yet to learn and figure out because this is absent.

plague006:
Link to see what it looks like in-browser: https://github.com/plague006/RedMew/blob/style/docs/STYLEGUIDE.md

SimonFlapse:
It gives us something to continue working on.

grilledham (approved):

iltar:
Can we add a rule to always have the last item of table, when split over multiple lines, have a trailing comma? This makes copy-paste issues when adding new lines a lot less likely and reduces the diffs slightly:
```diff
{
    'item1',
-    'item2'
+    'item2',
+    'item3'
}
````
vs
```diff
{
    'item1',
    'item2',
+    'item3',
}
````
A lot of IDEs have a "duplicate line", which simply duplicates it, but missing the last `,`, it will have 2 lines in a row missing a `,`, which sometimes breaks formatting and can be quite annoying. A lot of inexperienced people who use tools like notepad++ might not even notice the missing `,` and simply run into errors and ask us what's wrong.

File: docs/STYLEGUIDE.md [hunk 2]
```diff
@@ -0,0 +1,65 @@
+
+# RedMew style guide
+
+Not strictly enforced, but we appreciate if you try to remain consistent with our standards.
+
+* Include at least a brief one-line summary in a LuaDoc (a comment with 3 dashes `---`) before each function
```
Valansch: I don't like the wording here. Could you change it to say something like "each public/module function". I disagree that every function needs a LuaDoc entry.

Valansch (changes requested):
Like most of it. I just have one minor issue

docs/STYLEGUIDE.md
linaori: It doesn't necessarily need a description, but it would be nice to have parameter information documented for all public facing functions (I don't care as much about private)

Valansch: I agree. That was what i was trying to say :)

mheguy: Will make that change change.

Valansch: Very good good.

Valansch (approved):

PR #398, opened 2018-11-18
Split walkabout from custom_commands

### LINKED ISSUE #22: Disable /kill for players on a walkabout

plague006:
Fix walkabout edge cases:
Players on walkabout being able to /kill themselves to return to base.
Players logging out while on walkabout not having their force, character, or position restored when logging back in.

3a83d41 was tested offline solo and online with a target without issues. Solo testing used some code tweaks to be feasible.
f2fbd45 is untested.

Closes #22

### DIFF
--- a/features/walkabout.lua
+++ b/features/walkabout.lua
@@ -1,0 +1,157 @@
+local global ={}
+local Task = require 'utils.Task'
+local Game = require "utils.game"
+local Event = require "utils.event"
+local Token = require 'utils.global_token'
+local Utils = require 'utils.utils'
+
+global.walking = {}
+global.walking_storage = {}
+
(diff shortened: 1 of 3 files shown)


features/walkabout.lua
```diff
@@ -0,0 +1,158 @@
...
+
+--- Returns a player from walkabout after the timeout.
+-- If the player is no longer logged in, store their info
+-- in the global table walking_storage
+local custom_commands_return_player =
+    Token.register(
+    function(args)
+        local player = args.player
+        local index = player.index
+        if not player.valid then
```
iltar: I think `player.index` will already fail if the player is not valid. It's safe to move it below this if

PR #402, opened 2018-11-19
Diggy - New leveling system

SimonFlapse:
Transition of old Diggy level system to new based on ForceControl.lua
Allows removing stone requirements.

- [x] Setup basic control

- [x] Transfer formula

- [x] Handle awarding experience
    - Mining stone (Needs tweaking)
    - Killing biters/worms (Needs tweaking)
    - Research? (Needs tweaking. All research, except for infinity, gives almost 5 million XP)
    - Rocket launches? (Needs tweaking, and testing)

- [x] Implement buffs
    - Mining boost
    - Inventory
    - Health

- [x] Implement market items unlocking

- [x] Implement GUI interface

- [x] Change market prices (Initial - Needs tweaking)

- [x] Documentation

- [x]  DiggyHole.lua testing (Removed inventory full check)

- [x] Write CHANGELOG.md


features/force_control.lua
```diff
@@ -160,7 +160,7 @@ function ForceControl.register_force(lua_force_or_name)
     forces[force.name] = {
         current_experience = 0,
         current_level = 0,
-        experience_level_up_cap = next_level_cap_calculator.execute(0),
+        experience_level_up_cap = next_level_cap_calculator.execute(1),
```
linaori: Current level is 0, so if you want to start level 1, the current_level should also be 1. Either way is fine for me, but they should be kept in sync I think

features/force_control.lua
```diff
@@ -257,4 +257,18 @@ function ForceControl.get_force_data(lua_force_or_name)
...
+    if not force then
+        return
+    end
+
+    local force_config = forces[force.name]
+    if not force_config then
+        return
+    end
+
+    return 'Current experience: ' .. force_config.current_experience .. ' Current level: ' .. force_config.current_level .. ' Next level at: ' .. force_config.experience_level_up_cap .. ' Percentage to level up: ' .. ((force_config.current_experience / force_config.experience_level_up_cap) * 100)
```
linaori: `force_config.experience_percentage` can also be used (it's calculated in the function above)

map_gen/Diggy/Feature/Experience.lua
```diff
@@ -0,0 +1,57 @@
...
+-- this
+local Experience = {}
+
+local Config = {}
+local floor = math.floor
+local ceil = math.ceil
+local log10 = math.log10
+local force = 'player'
+
+function Experience.calculate_level(level) -- all configurable variables must be integers.
```
linaori: I don't think this function has to be publicly available

map_gen/Diggy/Feature/Experience.lua
```diff
@@ -0,0 +1,57 @@
...
+    local number = b*(level^3)+(start_value-b)
+    return number
+end
+
+local function on_player_mined_entity(event)
+    local entity = event.entity
+    if ((entity.name ~= 'sand-rock-big') and (entity.name ~= 'rock-huge')) then
+        return
+    end
+    local exp = math.random(19,24)
```
linaori: Can we make this a stable average number? As the player doesn't loot stone, we can also show this in a message via one of the Game print floating text functions. You can use it to show `+x experience`. Would show it at the rock, not the player. The same can be done for killing biters.

SimonFlapse: Gotcha. So no random experience for anything

features/force_control.lua
SimonFlapse: Yeah, I'm just having a problem with leveling up. Since level 0's cap is at 25 and not 50 with my formula. It seems that it calculates the upper bound of a level. But I know how to go around this now.

features/force_control.lua
SimonFlapse: It's not inside that scope. Just returns nil

linaori: So what if you do:
```lua
function ForceControl.get_formatted_force_data(lua_force_or_name)
    local force_config = ForceControl.get_force_data(lua_force_or_name)
    if not force_config then
        return
    end

    return 'Current experience: ' .. force_config.current_experience .. ' Current level: ' .. force_config.current_level .. ' Next level at: ' .. force_config.experience_level_up_cap .. ' Percentage to level up: ' .. force_config.experience_percentage .. '%'
end
```

features/force_control.lua
```diff
@@ -82,8 +82,9 @@ function ForceControlBuilder.register(level_matches, callback, lua_force_name)
 
     local function on_level_up(event)
         local level = event.level_reached
-        if level_matches(level, event.force) then
-            callback(level, event.force)
+        local force = get_valid_force(event.force)
+        if level_matches(level, force) then
```
linaori: Won't this prevent the `LuaForce` from reaching the callback? Meaning you can't directly call it anymore

map_gen/Diggy/Feature/Experience.lua
```diff
@@ -0,0 +1,233 @@
...
+    Event.add(defines.events.on_research_finished, on_research_finished)
+    Event.add(defines.events.on_rocket_launched, on_rocket_launched)
+    Event.add(defines.events.on_player_respawned, on_player_died)
+
+end
+
+function Experience.on_init()
+    local force = game.forces.player
+    ForceControl.register_force(force)
+    log(force.name)
```
linaori: Don't forget 😄

linaori (commented):
Looking good so far 😄

features/force_control.lua
SimonFlapse: It's not meant to be there, I used it to test why I couldn't add buffs to the force.

map_gen/Diggy/Feature/Experience.lua
```diff
@@ -101,12 +100,9 @@ local function on_player_mined_entity(event)
...
     local text = ''
-    local exp = 0
-    if entity.name == 'sand-rock-big' then
-        exp = 5
-        text = '+' .. exp .. XP_text
-    elseif entity.name == 'rock-huge' then
-        exp = 10
+    local exp
+    if entity.name == 'sand-rock-big' or entity.name == 'rock-huge' then
+        exp = Config.XP['' .. entity.name]
```
linaori: This looks a bit like an odd structure. What about `Config.XP[entity.name]` instead? Another option would be this:
```lua
local sand_rock_xp = Config.XP['sand-rock-big']
local huge_rock_xp = Config.XP['rock-huge']
local function on_player_mined_entity(event)
    if entity.name == 'sand-rock-big' then
        exp = sand_rock_xp
    end
    if entity.name == 'rock-huge' then
        exp = huge_rock_xp
    end
end
```
This has the benefit of no deep index lookup when it dies (500k+ in a big map). If you still have issues with the floating text, we could also drop coins instead of award them.

map_gen/Diggy/Feature/Experience.lua
```diff
@@ -176,23 +157,23 @@ local function on_rocket_launched(event)
...
-local function on_player_died(event)
-    local force = event.force
-    ForceControl.remove_experience(force, 50)
+local function on_player_respawned(event)
     local player = Game.get_player_by_index(event.player_index)
+    local force = player.force
 
     ForceControl.get_force_data(force)
-
-    local exp = ForceControl.get_force_data(force).total_experience*0.005
```
linaori: What if we add a function in `ForceControl` for this?
```lua
local exp_removed = ForceControl.remove_experience_percentage(death_penalty)
```

linaori (commented):
Can you add 'stone' to the item removed when mining in DiggyHole as well? (look for 'coal' and you'll find it)

map_gen/Diggy/Feature/Experience.lua
SimonFlapse: Yeah you're right, I'll define all these configurable XP on register to prevent all the table lookups for the exact same value.

map_gen/Diggy/Feature/Experience.lua
SimonFlapse: Alright :) Generic function anyways

features/force_control.lua
```diff
@@ -200,8 +201,33 @@ function ForceControl.remove_experience(lua_force_or_name, experience)
...
+    local min_experience = min_experience ~= nil and min_experience or 0
+    local force = get_valid_force(lua_force_or_name)
+    if not force then
+        return
+    end
+    local force_config = forces[force.name]
+    if not force_config then
+        return
+    end
+    local ceil = math.ceil
```
linaori: Can push this to the top of the file, it will only assign the reference once instead of every time it enters this scope

features/force_control.lua
```diff
@@ -200,8 +201,33 @@ function ForceControl.remove_experience(lua_force_or_name, experience)
     if not force_config then
         return
     end
-
+    local backup_current_experience = force_config.current_experience
     force_config.current_experience = max(0, force_config.current_experience - experience)
+    force_config.total_experience = (force_config.current_experience == 0) and force_config.total_experience - backup_current_experience or max(0, force_config.total_experience - experience)
+end
+
+---Removes experience from a force, based on a percentage of the total experience
```
linaori: total experience obtained, or total experience needed to level up?

features/force_control.lua
```diff
@@ -200,8 +201,33 @@ function ForceControl.remove_experience(lua_force_or_name, experience)
...
+    end
+    local force_config = forces[force.name]
+    if not force_config then
+        return
+    end
+    local ceil = math.ceil
+
+    local penalty = force_config.total_experience * percentage
+    penalty = (penalty >= min_experience) and ceil(penalty) or ceil(min_experience)
+    ForceControl.remove_experience(lua_force_or_name, penalty)
```
linaori: What if we move the removed XP to the remove_experience function? Then we could use the same functionality for that scope and just return the actual XP removed directly

map_gen/Diggy/Config.lua
```diff
@@ -298,7 +298,7 @@ local Config = {
             -- market config
             market_spawn_position = {x = 0, y = 3},
             stone_to_surface_amount = 50,
-            currency_item = 'stone',
+            currency_item = 'coin',
```
linaori, line 346: I think we can hard-code this one by now 😄

map_gen/Diggy/Config.lua
```diff
@@ -310,52 +310,42 @@ local Config = {
...
-                    {level = 58, price = {{"stone", 1500}, {"coin", 25}}, name = 'rocket-launcher'},
-                    {level = 58, price = {{"stone", 500}, {"coin", 5}}, name = 'rocket'},
-                    {level = 66, price = {{"stone", 1000}, {"coin", 10}}, name = 'explosive-rocket'},
-                    {level = 73, price = {{"stone", 1000}, {"coin", 200}}, name = 'satellite'},
-                    {level = 100, price = {{"stone", 5000}, {"coin", 9999}}, name = 'atomic-bomb'},
+                    {level = 58, price = 250, name = 'rocket-launcher'},
+                    {level = 58, price = 50, name = 'rocket'},
+                    {level = 66, price = 100, name = 'explosive-rocket'},
+                    {level = 73, price = 2000, name = 'satellite'},
+                    {level = 100, price = 1, name = 'iron-stick'},
```
linaori, line 391: 😂

map_gen/Diggy/Feature/DiggyHole.lua
```diff
@@ -254,6 +254,7 @@ function DiggyHole.register(config)
 
         if name == 'sand-rock-big' or name == 'rock-huge' then
             event.buffer.remove({name = 'coal', count = 100})
+            event.buffer.remove({name = 'stone', count = 100})
 
             -- this logic can be replaced once we've fully replaced the stone to surface functionality
             if enable_digging_warning then
```
linaori, line 257: Can this warning be removed now? I think we don't use it anywhere anymore once this feature is finished

map_gen/Diggy/Feature/Experience.lua
```diff
@@ -0,0 +1,247 @@
...
+
+local Config = {}
+local floor = math.floor
+local XP_text = ' XP'
+
+function Experience.update_mining_speed(force, level_up)
+    local buff = Config.buffs['mining_speed']
+    if level_up > 0 and buff ~= nil then
+        local value = (buff.double_level ~= nil and level_up%buff.double_level == 0) and buff.value*2 or buff.value
+        mining_efficiency.level_modifier = mining_efficiency.level_modifier + (value / 100)
```
linaori: Minor thing so it won't matter much, but it's interesting to do nevertheless. In lua multiplication is faster than division, so you can also do `(value * 0.01)`: http://lua-users.org/wiki/OptimisationCodingTips

Performance is probably neglect-able though

linaori (commented):
Looking good so far, make sure to not forget to test it in heavy mode

map_gen/Diggy/Feature/DiggyHole.lua
SimonFlapse: It very much can be removed :)

map_gen/Diggy/Config.lua
SimonFlapse: Oh yeah, I needed something for level 100 🗡

map_gen/Diggy/Config.lua
SimonFlapse: Doesn't hurt to give people the opportunity to change it?

linaori: It's fine either way, just makes it close to useless as it wasn't designed for it. Where's it actually still used? 🤔

features/force_control.lua
SimonFlapse: Not sure I fully understand, The reason i'm returning the value is that I needed it to display the floating text. And since you'd normally supply it with experience to add or remove you'd already know how much that is, using the other functions. The ceil part is because it's a percentage and I'd like to avoid subtracting weird fractions.

features/force_control.lua
SimonFlapse: Total experience obtained

map_gen/Diggy/Config.lua
SimonFlapse: It's used when you don't specify what the price of an item is in, it'll automatic assign it to currency_item

map_gen/Diggy/Feature/DiggyHole.lua
SimonFlapse: I'd need you to take look at DiggyHole.lua it see if I removed something wrong

features/force_control.lua
linaori: I'm referring to doing this: `return ForceControl.remove_experience(lua_force_or_name, penalty)`. This function already calculates the difference between how much you wanted to subtract and how much you _can_ subtract until your reach 0. If you let that function return that number, you only have to return that in here and not worry about the difference between going from (say) 25 xp - 50 to 0 or -25.

map_gen/Diggy/Feature/Experience.lua
```diff
@@ -164,6 +168,8 @@ local function on_research_finished(event)
 
     Experience.update_inventory_slots(force, 0)
     Experience.update_mining_speed(force, 0)
+
+    game.forces.player.technologies['landfill'].enabled = false
```
linaori, line 195: Could be moved up to where the atomic bomb is disabled (and use `force.technologies`

linaori (approved):
I think it's almost if not completely ready to be merged?

map_gen/Diggy/Feature/Experience.lua
SimonFlapse: But we don't need to set that more than once? The other it just to disable it when the research is researched. I could implement a check for if the nuclear research was the one being researched and only disable it then. But the current solution works :D

features/force_control.lua
SimonFlapse: Okay I understand now. Yeah I'll like that. That way I'm not relaying wrong information to the players via the floating text if we can't remove all the experience

SimonFlapse:
Fails heavy-mode

SimonFlapse:
This is causing me to fail heavy-mode. It's from the old diggy market (Seems to spawn it after 50 ticks). Removing it makes me pass heavy-mode
```lua
local on_market_timeout_finished = Token.register(function(params)
    Template.market(params.surface, params.position, params.player_force, {})
end)
```
```lua
function MarketExchange.on_init()
    Task.set_timeout_in_ticks(50, on_market_timeout_finished, {
        surface = game.surfaces.nauvis,
        position = config.market_spawn_position,
        player_force = game.forces.player,
    })
end
```
Okay so Template.market raises an event that triggers this function:
```lua
local function on_placed_entity(event)
    local market = event.entity
    if 'market' ~= market.name then
        return
    end
    diggy_market = market
end
```
It seems that me assigning the market to the local variable diggy_market causes the error

SimonFlapse:
~~Researching (all) technologies~~ leveling up causes a failure in heavy mode too

First gone gives the below error

```
1998.138 Error MainLoop.cpp:1178: Heavy mode check after-gameactionhandler-update-207 failed.
1999.869 Error MainLoop.cpp:1178: Heavy mode check after-scenario-update-208 failed.
```

This one/the second one gives
```
2902.875 Error MainLoop.cpp:1178: Heavy mode check after-scenario-update-880 failed.
2904.496 Error MainLoop.cpp:1189: Heavy mode heuristic check after-scenario-update-880 failed.
```

This part passes heavy-mode (Not configurable)
```lua
function Experience.register(cfg)
    Config = cfg
    local b = floor(Config.difficulty_scale) or 25 -- Default 25 <-- Controls how much stone is needed.
    local start_value = floor(Config.start_stone) or 50 -- The start value/the first level cost

    --Adds the function on how to calculate level caps (When to level up)
    ForceControl_builder = ForceControl.register(function (level_reached)
        if level_reached ~= 0 then
            return 25*((level_reached+1)^3)+(50-25) - (25*((level_reached)^3)+(50-25))
        else
            return 25*((level_reached+1)^3)+(50-25)
        end
    end)
end
```

This part doesn't pass heavy-mode (Configurable)
```lua
function Experience.register(cfg)
    Config = cfg
    local b = floor(Config.difficulty_scale) or 25 -- Default 25 <-- Controls how much stone is needed.
    local start_value = floor(Config.start_stone) or 50 -- The start value/the first level cost

    --Adds the function on how to calculate level caps (When to level up)
    ForceControl_builder = ForceControl.register(function (level_reached)
        if level_reached ~= 0 then
            return b*((level_reached+1)^3)+(start_value-b) - (b*((level_reached)^3)+(start_value-b))
        else
            return b*((level_reached+1)^3)+(start_value-b)
        end
    end)
end
```
Outputs the following error only when heavy-mode is on 
(Line 248 is `return b*((level_reached+1)^3)+(start_value-b) - (b*((level_reached)^3)+(start_value-b))`)

`4070.717 Script @C:/Users/simon/AppData/Roaming/Factorio/temp/currently-playing/utils/event.lua:22: ...p/currently-playing/map_gen/Diggy/Feature/Experience.lua:248: attempt to perform arithmetic on local 'b' (a nil value)`

linaori:
@Valansch would this issue be fixed by using a token instead of a global inside the force control?

SimonFlapse:
Made a workaround, it passes heavy mode now

map_gen/Diggy/Feature/Experience.lua
```diff
@@ -0,0 +1,296 @@
...
+    end
+    ForceControl.add_experience(force, exp)
+end
+
+---Awards experience when a player kills an enemy, based on type of enemy
+---@param event LuaEvent
+local function on_entity_died (event)
+    local entity = event.entity
+    local force = entity.force
+    local exp = 0
```
linaori: local assignment can go below the force.name check

map_gen/Diggy/Feature/Experience.lua
```diff
@@ -0,0 +1,296 @@
...
+    return Config.buffs
+end
+
+local level_table = {}
+---Get experiment requirement for a given level
+---Primarily used for the market GUI to display total experience required to unlock a specific item
+---@param level number a number specifying the level
+---@return number required total experience to reach supplied level
+function Experience.calculate_level_xp(level)
+        if level_table[level] == nil then
```
linaori: indent

linaori (approved):
Looks good!

Valansch:
@stickler-ci shut up :D

mheguy:
Rebase the PR and let the bot do its thing :p

map_gen/Diggy/Feature/Experience.lua
```diff
@@ -0,0 +1,295 @@
...
+    Config = cfg
+
+    --Adds the function on how to calculate level caps (When to level up)
+    ForceControl_builder = ForceControl.register(level_up_formula)
+
+    --Adds a function that'll be executed at every level up
+    ForceControl_builder.register_on_every_level(function (level_reached, force)
+        force.print(string_format('Leved up to %d!', level_reached))
+        force.play_sound{path='utility/new_objective', volume_modifier = 1 }
+        local Experience = require 'map_gen.Diggy.Feature.Experience'
```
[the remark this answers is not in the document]
SimonFlapse: This is required for the function to work

mheguy:
Don't take stickler's warnings as gospel. If something is harmless and you just don't have the time or energy to change it: it's fine.
Things like defining non-standard globals are important to address though.

map_gen/Diggy/Feature/Experience.lua
```diff
@@ -0,0 +1,295 @@
...
+    if force.name ~= 'enemy' then
+        return
+    end
+
+    local cause = event.cause
+
+    if not cause or cause.type ~= 'player' or not cause.valid then
+        return
+    end
+    local exp = 0
```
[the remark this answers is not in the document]
Valansch: My bad, this can be removed

features/force_control.lua
```diff
@@ -200,8 +199,32 @@ function ForceControl.remove_experience(lua_force_or_name, experience)
...
+end
+
+---Removes experience from a force, based on a percentage of the total obtained experience
+---@param lua_force_or_name LuaForce|string
+---@param percentage number percentage of total obtained experience to remove
+---@param min_experience number minimum amount of experience to remove (optional)
+---@return number the experience being removed
+---@see ForceControl.remove_experience
+function ForceControl.remove_experience_percentage(lua_force_or_name, percentage, min_experience)
+    local min_experience = min_experience ~= nil and min_experience or 0
```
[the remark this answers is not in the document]
SimonFlapse: It's supposed to be shadowed

SimonFlapse:
Passes heavy mode and should be ready to go. Fixed all relevant stickler-ci comments

PR #404, opened 2018-11-19
Fish market: split messages out, cover cases of no character and disconnected player

### LINKED ISSUE #224: Market speed buffs assumes each Player always has a Character.
We've been encountering this error in our logs recently:

> 62817.338 Script .../utils/event.lua:22: No character.
> stack traceback:
> 	.../fish_market.lua:223: in function 'reset_player_runningspeed'
> 	.../fish_market.lua:377: in function <.../fish_market.lua:372>
> 	[C]: in function 'pcall'
> 	.../utils/event.lua:20: in function 'call_handlers'
> 	.../event.lua:45: in function <.../utils/event.lua:43>

It looks like we had a player buy a speed buff, die before the buff expired, and then leave the game before respawning. They still have an entry in the `global.player_speed_boost_records` table.

When fish_market.lua changes `character_running_speed_modifier` it does so using the `LuaPlayer` object as its reference. According the [Lua documentation](https://lua-api.factorio.com/latest/LuaControl.html#LuaControl.character_running_speed_modifier) this can fail as a `Player` doesn't always have to have a `Character`. 

What it looks like we have is a `Player` in the `global.player_speed_boost_records` table, that doesn't have a `Character` associated. This causes an error when trying to set their `character_running_speed_modifier`, which event.lua's `pcall` protects us from and logs. This runs every 3 seconds and fills our logs.

A copy of the save can be found in our public read FTP. [Here](https://vault.win-serv.com/publicstorage/DYNAMIC-NETWORKS-READ-ONLY/GameServ/Factorio.one/01d/saves/_gsdata_/_history_/_autosave1_2018-0927-114556.zip)

The player in question is player index 12, they have a `ticks_to_respawn` value and no inventory so I believe they are dead. 

A quick fix I tested is to put a check if the player has a character around setting the value each time. IE:
```
if (player.character) then
    player.character_running_speed_modifier = global.player_speed_boost_records[player.index].pre_boost_modifier
    global.player_speed_boost_records[player.index] = nil
end
```
Though this would be cumbersome in the long run.

This applies to quite a few properties that are shared between `Character` and `Player` but really only apply to `Character`.

What should be done about this? I can probably make the changes and submit a pull with a bit of direction.

plague006:
Tested having no character and disconnecting/reconnecting in MP and SP with debug on.

Closes #224


features/fish_market.lua
```diff
@@ -384,19 +329,27 @@ local function on_180_ticks()
         if global.player_speed_boost_records then
             for k, v in pairs(global.player_speed_boost_records) do
                 if game.tick - v.start_tick > 3000 then
-                    reset_player_runningspeed(Game.get_player_by_index(k))
+                    local player = Game.get_player_by_index(k)
+                    if player.connected and player.character then
```
iltar: Should we also check for `player.valid` at this point? What if I call `player.destroy()` and this triggers without it?

plague006: Checking player.valid here is moot. `get_player_by_index` is either returning a `LuaPlayer` or `nil`, and the only reason it would be `nil` is if `global.player_speed_boost_records` got corrupted or got fed a bad entry, in which case those should be addressed.

Checking character.valid might be worth exploring, but we have to check player.character before we can check player.character.valid anyways. So if the player has a character, under what circumstances would that character not be valid?

plague006: `die()` and `destroy()` both remove the association between the player and character, so in those cases player.character is nil

plague006: So in the cases of both the character and the player: they're not objects we're storing, they're calls to the API and those should never return invalid objects.

PR #406, opened 2018-11-19
Add guidelines for contributing to redmew

mheguy:
Mostly just copied from an old project. Would happily change the wording of any part.


linaori (approved):

linaori:
We should look into linking these files in the readme as well 👍

PR #408, opened 2018-11-19
Improved biter spawning algorithm

linaori:
The algorithm no longer relies on `surface.find_non_colliding_position()`, but tries to find a location based on biter size + some extra margin, away from void. Once spawned, it tries to mine the rocks on that location to free it. It's kinda boring atm because it's not like it breaks free, but we can add that some other time.

### DIFF
--- a/map_gen/Diggy/Config.lua
+++ b/map_gen/Diggy/Config.lua
@@ -113,21 +113,27 @@ local Config = {
             minimal_treasure_chest_distance = 25,
 
             -- chances to receive a coin when mining
-            mining_artefact_chance = 0.10,
-            mining_artefact_amount = {min = 1, max = 4},
+            mining_coin_chance = 0.10,
+            mining_coin_amount = {min = 1, max = 4},
 
             -- lets you set the coin modifiers for aliens
             -- the modifier value increases the upper random limit that biters can drop
             alien_coin_modifiers = {
-                ['small-biter'] = 1,
-                ['small-spitter'] = 1,
-                ['medium-biter'] = 2,
-                ['medium-spitter'] = 2,
-                ['big-biter'] = 4,
-                ['big-spitter'] = 4,
-                ['behemoth-biter'] = 6,
-                ['behemoth-spitter'] = 6,
+                ['small-biter'] = 2,
+                ['small-spitter'] = 2,
+                ['small-worm'] = 2,
+                ['medium-biter'] = 3,
+                ['medium-spitter'] = 3,
+                ['medium-worm'] = 3,
+                ['big-biter'] = 5,
+                ['big-spitter'] = 5,
+                ['big-worm'] = 5,
+                ['behemoth-biter'] = 7,
+                ['behemoth-spitter'] = 7,
             },
+
+            -- chance of aliens dropping coins between 0 and 1, where 1 is 100%
+            alien_coin_drop_chance = 0.30,
 
             -- shows the chest locations, only use when debugging
             display_chest_locations = false,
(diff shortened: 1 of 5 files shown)


map_gen/Diggy/Config.lua
```diff
@@ -36,7 +36,7 @@ local Config = {
                 character_inventory_slots_bonus = 1000,
                 character_running_speed_modifier = 2,
                 starting_items = {
-                    {name = 'modular-armor', count = 1},
+                    {name = 'power-armor-mk2', count = 1},
```
SimonFlarup: Maybe make it clear for the player that they can choose? Like have a couple of options commented out that they can enable? Some might think MK2 is very powerful to begin with.

SimonFlapse, line 39: Maybe make it clear for the player that they can choose? Like have a couple of options commented out that they can enable? Some might think MK2 is very powerful to begin with.

linaori: I'll make a note about it, something we can document a lot better in general here

map_gen/Diggy/Feature/AlienSpawner.lua
```diff
@@ -37,18 +133,27 @@ end
 function AlienSpawner.register(config)
     local alien_minimum_distance_square = config.alien_minimum_distance ^ 2
 
+    local alien_probability = config.alien_probability
+
     Event.add(Template.events.on_void_removed, function (event)
-        game.forces.enemy.evolution_factor = game.forces.enemy.evolution_factor + 0.0000012
+        local force = game.forces.enemy
+        local evolution_factor = force.evolution_factor
+        force.evolution_factor = evolution_factor + 0.0000012
```
SimonFlapse, line 190: Maybe we should move this value out into the config?

linaori: probably a good idea, but I'll leave it in here for now. As discussed earlier on Discord, we can fix it when we disable pollution as well

linaori:
I've added some... challenge to mining, especially end-game.

SimonFlapse (approved):

map_gen/Diggy/Config.lua
```diff
@@ -285,7 +291,30 @@ local Config = {
...
 
             -- chance of spawning aliens when mining
-            alien_probability = 0.07,
+            alien_probability = 0.05,
+
+            -- spawns the following units when they die. To disable change it to:
+            --hail_hydra = nil,
+            -- any non-rounded number will turn into a chance to spawn an additional alien
+            -- example: 2.5 would spawn 2 for sure and 50% chance to spawn one additionally
+            hail_hydra = {
```
Valansch, line 300: hail_hydra? WTF :D

linaori: hydra mode, cut 1 down, 2 more appear

Valansch: More like Medusa mode^^

Valansch:
Looking good all round. I like the hydra mode. Next thing to implement; banelings: Small biters that were spawned from bigger biters explode on death!

mheguy:
I think this is the 3rd place in code where larger aliens spawn smaller aliens on death. We should break this out into a base feature.

linaori:
@plague006 I agree, though this spawns more than just smaller aliens. I can look into making it a generic feature if you can point me to the code that does this

mheguy:
I know of:
https://github.com/Refactorio/RedMew/blob/7f68f0ba9250c858bb4ef53b9dd702b37fc4f4df/map_gen/presets/crash_site/entity_died_events.lua

In a recent version of crash site we did, I adapted some of that code to spawn smaller versions of existing biters. When I wrote it, I was asked if I used code that was already in the codebase, but don't remember which map was cited.

mheguy:
Ultimately, diggy has a bunch of functionality that would be nice if they were converted for use in any map, so this one thing isn't worth making a specific fuss over.

linaori:
If we can identify what we need, we can start working on that when we have time/feel like it 😄

PR #410, opened 2018-11-20
Add changelog

mheguy:
Moving forward it would be nice if, as part of merges, we worked together to keep the changelog up to date.


linaori (approved):
Nice! Let's try to keep this up-to-date

PR #413, opened 2018-11-20
Root cleanup

mheguy:


map_layout.lua
```diff
@@ -154,9 +153,9 @@ if #terrain_modules > 0 then
 end
 
 if shape then
-	local surfaces = {
-		['nauvis'] = shape,
-	}
+    local surfaces = {
+        ['nauvis'] = shape
```
linaori: Should the trailing `,` stay here? As per https://github.com/Refactorio/RedMew/pull/396/commits/b1465ff2a6660e189f07281d082d4ac8a7b675d9

mheguy: Will change that back.

linaori (approved):

PR #417, opened 2018-11-20
Added /find command to act the same as find_player

Jayefuu:
/find isn't used for anything else and /find-player takes longer to type when looking for a player who's being accused of griefing. This will shorten the time and be more intuitive. Left in /find-player so that people who are used to it can still use it.


mheguy (commented):
I think changing from /find-player to /find makes sense.
We spoke on discord and you mentioned that you'll modify this PR before merging so will wait on that before moving forward.

SimonFlapse:
I agree on the renaming of /find-player to only /find

Valansch (changes requested):
Remove find-player

Jayefuu:
No problem. I will make changes that will print a warning that /find-player is deprecated and will be removed in January 2019.

mheguy (commented):
Wish you could delete these reviews. -_-

mheguy (changes requested):
In find_player, need to check for player before calling player.print

mheguy:
Will commit against this to tighten it up and resolve the case of `/regular demote`

Valansch (approved):

Valansch:
Looking good.

Issue #419, opened 2018-11-21
Fix creep_spread


mheguy:
There are more linting warnings in creep_spread than I can count. I practically need to rewrite it from the ground up.

linaori:
ouch 😅

mheguy:
Closed via https://github.com/Refactorio/RedMew/pull/524

PR #420, opened 2018-11-21
Cleanup map_gen/misc

mheguy:
Formatting only, linting warnings unaddressed

### DIFF
--- a/map_gen/day_night_cycles/day_night_cycle.lua
+++ b/map_gen/day_night_cycles/day_night_cycle.lua
@@ -1,4 +1,4 @@
-local Event = require "utils.event"
+local Event = require 'utils.event'
 
 -- Uncomment the desired day/night cycle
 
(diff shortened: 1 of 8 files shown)


map_gen/misc/creep_spread.lua
```diff
@@ -135,23 +134,28 @@ local all_tiles ={
...
     'water',
-    'water-green'
+    'water-green',
 }
 
 local function get_tiles(tile_filter)
-    tiles = find_tiles_filtered({area={{chunkcoord.x-16, chunkcoord.y-16},{chunkcoord.x+16, chunkcoord.y+16}},
-                name = creep_expandable_tiles})
+    tiles =
+        find_tiles_filtered(
```
linaori: This change looks really odd, I would expect the assignment to be on the same line:

```lua
tiles = find_tile_filtered(
```
I see this a couple of times in this file, sometimes borking the indents, what do you think?

linaori (approved):

map_gen/misc/creep_spread.lua
mheguy: Creep spread in particular needs a lot more love than some formatting changes (#419) so I'll just omit it from this PR.

PR #424, opened 2018-11-21
Added particles when a biter mines a rock upon spawning

linaori:
- Added particles when breaking up rocks, with diminishing amounts for stacked rocks
 - Documented more cheats
 - Added more cheats

Particle effect + hydra mode can be seen here: https://www.youtube.com/watch?v=rSQ3_GkEvRA

### DIFF
--- a/map_gen/Diggy/Config.lua
+++ b/map_gen/Diggy/Config.lua
@@ -5,7 +5,7 @@ local Config = {
     -- enable debug mode, shows extra messages
     debug = false,
 
-    -- allow cheats. Example: by default the player will have X mining speed
+    -- allow cheats, primarily configured under SetupPlayer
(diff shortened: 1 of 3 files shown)


SimonFlapse:
I like the particles. I'll experiment with it and see if I can use it in the force control system. How is performance with it? Is it affected by graphics settings? Could it cause performance issues for low end PC players?

linaori:
I believe there's a setting to hide the particles, but I'm not sure. Testing showed that there was performance overhead, but not even close to spilling about 200 stone.

Valansch (approved):
Cood is good, i like the look of the 'animation' :)

mheguy (approved):
Tested and works well. Looks cool.

PR #425, opened 2018-11-21
Fix/walkabout

mheguy:
I cannot even begin to guess at my thought process when I made `global` a local, empty table.


mheguy:
@SimonFlapse gave it a thumbs up on discord, but was too lazy to actually hit the review button on GitHub :D

PR #427, opened 2018-11-22
Shift command logging to free_item_logging, fix cases of server as actor

mheguy:
Kept logging of jail since free_item_logging won't catch jailing through the GUI.

### DIFF
--- a/features/custom_commands.lua
+++ b/features/custom_commands.lua
@@ -88,7 +88,7 @@ local function kill(cmd)
     end
 
     if global.walking then
-        if (player and global.walking[player.index] == true) or global.walking[target.index] == true then
+        if (player and global.walking[player.index]) or global.walking[target.index] then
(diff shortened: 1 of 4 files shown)


features/custom_commands.lua
```diff
@@ -91,8 +88,8 @@ local function kill(cmd)
     end
 
     if global.walking then
-        if global.walking[player.index] == true or global.walking[target.index] == true then
-            player.print("A player on walkabout cannot be killed by a mere fish, don't waste your efforts.")
+        if (player and global.walking[player.index] == true) or global.walking[target.index] == true then
```
Valansch: == true can be omitted

Valansch (changes requested):
Very nice, just one minor thing. Dismiss this and merge if you disagree with my suggestion

mheguy:
Nah you're totally right, it's cleaner and more sensible.

Valansch (approved):

PR #428, opened 2018-11-22
Add player renaming

mheguy:
Players can be renamed via command (self or admin-targeted) or on player creation. (configurable)
Names are generated randomly by assembling adverb + adjective + noun from the 3 tables in naming_words.lua. Currently there are 329,280 possible names.

I *think* I covered all the bases to avoid nasty situations:

\- Store used silly names to avoid duplicates
\- Reset silly names in case the table gets too full

\- Store a player's actual name in a global table
\- Add function to get player's actual name
\- Add function to get player ID (for purposes of banning, etc)

Sincerely,
Miserably Quaint Snowball


control.lua
```diff
@@ -28,6 +28,7 @@ require 'features.free_item_logging'
 require 'features.nuke_control'
 require 'features.player_colors'
 require 'features.reactor_meltdown'
+require 'features.silly_player_names'
```
linaori: With this feature being eagerly loaded, all scenarios will receive this in memory, even when it's not used.

linaori (commented):
How does banning work with this?

linaori:
I just remembered, this will not work with the notification system due to spaces in names I fear

mheguy:
I'll change to underscores and that should remove any potential issues that spaces might bring

linaori:
Would dashes work? I know dots would 🤔

mheguy:
Do you foresee an issue with underscores?

mheguy:
CamelCaseForTheWin!

linaori:
No, it's just a matter of picking a preferred word separator. I'm sure there are more alternatives 😄 

```
Miserably_Quaint_Snowball
MiserablyQuaintSnowball
Miserably.Quaint.Snowball
Miserably-Quaint-Snowball
Miserably/Quaint/Snowball
Miserably,Quaint,Snowball
Miserably<&nbsp;>Quaint<&nbsp;>Snowball
```

Jayefuu:
This is for one off use for the Christmas themed maps....? Or what? Will it still show their usual name after the three words?

mheguy:
This was written with the winter holidays in mind, but as much as possible, I try to write things to be open for use in multiple contexts. So for example with a bit of tweaking we could enable halloween names or dwarf names for diggy or valentines or really any set of names.

All that has to be worked on is basically getting words into categories.
Like `common` words and `winter holiday` words and `dwarf` words.
Maybe I'll change the tables around to use a tagging system.

This will actually change people's names. It introduces the command `/name-check` to find a person's "real" name, and a command to find a player's index. It also shows admins the index of players changing their names (since, for example, you can jail based on index). Players can `/name-roll` themselves a new name or admins can `/name-roll target`, and lastly anyone can `/name-restore` to get their boring old username back.

Jayefuu:
I think you should append their username to the end of the 3 words, in brackets maybe. It's going to be really annoying not knowing who is who as I associate different levels of competency, friendlyness, etc with each player nickname so would want to continue to see who is who.

Valansch (changes requested):
Logo

mheguy:
I hear you Jay, but I think we can give this a single run and see if it works or has issues. If it seems too confusing we can add the player's name or just re-think the naming scheme to incorporate the players' names.

Valansch (approved):

Issue #426, opened 2018-11-22
Error when using console on new server inteface


mheguy:
427.329 Script @/factorio/2/temp/currently-playing/utils/event.lua:22: /factorio/2/temp/currently-playing/utils/game.lua:21: bad argument #2 to '__index' (string expected, got nil)

Arises from sending text from the server console into the game.

Valansch:
I saw that too, not sure wahts causing it the error msg is not helpful. Needs investigating!

mheguy:
It's coming from chat_triggers

PR #431, opened 2018-11-22
Fix null ref (event.player_index nil if server)

mheguy:


Valansch:
Good stuff good stuff

Valansch:
closes #426

SimonFlapse (approved):

PR #434, opened 2018-11-22
Added bot mining

linaori:
Bot behavior is a bit quirky at times, but it seems to work just fine. Also fixed a bug that let bots mine tiles anyway, despite being support beams.

You can see how it works here: https://www.youtube.com/watch?v=7bJtkjlgCnM

### DIFF
--- a/map_gen/Diggy/Feature/DiggyCaveCollapse.lua
+++ b/map_gen/Diggy/Feature/DiggyCaveCollapse.lua
@@ -231,15 +231,6 @@ local function on_player_mined_tile(event)
     end
 end
 
-local function on_robot_mined_entity(event)
-    local entity = event.entity
-    local strength = support_beam_entities[entity.name]
-
-    if strength then
-        stress_map_add(entity.surface, entity.position, strength, false, entity.last_user.index)
-    end
-end
-
 local function on_mined_entity(event)
     local entity = event.entity
     local name = entity.name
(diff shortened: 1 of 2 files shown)


Jayefuu:
Yay! I really dig this feature. Are you doing anything balance-wise to make them super slow to start with, with tech upgrades improving it a lot?

map_gen/Diggy/Feature/DiggyCaveCollapse.lua
```diff
@@ -360,7 +350,12 @@ function DiggyCaveCollapse.register(cfg)
     Event.add(defines.events.on_surface_created, on_surface_created)
 
     Event.add(defines.events.on_marked_for_deconstruction, function (event)
-        if (nil ~= support_beam_entities[event.entity.name]) then
+        local name = event.entity.name
+        if name == 'sand-rock-big' or name == 'rock-huge' then
+            return
+        end
+
+        if name == 'deconstructible-tile-proxy' or nil ~= support_beam_entities[name] then
```
Valansch, line 358: `nil ~= support_beam_entities[name]`
can just be
`not support_beam_entities[name]`

Valansch (approved):
I didnt understand why you are creating the rock again after removing it. So i tried it. THIS IS AWSOME

plague006:
Tested this some: rooms generate fine, biters generate, collapses work. LGTM

mheguy (approved):
Tested this some: rooms generate fine, biters generate, collapses work. LGTM

map_gen/Diggy/Feature/DiggyCaveCollapse.lua
linaori: old code, will fix this

linaori:
@Jayefuu later this evening I will work on the following:
 - set the base damage to 4 (10 atm)
 - increase the base damage by 1 per mining productivity level. That will make it roughly 7 by the time you can make bots I think

mheguy:
Interesting, I like the idea of increasing the value of mining productivity research.

PR #435, opened 2018-11-22
Add stickler_ci

mheguy:
luacheckrc ripped nearly verbatim from https://github.com/Nexela/Factorio-luacheckrc (thanks @Nexela)

Added some redmew-specific globals since we add a few and modified a few existing globals (like table)


mheguy:
Spoke with Val on discord, he liked the idea and since this doesn't change scenario code, doesn't block merging, and can easily be adjusted I'll push this through.

PR #443, opened 2018-11-23
Added a new command wrapper to reduce boilerplate

### LINKED ISSUE #56: add /savecommand /sc
local func = loadstring(str) if func then local status,error = pcall(func) if not status then game.player.print(error) end else game.player.print("Parse Error.") end

### LINKED ISSUE #358: Command wrapper to reduce the amount of code we need to write to add one
Came out of: #355

I propose the following API to add a command. This will let us reduce the boilerplate and can implement generic solutions for code we often have to duplicate

```lua
function Command.add(name, options, callback) end

Command.add('teleport-to-player', {
    description = 'Teleports you to the player', -- optional, defaults to empty string
    arguments = {
        show_message = {required = false, type = 'boolean'}, -- framework should handle the argument order
        player = {required = true, type = 'string'}, -- both the defaults
    },
    admin_only = true, -- defaults to false
    log_command = true, -- defaults to false unless admin only, then defaults to true
}, function (arguments)
    -- arguments = {show_message = true, player = 'Linaori'}
end)
-- The above example would automatically generate the help message
-- "<player_name> <show_message:optional>Teleports you to the player (Admin only)"
-- "<required args> <optional args:optional> <description> <permissions>"
```

This gives is a bit more power and less duplication in code we have to write when creating a command:
 - Arguments are automatically handled based on (A) order as defined (B) optional arguments as last
 - Automatic protection based on the admin option
 - Automatic logging based on admin usage
 - No more fuzzy argument parsing in each command
 - Consistent command arguments
 - Makes easier to manipulate every command in one way or another (think time based usage, limit to player names, limit to debug only etc)

linaori:
This wrapper should make it slightly easier to write custom commands. It supports:
 - logging when a command failed
 - error out when in debug instead of logging
 - should be supporting the server as actor, but untested
 - log prints if the actor doesn't exist (thus server)
 - automatic logging for admin only commands
 - optional logging for player commands
 - named arguments from commands and validation for missing arguments

```lua

local command_options = {
    description = 'this is a test command',
    arguments = {'player_index', 'x', 'y'},
    admin_only = true,
}

Command.add('test-command', command_options, function (arguments, player, tick)
    game.print(string.format('[t:%d][p:%s]]', tick, player.name))
    game.print(serpent.line(arguments))
end)
```
Edit by plague006:
Closes #358, Closes #56

### DIFF
--- a/utils/command.lua
+++ b/utils/command.lua
@@ -1,4 +1,7 @@
+require 'utils.list_utils'
+
 local insert = table.insert
+local size = table.size
 local format = string.format
 
 local Command = {}
@@ -14,6 +17,7 @@ local Command = {}
 ---    allowed_by_server = false -- lets the server execute this, defaults to false
 ---    allowed_by_player = true -- lets players execute this, defaults to true
 ---    log_command = true, -- defaults to false unless admin only, then always true
+---    capture_excess_arguments = true, defaults to false, captures excess arguments in the last argument, useful for sentences
 ---}
 ---
 ---The callback receives the following arguments:
@@ -30,9 +34,11 @@ function Command.add(command_name, options, callback)
     local default_values = options.default_values or {}
     local admin_only = options.admin_only or false
     local debug_only = options.debug_only or false
+    local capture_excess_arguments = options.capture_excess_arguments or false
     local allowed_by_server = options.allowed_by_server or false
     local allowed_by_player = options.allowed_by_player
     local log_command = options.log_command or options.admin_only or false
+    local argument_list_size = size(arguments)
     local argument_list = ''
 
     if nil == options.allowed_by_player then
@@ -47,13 +53,17 @@ function Command.add(command_name, options, callback)
         error(format("The command '%s' is not allowed by the server nor player, please enable at least one of them.", command_name))
     end
 
-    for _, argument_name in ipairs(arguments) do
+    for index, argument_name in ipairs(arguments) do
         local argument_display = argument_name
         for default_value_name, _ in pairs(default_values) do
             if default_value_name == argument_name then
                 argument_display = argument_display .. ':optional'
                 break
             end
+        end
+
+        if argument_list_size == index and capture_excess_arguments then
+            argument_display = argument_display .. ':sentence'
         end
 
         argument_list = format('%s<%s> ', argument_list, argument_display)
@@ -96,8 +106,18 @@ function Command.add(command_name, options, callback)
 
         local named_arguments = {}
         local from_command = {}
+        local raw_parameter_index = 1
         for param in string.gmatch(command.parameter or '', '%S+') do
-            insert(from_command, param)
+            if capture_excess_arguments and raw_parameter_index == argument_list_size then
+                if not from_command[raw_parameter_index] then
+                    from_command[raw_parameter_index] = param
+                else
+                    from_command[raw_parameter_index] = from_command[raw_parameter_index] .. ' ' .. param
+                end
+            else
+                from_command[raw_parameter_index] = param
+                raw_parameter_index = raw_parameter_index + 1
+            end
         end
 
         local errors = {}
--- a/utils/command.lua
+++ b/utils/command.lua
@@ -1,4 +1,4 @@
-require 'utils.list_utils'
+require 'utils.table'
 
 local insert = table.insert
 local size = table.size
--- a/utils/command.lua
+++ b/utils/command.lua
@@ -85,7 +85,7 @@ function Command.add(command_name, options, callback)
         error(format("The command '%s' is not allowed by the server nor player, please enable at least one of them.", command_name))
     end
 
-    for index, argument_name in ipairs(arguments) do
+    for index, argument_name in pairs(arguments) do
         local argument_display = argument_name
         for default_value_name, _ in pairs(default_values) do
             if default_value_name == argument_name then
@@ -152,7 +152,7 @@ function Command.add(command_name, options, callback)
 
         local errors = {}
 
-        for index, argument in ipairs(arguments) do
+        for index, argument in pairs(arguments) do
             local parameter = from_command[index]
 
             if not parameter then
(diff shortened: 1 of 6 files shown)


utils/command.lua
```diff
@@ -0,0 +1,90 @@
...
+    local argument_list = ''
+
+    for _, argument in ipairs(arguments) do
+        argument_list = string.format('%s<%s> ', argument_list, argument)
+    end
+
+    commands.add_command(command_name, argument_list .. description .. (admin_only and ' (Admin Only)' or ''), function (command)
+        local print
+        local player_index = command.player_index
+        local player = game.players[player_index]
```
grilledham: If the server runs a command `command.player_index` is nil. Indexing the `game.players` with a nil key throws an error.
Also you should use `Game.get_player_by_index(player_index)` to avoid the Factorio bug where `game.players` returns the wrong player when used with a player index.
`Game.get_player_by_index(player_index)` also requires you to not pass in a nil value.

utils/command.lua
```diff
@@ -0,0 +1,90 @@
...
+                '[%s Command] %s, used: %s %s',
+                admin_only and 'Admin' or 'Player',
+                player and player.valid and player.name or '<server>',
+                command_name,
+                serpent.line(named_arguments)
+            ))
+        end
+
+        if _DEBUG then
+            -- in debug mode it will crash and report errors directly
```
grilledham: When we merge this, we must remember to remove the code from custom_commands.lua that does the same thing, assuming all custom commands will be re-written to use this module.
Code from  custom_commands.lua.
```lua
if not _DEBUG then
    local old_add_command = commands.add_command
    commands.add_command =
        function(name, desc, func)
        old_add_command(
            name,
            desc,
            function(cmd)
                local success, error = pcall(func, cmd)
                if not success then
                    log(error)
                    Game.player_print('Sorry there was an error running ' .. cmd.name)
                end
            end
        )
    end
end
```

utils/command.lua
Valansch: I suggest instead using game.player. As that does the same as Game.get_player_by_index(player_index).

utils/command.lua
linaori: In theory this shouldn't cause problems (I'll test to be sure), as it just gets wrapped twice, with my error handler taking precedence

linaori: Tested and it works

utils/command.lua
linaori: Using `game.player` now

linaori:
Seems like my admin only check is incomplete

linaori:
Okay, should be ready now

mheguy:
These are the major test cases I could come up with:

```
1 fixed-length, 1 variable-length
report, popup, tag

none
zoom, tppos, meltdown

1 param
invoke, jail, follow

2 param (1 optional)
walkabout swap-players

2 param
tempban

debug-only
all-tech, cloud, clean

exceptional cases
regular: can take 1 param or 2 depending on param 1
kill: an exceptional case that should work with the wrapper as is -- it can be marked as not admin-only in the wrapper and check for admin status inside the function
```

linaori:
What about a default value for arguments if they are not defined?

mheguy:
Being able to set that would fix the case of `regular`. I think it sounds good.

mheguy:
Could also use a `debug-only` switch

linaori:
`Command.add_debug()`?

Perhaps we should go for a `Debug.add_command` instead, and borrow some of Diggy's functionality to base it on?

mheguy:
I'm not picky about implementation

mheguy:
A `command_options` entry of `server_allowed` or `player_required` or something would be nice to filter out the server trying to run commands that are only for players

linaori:
Hmm that would indeed be a nice option. I'll pass them as flags

linaori:
Added the debug/player/server flags. Only thing left to add from what I can tell, is the option to grab all remaining arguments and shove them into the last parameter

linaori:
Excess arguments can now be captured into the last argument:
image

linaori:
Did a lot of testing with @plague006 and based on that I made some additional changes. Should be good for review. If you can think of more edge-cases with existing commands, let me know!

mheguy:
Because this goes best combined with an overhaul of existing commands, going to push this to 1.2.0. Please do not merge until 1.1.0 is complete.

mheguy:
Happy to merge this when the conflict is resolved

utils/command.lua
```diff
@@ -0,0 +1,209 @@
...
+
+    commands.add_command(command_name, argument_list .. description .. extra, function (command)
+        local print -- custom print reference in case no player is present
+        local player_index = command.player_index
+        local player = game.player
+        local player_name = player and player.valid and player.name or '<server>'
+        if not player or not player.valid then
+            print = function (message)
+                log(format('Trying to print message to player #%d, but not such player found: %s', player_index, message))
+            end
```
grilledham: If game.player is nil we should set print to `_G.print` that way if you run the command at the web console you will see the errors in the web console.

utils/command.lua
```diff
@@ -0,0 +1,209 @@
...
+
+        local success, error = pcall(function ()
+            callback(named_arguments, player, command.tick)
+        end)
+
+        if not success then
+            local serialized_arguments = serialize(named_arguments)
+            if _DEBUG then
+                game.print(format("%s triggered an error running a command and has been logged: '%s' with arguments %s", player_name, command_name, serialized_arguments))
+                game.print(error)
```
grilledham: `game.print` doesn't show up in the web console, if player is nil we should consider also outputting to `_G.print`

utils/command.lua
linaori: Should I do both `log` and `print` in that case?

grilledham: As far as I'm aware if `game.player` is nil that can only mean the command came from the server console. So I don't think logging is necessary.

linaori:
@grilledham can you check f88b054 and see if this is correct?

linaori:
Looks like everything is fixed now, ready for review & merge if nothing comes up

grilledham (approved):

mheguy:

PR #451, opened 2018-11-24
Case sensitivity causing Task.lua not to be found

SimonFlapse:


linaori (approved):

mheguy:
We should probably go the other way with this and change `Task`, `Queue`, and `PriorityQueue` to conform with the other utilities' naming conventions. But that's beyond the scope of this fix and I'll open an issue for that.

linaori:
I think we should keep the convention of keeping files lowercase, to avoid this

PR #457, opened 2018-11-25
Broaden table.get_random to work on unordered tables

mheguy:


utils/list_utils.lua
```diff
@@ -95,7 +95,8 @@ end
 
 --- Chooses a random entry from a table
 table.get_random = function(this_table)
-    return this_table[math.random(#this_table)]
+    local size = table.size(this_table)
+    return this_table[math.random(size)]
```
linaori, line 99: I fear I have more lines of feedback than the size of your PR 😅 

I double checked this on the Factorio discord, as this introduces more problems with different table types.

For sequentially indexed tables, the existing solution works fine. However, for indexed arrays this won't work:
 - table.size performance is _a lot_ slower. This can be fixed however. Factorio provides a `table_size` (undocumented), which does exactly what `table.size` does, except it's a lot faster because it's C++ (so we can delegate our table.size or remove it)
 - math.random will give you a random value, but if the table is not sequentially indexed, it will refer to keys that don't exist

The solution would be to a lot more complex than this sadly. First you have to check whether or not it's a sequentially indexed table: `table_size(this_table) == #this_table`. If that works, you can safely do `return this_table[random(#this_table)]`. If it's not the case, you have to build a new table that's sequentially indexed containing the index of the other table as value. That way you can pick a random sequential index that way and grab it from original table.

mheguy: I thought about some of what you mentioned yesterday, but was worried I was overthinking it. Indeed I wasn't. As it stands this PR is all negative and no positive since it slows down processing of sequential tables and won't actually return a valid result in non-sequential tables.

Will commit a more thorough solution and ask for feedback on that.

mheguy:
Superceded by #464

PR #458, opened 2018-11-25
Catch if player is not returned

mheguy:


mheguy:
I gave it a rethink: Realistically if this comes back without a player it means data is or there is a fault earlier in the data chain (not recording the correct index or a player getting lost from game.players) and we probably want an error to pop up.

PR #461, opened 2018-11-25
Added a Market Retailer to manage the market contents of market groups

linaori:
The Retailer supports setting items to a group of markets, without having any markets registered. You can place multiple markets under the same group or put each market in their own group (different forces for example). To update the market contents, call `ship_items` and all markets will have all contents removed and re-added. Each item can only exist in the market once, and prices are always in coins. In the future I'd like to add a command to manually add market items with a price. 

I've implemented this in Diggy as it solves several issues. The Experience feature now has full responsibility to add market items. The first market is added to the StartingZone feature instead, removing the MarketExchange feature completely.


mheguy:
If you want to avoid the wrath of the stickler in your PRs: https://github.com/mpeterv/luacheck

linaori:
@plague006 I want want Herr Stickler's approval 🤔

mheguy:
For sure. I just mean you can locally run the same linting that stickler does if you don't want him to shame you. :p

map_gen/Diggy/Config.lua
```diff
@@ -43,7 +46,7 @@ local Config = {
                 character_health_bonus = 1000000,
 
                 -- unlock all research by default, only useful when testing
-                unlock_all_research = false,
+                unlock_all_research = true,
```
SimonFlapse, line 49: Did we fix the problem with unlocking all research before the market was spawned?

SimonFlapse (approved):

map_gen/Diggy/Config.lua
linaori: Yes, that's kind of the primary reason of this PR 😄

PR #464, opened 2018-11-26
Overhaul utils and add minor functionality

mheguy:
daf3614: branched from #463
ea6c779 is simply renaming and find+replace for requires.
4fbadb6: renames `utils.utils` to `utils.core`, renames `list_utils` to `table`
**core**
`print_except`: mostly just swap where we check for online
`format_time`: push definitions of time into the function
**table**
`verify_mult_types`: new function, would love a better name
`assert_argument_valid`: removed to decrease performance impact
`get_random`: overhauled to take sorted and unsorted tables
`size` removed since factorio has native/undocumented table_size function (in c++)
`get` removed

c6d3510: use `get_random` and `verify_mult_types` in existing code

No testing done at present.


linaori (approved):

Valansch:
Do you still want to test this?

mheguy:
Yes I want to test this before we merge. Like I haven't even checked if the scenario launches.

mheguy:
If this merges after #469, need to change `server.lua` Line 1 to require `utils.token`

plague006:
I don't know if that will actually force windows computers to update the filenames. My feeling is that it likely won't.

mheguy:
Renaming `Queue` to `queue` and `Task` to `task` will be addressed in a different PR after the release of v1.1.0.

utils/table.lua
```diff
@@ -32,21 +23,11 @@ table.add_all = function(t1, t2)
...
-    end
-    return size
-end
-
 --- Checks if a table contains an element
 -- @param t table to search
 -- @param e element to search for
 -- @returns the index of an element or -1
 table.index_of = function(t, e)
-    assert_argument_valid(t)
```
linaori, line 49: Shouldn't the asserts be replaced?

utils/table.lua
```diff
@@ -80,22 +62,22 @@ table.set = function(t, index, element)
...
-        if i == index then
-            return t[k]
+--- Chooses a random entry from a table
+--@param t table to select an element from
+--@param sorted boolean to indicate whether the table is sorted by numerical index or not
+--@return a random element of table t
+table.get_random = function(t, sorted)
+    if sorted then
+        return t[math.random(#t)]
+    else
```
linaori: The else can be omitted here, as the if already returns

mheguy: Quite right, fixed.

utils/table.lua
mheguy: Be replaced as in be used?
I asked Val if asserting was computationally costly and he indicated they were. Because we suppress errors outside of debug anyways, it doesn't seem beneficial to spend computation time asserting arguments.

linaori: Fair enough! It's better to have this kind of validation only when it's not relying on performance.

linaori (approved):

mheguy:
Should be good to go.

mheguy:
Rebased and caught up again

mheguy:
Probably best to push this to 1.2 and just fix the last couple bugs before 1.1

mheguy:
This has been tested a bunch and I want to merge it before the next batch of PRs either conflict or break this by not using the new utils names. :)

PR #465, opened 2018-11-26
Break chat_trigger functions apart

### LINKED ISSUE #306: Is RedMew really a Christian server?
>Religious preferences in games are always a good way to fuel a flame war.

_Originally posted by @iltar in https://github.com/Valansch/RedMew/pull/293_

----

I don't know what the Christian server / naughty words list was ever about. Should we remove this entirely or change it to be less offensive?

mheguy:
Requires #466 at the same time. Closes #306
Would appreciate your eyes @SimonFlapse since this touches `mentions`.

Untested


features/chat_triggers.lua
```diff
@@ -7,78 +7,44 @@ local Hodor = require 'resources.hodor_messages'
...
-    end
-
+--- Check for player and get player
+local function get_player(event)
     -- player_index is nil if the message came from the server,
     -- and indexing Game.players with nil is apparently an error.
     local player_index = event.player_index
     if not player_index then
-        return
+        return false
```
linaori: Imo it's better to return nil instead of false. In other languages you have the concept of `nullable`, which means that you _might_ get a return value if it exists, or null otherwise. If you _always_ expect something, an error should be thrown if it's not found.

linaori (approved):

SimonFlapse:
Seems good to me. I have to change the admin and moderator trigger anyways so I can fix it then

mheguy:
> I can fix it then

If something is broken, I don't want this merged. If you want, feel free to commit to this branch/PR.

features/chat_triggers.lua
```diff
@@ -87,85 +52,93 @@ local function hodor(event)
...
+        end
+        if not success then
+            for _, p in ipairs(game.connected_players) do
+                local word_front_trim = string.sub(word, 2, string.len(word))
+                local word_back_trim = trimmed_word
+                local word_front_back_trim = string.sub(word_front_trim, 0, string.len(word_front_trim) - 1)
+                local word_back_double_trim = string.sub(word_back_trim, 0, string.len(word_back_trim) - 1)
+                word = (trimmed_word == 'admin' or trimmed_word == 'moderator') and trimmed_word or word
+                if admin_call and p.admin then
+                    p.print(prefix .. Game.get_player_by_index(event.player_index).name .. ' mentioned ' .. word .. '!', {r = 1, g = 1, b = 0, a = 1})
```
Valansch: This can be:

```lua
local message = string.format('%s%s mentioned %s!', prefix, Game.get_player_by_index(event.player_index).name, word )
p.print(message, {r = 1, g = 1, b = 0, a = 1})
```

Valansch (approved):
Looking good overall.
I suggest using string.format for string concatenation with variables.

features/chat_triggers.lua
mheguy: Done, anything else to merge this? I'm pretty sure https://github.com/Refactorio/RedMew/pull/466 puts the scenario into a state that will not load until this is merged.

SimonFlapse (approved):

mheguy:
Testing went well

PR #467, opened 2018-11-26
Fixed antigrief desyncs

Valansch:


map_gen/Diggy/Feature/Antigrief.lua
```diff
@@ -12,44 +12,45 @@ local Report = require 'features.report'
...
         local count = player_collapses[player_index] or 0
         count = count + 1
         player_collapses[player_index] = count
         local player = Game.get_player_by_index(player_index)
-        if global.Antigrief.autojail and count > allowed_collapses_first_hour and player.online_time < 216000 and not global.Antigrief.jailed_players[player_index] then
+        if global_primitives.autojail and count > allowed_collapses_first_hour and player.online_time < 216000 and not jailed_players[player_index] then
             Report.jail(player)
             Report.report(nil, player, string.format("Caused %d collapses in the first hour", count))
-            global.Antigrief.jailed_players[player_index] = true
+            jailed_players[player_index] = true
```
linaori, line 55: This is never removed from the table, would it make sense for that to happen?

Valansch: If a person is jailed once they will never be jailed again. This is intended that way yes. If an admin already dealt with it and it happends again then it will most likely be an accident.

linaori (approved):

mheguy (changes requested):
In testing:

```
"Error while running event level::on_entity_died (ID 4)
LuaEntity doesn't contain key index.
stack traceback:
...ntly-playing/map_gen/Diggy/Feature/DiggyCaveCollapse.lua:249: in function 'handler'
/factorio/6/temp/currently-playing/utils/event.lua:16: in function 'call_handlers'
/factorio/6/temp/currently-playing/utils/event.lua:30: in function </factorio/6/temp/currently-playing/utils/event.lua:28>"
```

In addition, when a cave collapse killed a player there was a corpse and a character standing there unattached to a player. The player then respawned as expected.

linaori:
I think that's the bug I fixed in 470 yesterday while we were testing: https://github.com/Refactorio/RedMew/blob/6b589fb8f6d33a8508b52fcf0d4aeed7782b30ed/map_gen/Diggy/Feature/DiggyCaveCollapse.lua#L248

mheguy (approved):
Both bugs went away after rebasing the branch to dev. Merging with my thanks.

PR #470, opened 2018-11-26
Fixed a bunch of performance issues and bugs in Diggy

linaori:
Contains a bunch of small performance fixes, a small change to the biter spawning algorithm to spread the spawning over different ticks to reduce stuttering and fixed worms not giving XP/coins.


linaori:
Tested the second commit and for 10k iterations on both, count was 13ms vs 20ms on get, so this is a decent improvement I hope.

mheguy (commented):
I think you're in a rush to merge, but I can't give a full review. I'll say nothing jumped out at me while I skimmed this.
I whispered to Herr Stickler and he gave me this list for you :p
https://gist.github.com/plague006/6fe3d8967020909ef0d4b717cf1386ca

linaori:
Not ready for merging yet, will continue tomorrow. There's something iffy with performance on the last change

Valansch (approved):
Looking good

map_gen/Diggy/Scanner.lua
```diff
@@ -13,30 +19,32 @@ local Scanner = {}
...
-    if (tile_search == get_tile(position.x, position.y + 1).name) then
-        insert(tile_found, {x = position.x, y = position.y + 1})
-    end
-
-    -- west
-    if (tile_search == get_tile(position.x - 1, position.y).name) then
-        insert(tile_found, {x = position.x - 1, y = position.y})
+    local tiles_found = {}
+    local next = 0
+    local count_tiles_filtered = surface.count_tiles_filtered
```
linaori: Going to remove the scanner and inline the functionality. Saves 1 extra function call reference, Additionally I will replace `insert()` with a faster variant and put the x/y into locals. 

Calculating and generating the area table seems to cause heavier load than actually saving with the API calls.

map_gen/Diggy/Config.lua
```diff
@@ -416,8 +417,21 @@ local Config = {
...
+                ['small-worm-turret'] = 2,
+                ['medium-biter'] = 3,
+                ['medium-spitter'] = 3,
+                ['medium-worm-turret'] = 3,
+                ['big-biter'] = 5,
+                ['big-spitter'] = 5,
+                ['big-worm-turret'] = 5,
+                ['behemoth-biter'] = 7,
+                ['behemoth-spitter'] = 7,
+            },
```
SimonFlapse, line 433: Since these are the same values af the coin modifiers do you think it is necessary to have them?
https://github.com/Refactorio/RedMew/blob/92703968a13534e52692abd822c58a3c0064f094/map_gen/Diggy/Config.lua#L374 Continuing. Is this line necessary anymore? I like them all being configurable so I don't mind.

SimonFlapse (approved):

map_gen/Diggy/Config.lua
linaori: I removed the comment as it felt out of place

Regarding the table, yes it's needed to duplicate as they otherwise cause cross feature dependencies

mheguy (approved):
Testing went well, LGTM

PR #473, opened 2018-11-27
Fix /kill checks around walkabout

mheguy:


mheguy:
Doesn't stop the server using /kill one someone on walkabout

mheguy:
Should be good to go.

linaori (approved):

PR #476, opened 2018-11-28
Ban sync

grilledham:
Due to the way the web interface synchronises bans, bans that happen in scripts with `game.ban_player` wont be synced.

This PR adds `Server.ban_sync` and `Server.unban_sync` which allows bans that happen from scripts to be synchronised.

I also added `Server.ban_non_sync` and `Server.unban_non_sync` for symmetry but they just call `game.ban_player` and `game.unban_player` respectively.

### DIFF
--- a/features/server.lua
+++ b/features/server.lua
@@ -413,7 +413,7 @@ function Public.unban_sync(username, admin)
         admin = '<script>'
     end
 
-    -- game.ban_player errors if player not found.
+    -- game.unban_player errors if player not found.
     -- However we may still want to use this function to unban player names.
     local player = game.players[username]
     if player then
--- a/features/server.lua
+++ b/features/server.lua
@@ -357,7 +357,7 @@ local function escape(str)
 end
 
 --- If the player exists bans the player.
--- Regardless of if the player exists the name is synchronized with other servers
+-- Regardless of whether or not the player exists the name is synchronized with other servers
 -- and stored in the database.
 -- @param  username<string>
 -- @param  reason<string?> defaults to empty string.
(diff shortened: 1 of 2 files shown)


features/server.lua
```diff
@@ -350,4 +352,86 @@ function Public.get_tracked_data_sets()
...
+-- @param  username<string>
+-- @param  reason<string?> defaults to empty string.
+-- @param  admin<string?> admin's name, defaults to '<script>'
+function Public.ban_sync(username, reason, admin)
+    if type(username) ~= 'string' then
+        error('username must be a string')
+    end
+
+    if type(reason) ~= 'string' then
+        reason = ''
```
mheguy, line 371: If reason is coming in as a non-string, would that indicate something really went wrong? Do we want to let things keep running? Can we do the whole "tostring" thing or whatever and if it's a type that can't become a string then error out?

features/server.lua
```diff
@@ -350,4 +352,86 @@ function Public.get_tracked_data_sets()
...
+    end
+
+    if type(admin) ~= 'string' then
```
mheguy: Same deal here. If admin is nil then setting it to `<script>` makes sense, but should we let it slide if a non-string argument is coming in?

features/server.lua
```diff
@@ -350,4 +352,86 @@ function Public.get_tracked_data_sets()
...
+function Public.unban_sync(username, admin)
+    if type(username) ~= 'string' then
+        error('username must be a string')
+    end
+
+    if type(admin) ~= 'string' then
+        admin = '<script>'
+    end
+
+    -- game.ban_player errors if player not found.
```
mheguy: game.unban_player

features/server.lua
grilledham: I wasn't quite sure about this. I originally had them as nil checks so that it matched the factorio api. But for some reason it felt safer as string checks. 
If we change it back to a nil check it should probably error if not string as that's what game.ban_player does

mheguy: An error on wrong type makes sense to me.

features/server.lua
grilledham: fixed

features/server.lua
grilledham: Changed to throw error on non string input.

mheguy (approved):

PR #480, opened 2018-11-29
New evolution calculation and item balancing

SimonFlapse:
Temporarily replaced by #481

### DIFF
--- a/map_gen/Diggy/Feature/AlienSpawner.lua
+++ b/map_gen/Diggy/Feature/AlienSpawner.lua
@@ -198,7 +198,6 @@ function AlienSpawner.register(config)
     Event.add(Template.events.on_void_removed, function (event)
         local force = game.forces.enemy
         local evolution_factor = force.evolution_factor
-        force.evolution_factor = evolution_factor + 0.0000012
 
         local position = event.position
         local x = position.x
@@ -224,8 +223,8 @@ Minimum spawn distance: ]] .. config.alien_minimum_distance .. ' tiles'
 end
 
 function AlienSpawner.on_init()
-    -- base factorio =                pollution_factor = 0.000015
-    game.map_settings.enemy_evolution.pollution_factor = 0.000004
+    -- base factorio =                time_factor = 0.000004
+    game.map_settings.enemy_evolution.time_factor = 0.000008
 end
 
 return AlienSpawner
--- a/map_gen/Diggy/Feature/Experience.lua
+++ b/map_gen/Diggy/Feature/Experience.lua
@@ -4,6 +4,7 @@ local Game = require 'utils.game'
 local Global = require 'utils.global'
 local ForceControl = require 'features.force_control'
 local Debug = require 'map_gen.Diggy.Debug'
+local Template = require 'map_gen.Diggy.Template'
 local Retailer = require 'features.retailer'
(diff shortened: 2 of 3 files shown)


map_gen/Diggy/Feature/Experience.lua
```diff
@@ -287,6 +289,24 @@ local function on_player_respawned(event)
     end
 end
 
+local function on_void_removed(event)
```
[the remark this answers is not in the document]
SimonFlapse: True

map_gen/Diggy/Feature/AlienSpawner.lua
```diff
@@ -224,8 +223,8 @@ Minimum spawn distance: ]] .. config.alien_minimum_distance .. ' tiles'
 end
 
 function AlienSpawner.on_init()
-    -- base factorio =                pollution_factor = 0.000015
-    game.map_settings.enemy_evolution.pollution_factor = 0.000004
+    -- base factorio =                time_factor = 0.000004
+    game.map_settings.enemy_evolution.time_factor = 0.000008
```
linaori, line 227: Where's pollution disabled?

map_gen/Diggy/Feature/Experience.lua
```diff
@@ -51,6 +52,7 @@ local gain_xp_color = {r = 144, g = 202, b = 249}
 local lose_xp_color = {r = 255, g = 0, b = 0}
 local unlocked_color = {r = 255, g = 255, b = 255}
 local locked_color = {r = 127, g = 127, b = 127}
+local count = -189
```
linaori, line 55: What's this local for? I think it would cause desyncs

map_gen/Diggy/Feature/Experience.lua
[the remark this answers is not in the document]
mheguy: You can use `_` for unused data. That way sticker will leave you alone and people reading it will understand it's intentional and not a mistake.
So in this context: `local function on_void_removed(_)`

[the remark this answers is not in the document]
linaori: In case of unused arguments, you can even omit them. Lua won't complain if you define less parameters than passed
```lua
local function foo() end
foo(1, 2, 3) -- works
```

map_gen/Diggy/Feature/AlienSpawner.lua
SimonFlapse: Oh yeah that's what I forgot. You can disabled it in the mapgen settings before generating the map. Not sure how else to do it

map_gen/Diggy/Feature/Experience.lua
SimonFlapse: Could be, haven't checked with more than one player. It's used to prevent the table lookup further down at every void removed, but only every 50th removed. And since the starting zone removed 189 i'm trying to prevent this from affecting the evolution

SimonFlapse: It is actually redundant now since everything below level 1 just defaults to 0.1 evolution

Valansch (changes requested):
See @iltar s comments

SimonFlapse:
Outdated, closing this PR. I need to reevaluate the balancing needs.

PR #484, opened 2018-11-29
Expanded stat tracking

### LINKED ISSUE #472: Expanding stat tracking
Opening this issue to brainstorm some ideas on what stats we might want to track on players both long (server side across maps) and short term (per map) now have the ability to track stats server-side.

Don't be shy to add or suggest silly things, it can lead to something awesome.

* Playtime
* Coins
* Distance traveled
* Deaths
* Deaths
* Entities built
* Trees chopped
* Total kills by train

mheguy:
Greatly expands out stat tracking **but** doesn't display or make any use of any of the per player stats.

Tracking now includes:

**Per player**
* Items hand crafted
* Time chatted in console
* Damage taken (does not record damage to shields)
* Damage dealt

**Totals**
* Number of kills by train
* Number of trees mined by players
* Number of rocks mined by players
* Number of entities built by robots
* Number of entities built by players

**Stored server-side**
* Total playtime

Closes #472

Would love thoughts on if the extra tracking will create any meaningful computational burden. In particular the damaged_entity event could be called *a lot* on certain maps.


features/player_stats.lua
```diff
@@ -131,13 +244,55 @@ function Public.get_death_count(player_index)
...
--- Returns a dictionary of casue_name -> count
+-- Returns a dictionary of cause_name -> count
 function Public.get_all_death_counts_by_casue(player_index)
     return player_deaths[player_index].causes or {}
 end
 
+-- Returns the amount of time the player spent on redmew before joining this map
+function Public.get_player_on_create_playtime(player_index)
+    return player_on_create_playtime[player_index] or 0
+end
```
grilledham: It would be more consistent with the other properties to initialise this value to 0 in the `player_created` function then remove the or 0

mheguy: I did end up initializing it properly in on L75 but forgot that bit of cleanup. Thanks for the spot.

features/player_stats.lua
```diff
@@ -131,13 +244,55 @@ function Public.get_death_count(player_index)
...
+
+-- Returns the amount of time the player spent on previous maps in addition to the current
+function Public.get_player_total_playtime(player_index)
+    return calculate_player_total_playtime(player_index)
+end
+
 function Public.get_total_player_count()
     return total_players[1]
 end
```
grilledham, line 244: I know I wrote this, but I now know that `#game.players` does the same thing.
So we could replace `get_total_player_count()` with `#game.players` in other modules that use this.

features/player_stats.lua
grilledham: `player_on_create_playtime[player_index] = 0` should still be added to `player_created` as the callback is called at a later tick or never if the web server is down

features/player_stats.lua
mheguy: Or maybe you're actually a prophet

>[2:59 PM] Godmave: but when is a player removed? it is not when he leaves the game (I know there is an event for that). so when is he "removed"?
[2:59 PM] Rseding91: https://lua-api.factorio.com/latest/events.html#on_player_removed
👌
2
[3:02 PM] Godmave: only time I got the removed was with the clusterio player sync which does that
[3:02 PM] Bilka: the game itself doesnt remove players at all afaik(edited)
[3:02 PM] Bilka: inb4 Rseding says I'm wrong
[3:05 PM] Rseding91: As of 0.16 you're correct.
[3:05 PM] Bilka: :D
[3:05 PM] GunnerWolf: as of 0.17 he's wrong? :stuck_out_tongue:
[3:05 PM] Rseding91: To be determined.

grilledham: OK, so `get_total_player_count()` and `#game.players` would return different numbers if we ever used `game.remove_offline_players`. I don't know if that is a good enough reason to keep `get_total_player_count()` thought 🤔

mheguy: They actually weren't referring to `game.remove_offline_players`, rseding is alluding to something in 0.17 that might remove players but didn't specify the what or how:

>3:05 PM] Rseding91: To be determined.
[3:06 PM] Bilka: ohhh I know why! Knowledge is power!
[3:06 PM] Godmave: Ok, will catch it anyway so it can be used with clusterio :p
[3:06 PM] Bilka: Please dont add it though, it would mess up so many mods(edited)
[3:06 PM] Rseding91: Why would it? It's already part of the game like surfaces and forces.
[3:08 PM] Bilka: Because most people don't know how the entire game works, so they never even thought about subscribing to the event. Since the action is currently so rare, they probably also never got bug reports about it, so it is not "normal" to them. By adding a way to easily delete players you are suddenly making it normal, for imo no good reason.
[3:08 PM] Godmave: only place where it fell on my foot was the gui stuff that reacted allergic at a missing player
[3:09 PM] Rseding91: Not adding some feature because mods don't currently handle things correctly is never going to be a good enough excuse for me.
[3:09 PM] Bilka: Right, you. Please add on_pre_player_removed then if you do add that feature.
[3:10 PM] Rseding91: The feature already exists though.
[3:11 PM] Rseding91: https://lua-api.factorio.com/latest/LuaGameScript.html#LuaGameScript.remove_offline_players
[3:11 PM] Bilka: You know what I mean. There is a big difference between a lua function and a gui. Saying that, "research all technologies" in map editor when?
[3:12 PM] Rseding91: When they stop messing with the technology GUI.
[3:12 PM] Bilka: :thumbsup:
3:27 PM] Klonan: I dont see anything wrong with removing players
[3:28 PM] Klonan: all you need is the player index or name and purge all global data you might have about them

I'm cool with changing it either way.

grilledham:
Looking at this again, I'm not sure if this way of tracking player's total playtime would work.
Two cases come to mind:
1. What if a player makes the save in single player than uploads it to the server to be played. That player would have 0 for `player_on_create_playtime`
2. A player leaves server 1 and goes plays on server 2, then after some time comes back to server 1 that is still running the same save.

I think we may need to do something like:
- on_player_joined fetch up to date total playtime.
- record how much online_time the player already has on join, so we can offset that against how much time is added to total player time when the player leaves.

Another option would be to have the web server handle player's total play time.

linaori:
@grilledham what if we do something like: `on_player_joined_game` -> `Server.player_time_start(name)` and `on_player_left_game` -> `Server.player_time_stop(name)`. You can only be active on 1 server at the same time, so this date range could solved server side.

grilledham:
@iltar If the web server calculates the total player time, it doesn't need to be told when players join or leave, it knows already.

linaori:
Then all we have to do is query it from the server and cache it locally, right?

grilledham:
I think that is correct.
If the web server calculates the total player time, all the script has to is fetch it on player_joined and cache it.
Of course the code for the web server would have to be written.

mheguy:
If we want to go server-side that's cool with me. Just need to know which direction you guys want to go and I'll change accordingly.

mheguy:
I wanted to start using some of the stats and while I was at it I took another swing at game-side playtime tracking. I don't know if it covers every edge case, I know that in testing it carried SP time up into MP. But that could also be affected by if a player's name is already in the table or not.

Either way server-side tracking would likely be more robust, I just wanted to do a better job than my first attempt.

mheguy:
With total playtime stripped out can this now be merged?

linaori (approved):

grilledham (approved):

PR #487, opened 2018-11-30
Server time

grilledham:
Added a method to the Server module to get the web server's current time.

Also added a timestamp module to utils for manipulating timestamps into easier to manage forms.

Also added a custom command /server-time for printing the web server's current time


linaori (approved):

mheguy (approved):
Insofar as I was able to test it, I didn't run into any issues.

PR #488, opened 2018-11-30
Fixed a few left-over requires in Experience

linaori:
While it shouldn't cause any performance issues, this is cleaner. I've tested it with heavy-mode to test for desyncs and everything works fine.


mheguy (approved):
Nothing in the code jumps out at me, and local testing didn't reveal any issues.

PR #489, opened 2018-11-30
Added big rocks to the table

linaori:
Because why not 😄

Size wise it's the same as sand-rock-big, but in the same style as rock-huge.

### DIFF
--- a/map_gen/Diggy/Feature/DiggyCaveCollapse.lua
+++ b/map_gen/Diggy/Feature/DiggyCaveCollapse.lua
@@ -87,11 +87,12 @@ DiggyCaveCollapse.events = {
     ]]
     on_collapse = script.generate_event_name()
 }
+local collapse_rocks = {'sand-rock-big', 'rock-big', 'rock-big'}
+local collapse_rocks_size = #collapse_rocks
 
 local function create_collapse_template(positions, surface)
     local entities = {}
     local entity_count = 0
-
     local find_entities_filtered = surface.find_entities_filtered
 
     for _, position in pairs(positions) do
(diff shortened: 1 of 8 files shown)


map_gen/Diggy/Feature/DiggyCaveCollapse.lua
```diff
@@ -112,15 +113,15 @@ local function create_collapse_template(positions, surface)
 
         if do_insert then
             entity_count = entity_count + 1
-            entities[entity_count] = {position = {x = x, y = y}, name = 'sand-rock-big'}
+            entities[entity_count] = {position = position, name = collapse_rocks[random(1, collapse_rocks_size)]}
```
mheguy: If you wanna be 😎 cool you could just use `random(collapse_rock_size)`. The lower limit of 1 is implies.

map_gen/Diggy/Feature/DiggyHole.lua
```diff
@@ -127,7 +130,7 @@ function DiggyHole.register(config)
     Event.add(defines.events.on_entity_died, function (event)
         local entity = event.entity
         local name = entity.name
-        if name ~= 'sand-rock-big' and name ~= 'rock-huge' then
+        if name ~= 'sand-rock-big' and name ~= 'rock-huge' and name ~= 'rock-big' then
```
mheguy: This exact rock check and return pops up at least 4 times in this file, would like to see it split into a function.

mheguy (approved):
Ran into no issues during testing and no issues popped out in looking at the code. Would like to see the function created for that rock check, but it's not something that blocks this merge IMO.

map_gen/Diggy/Feature/DiggyCaveCollapse.lua
linaori: I knew that! 😅

map_gen/Diggy/Feature/DiggyHole.lua
linaori: I think that's indeed a nice thing to have about now. I'll fix this once I get back home later today

linaori:
Fixed the random and moved the definition of the diggy rocks table + check to the Template lua file

mheguy:

PR #490, opened 2018-11-30
Add core cheats

mheguy:
Mainly just for testing. Adds armor, ammo, axes, bots, etc. for all-purpose testing, adds 'unique' items like small-plane and infinity chests so that they don't have to be inserted by command, turns on cheat mode, gives mining speed and run speed modifiers and gives a big health bonus. Health bonus intentionally only applied on_create because there are times we want regular health so a quick `/kill` gets us there.


features/player_create.lua
```diff
@@ -27,6 +27,47 @@ local function player_created(event)
...
         info.show_info({player = player})
     end
+
+    if _CHEATS then
+        local force = player.force
+        local i = player.insert
+
+        i {name = 'power-armor-mk2', count = 1}
+        local p_armor = player.get_inventory(5)[1].grid
+        p_armor.put({name = 'fusion-reactor-equipment'})
```
linaori: You can reference this function call as well

linaori (approved):

Valansch:
Finally someone took the time to actually implement this. I think we all wanted this :D

PR #493, opened 2018-12-01
Add camera

mheguy:
Adaptation of https://mods.factorio.com/mod/camera used under MIT licensing.

`/camera` opens a window with nice zoom and size controls, allows you to choose a target and get a camera that follows them as well as your latency allows.

Thanks to @iltar's great command wrapper we can quickly and easily toggle it to admin-only if over-use causes any issues.

### DIFF
--- a/features/gui/camera.lua
+++ b/features/gui/camera.lua
@@ -184,5 +184,5 @@ local function watch_close_button(event)
 end
 
 Command.add('watch', {description = 'Allows you to watch other players. Use /watch to close the camera.', arguments = {'target'}, default_values = {target = false}, admin_only = false}, watch_command)
-Event.add(defines.events.on_tick, on_tick)
+Event.on_nth_tick(120, on_tick)
 Gui.on_click(main_button_name, watch_close_button)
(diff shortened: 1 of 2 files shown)


mheguy:
This isn't in a stable state yet.

```
482.117 Script @/factorio/6/temp/currently-playing/utils/event.lua:22: /factorio/6/temp/currently-playing/features/gui/camera.lua:66: attempt to index local 'target' (a nil value)
```

mheguy:
Still rewriting, but stumbled upon some unexpected behaviour from the command wrapper so pushed a commit to refer to.

@iltar it's possible the error is on my part, but I can't use the command without an argument, despite the argument being optional. It works as expected when given an argument,
image

linaori:
Your code looks good to me, it shouldn't be complaining. Might have introduced a bug at some point after the merge, will look into it later

linaori:
Actually, the problem might be this: https://github.com/Refactorio/RedMew/blob/3883c9e7705644f7c5ab4494f88f29fcb46db4c2/utils/command.lua#L166-L168

It's doing a loose type check rather than checking for nil, while the default value is false. If you change that to `parameter ~= nil`, does it work?

mheguy:
Good call, changing it to `parameter == nil`  makes it act as expected. =)

mheguy:
I would love to use `utils.gui` for the whole thing, but I wasn't able to bumble my way into a solution that worked. Instead it's a `mod-gui`/`utils.gui` hybrid.

Would benefit from MP testing, but SP works fine.

features/gui/camera.lua
```diff
@@ -0,0 +1,171 @@
...
+        sizeselection = targetframe.add {type = 'drop-down', name = 'sizeselection', items = sizelabels, selected_index = 4}
+    end
+
+    local size = sizelevels[sizeselection.selected_index]
+    local visible = (size ~= 0)
+    return size, visible
+end
+
+local function on_tick()
+    for table_key, camera_table in pairs(global.camera_users) do
```
linaori: As this is executed _every_ tick, I fear this might become slow. Should it be reduced to every nth tick? I'm sure we can already extract the `global` here to a local with `Global.register` though

linaori (commented):
I'm not 100% sure at the moment if this will be impacting performance or cause very little overhead

features/gui/camera.lua
mheguy: Global register in 64df783

I had tried every 2nd tick and found the latency motion-sickness inducing. But your comment pushed me to try every 60th tick and that produces a very NES-feeling camera since the time it takes a player to run to the edge of the camera viewport = the time between viewport updates.

mheguy:
There's a runtime killswitch in case things go bad. That switch does create overhead but I think it's worthwhile in case things go awry. And with a 2 second delay the camera is still just as valuable since the motion of the player is tracked real-time, it's just the frame itself that needs to be redrawn.

PR #494, opened 2018-12-01
Added 'rotten_apples' map preset and supporting files.

Jayefuu:


map_gen/presets/rotten_apples.lua
```diff
@@ -0,0 +1,402 @@
...
+        local d_sq = x * x + y * y
+        return base + mult * d_sq ^ (pow / 2) -- d ^ pow
+    end
+end
+
+-- removes spawners when called
+local function no_spawners(x, y, world, tile)
+    for _, e in ipairs(
+        world.surface.find_entities_filtered(
+            {force = 'enemy', name = names, position = {world.x, world.y}}
```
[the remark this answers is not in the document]
mheguy: This one is an actual problem, feel free to ignore the others.

mheguy (changes requested):
`no_spawners` seems to be the same as tetris, it's just missing the table `names` with the 2 types of spawners listed.

map_gen/presets/rotten_apples.lua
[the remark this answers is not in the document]
Jayefuu: Updated, thank you.

map_gen/presets/rotten_apples.lua
```diff
@@ -0,0 +1,408 @@
...
+    ['laser-turret'] = 0.12,
+}
+
+local function modify_damage(force, mult)
+    for type, mod in pairs(player_ammo_research_modifiers) do
+        local current_m = force.get_ammo_damage_modifier(type)
+        if (current_m + (mod * mult))<=-0.9 then
+            force.set_ammo_damage_modifier(type, -0.9)
+        else
+            force.set_ammo_damage_modifier(type, current_m + (mod * mult))
```
linaori, line 176: considering `force.get_ammo_damage_modifier` and `force.set_ammo_damage_modifier` are called in a loop, I recommend to extract the references to those functions to just above the loop, meaning it won't have to talk to the factorio API to get these functions

map_gen/presets/rotten_apples.lua
```diff
@@ -0,0 +1,408 @@
...
+
+local names = {
+    'biter-spawner',
+    'spitter-spawner'
+}
+
+-- removes spawners when called
+local function no_spawners(_, _, world, tile)
+    for _, e in ipairs(
+        world.surface.find_entities_filtered(
```
linaori, line 217: ~~`local find_entities_filtered = world.surface.find_entities_filtered` would reduce overhead here.~~ Additionally you could use pairs instead of ipairs, as it's faster. If you want to go for speed, use a for loop with a count

There are some more places where you can do this

Edit: nvm, didn't see that it was _in_ the ipairs

Jayefuu: It doesn't run very often and I'm not sure I'll ever reuse the code. It wasn't a particularly fun mechanic, not with the balance we had anyway.

map_gen/presets/rotten_apples.lua
Jayefuu: If you have time, please could you make this change so I've got an example of the correct way to do it the next time?

linaori: Example:
```lua
local function modify_damage(force, mult)
    local get_ammo_damage_modifier = force.get_ammo_damage_modifier
    local set_ammo_damage_modifier = force.set_ammo_damage_modifier
    for type, mod in pairs(player_ammo_research_modifiers) do
        local current_m = get_ammo_damage_modifier(type)
        if (current_m + (mod * mult))<=-0.9 then
            set_ammo_damage_modifier(type, -0.9)
        else
            set_ammo_damage_modifier(type, current_m + (mod * mult))
```
Every time `force.<something>` is called, it has to make an API call to the object in Factorio (C++), meaning you have some overhead each time you call it. This way the reference to the function is stored locally and you only call it when initially fetching it. Afterwards you're calling that local variant, reducing overhead.

mheguy (approved):

PR #496, opened 2018-12-01
sc command

### LINKED ISSUE #56: add /savecommand /sc

grilledham:
closes #56

On server 6 as `quite_command_test-deflated.zip` if anyone wants to test it

### DIFF
--- a/features/custom_commands.lua
+++ b/features/custom_commands.lua
@@ -7,6 +7,7 @@ local Game = require 'utils.game'
 local Report = require 'features.report'
 local Server = require 'features.server'
 local Timestamp = require 'utils.timestamp'
+local Command = require 'utils.command'
 
 --local Antigrief = require 'features.antigrief'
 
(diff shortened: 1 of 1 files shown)


mheguy (commented):
Works as advertised and code looks good. But can I coax you into using the new command wrapper?

grilledham:
Currently I don't think the Command warpper supports what I need for this command.

Here is my current attempt, I've put comments in for what I want.
```lua
Command.add(
    'sc',
    {
        description = 'silent-command',
        arguments = {'str'},
        admin_only = true,
        capture_excess_arguments = true,
        allowed_by_server = true,
        allowed_by_player = true
    },
    function(args)
        local func, err = loadstring(args.str)
        if not func then
            error(err) -- I want to print err to the player or server
        end

        local _, err2 = pcall(func)
        if err2 then
            local i = err2:find('\n')
            error(err2:sub(1, i))  -- I want to print err to the player or server
        end
    end
)
```
Maybe for the command wrapper if player is nil we could mock it with 
```lua
player = {valid = true, admin = true, index = 0, print = _G.print}
```
@iltar 

edit:
Thinking about it again I think I can do what I want in the command.
```lua
Command.add(
    'sc',
    {
        description = 'silent-command',
        arguments = {'str'},
        admin_only = true,
        capture_excess_arguments = true,
        allowed_by_server = true,
        allowed_by_player = true
    },
    function(args, player)
        local p
        if player then
            p = player.print
        else
            p = print
        end

        local func, err = loadstring(args.str)
        if not func then
            p(err)
        end

        local _, err2 = pcall(func)
        if err2 then
            local i = err2:find('\n')
            p(err2:sub(1, i))
        end
    end
)
```

linaori:
@grilledham your scenario is something I've considered building, but as Lua doesn't have actual types, mocking a LuaPlayer with a normal table can break a lot as people will assume it's a player object.

The edit you made should indeed work

Valansch (approved):

mheguy:
Code looks good, works well. Merging with my thanks.

PR #498, opened 2018-12-01
Improved Biter aggression and scales a bit higher

linaori:
- Killing aliens will now spawn up to 60 blood particles at the alien that died, based on how many other aliens it spawns
 - Fixed some (potential) bugs in the Experience feature
 - Tweaked initial alien spawning values from hydra
 - Hydra spawns now add the evolution as raw chance to each additional spawn, meaning the higher evo, the more hydra spawns
 - Hydra spawns now try to go for the target that killed whatever caused them to spawn

### DIFF
--- a/map_gen/Diggy/Feature/AlienSpawner.lua
+++ b/map_gen/Diggy/Feature/AlienSpawner.lua
@@ -3,6 +3,7 @@
 ]]
 
 -- dependencies
+require 'utils.table'
 local Event = require 'utils.event'
 local Global = require 'utils.global'
 local Token = require 'utils.token'
@@ -14,7 +15,7 @@ local CreateParticles = require 'features.create_particles'
 local random = math.random
 local floor = math.floor
 local ceil = math.ceil
-local insert = table.insert
+local size = table.size
 local raise_event = script.raise_event
 
 -- this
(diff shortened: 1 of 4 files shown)


map_gen/Diggy/Feature/AlienSpawner.lua
```diff
@@ -3,6 +3,7 @@
 ]]
 
 -- dependencies
+require 'utils.table'
```
mheguy, line 6: Not in relation to this PR but as a general side note/thought: should we just require utils.table in control since it's extending the core lua table?

mheguy:
<3 that bot

mheguy (approved):
Nothing in the code jumps out at me, and 5 mins of testing didn't reveal any issues.
Merging with my thanks.

linaori:
Yes, Herr Stickler is a great addition to the team 😄

map_gen/Diggy/Feature/AlienSpawner.lua
linaori: I think we already do, but I like to explicitly define those things. Should we create a special namespace/directory for these kind of things so we can tell just by looking at it "oh this is required by default, it's the core, I don't need to do this anymore"

mheguy: Something like that would be nice. But as I think about it more, it's probably best to require it explicitly since even though we're extending something like table, people shouldn't have the expectation that non-standard functionality is present. From a more selfish perspective, I don't think my intellisense would pick up on it. :D

PR #499, opened 2018-12-01
Minor fixes in the config

linaori:
Structure is still the same, but defined differently. Also added some references to `table.insert` in poll and fixed a typo.


mheguy (approved):

PR #502, opened 2018-12-01
Added a basic RedMew configuration setup

linaori:
For existing config I've made sure their structure is the same. For other configurable stuff I've made sure the config is respected and prevents loading stuff if it's not used. I haven't added config to everything, but the most obvious stuff is configured. This prevents people from having to go to control.lua and makes it easier to configure what's loaded from the context of scenarios (think Diggy). This will reduce eagerly loaded event listeners for example. Additionally we can now document all the features at a single point; the config.

### DIFF
--- a/.luacheckrc
+++ b/.luacheckrc
@@ -56,7 +56,7 @@ local STD_BASE_CONTROL = 'lua52c+factorio+factorio_control+factorio_defines+fact
 --[Assume Factorio Control stage as default]--
 -------------------------------------------------------------------------------
 std = STD_CONTROL
-globals = {'print', 'math', 'table', '_DEBUG', '_CHEATS', 'MARKET_ITEM', 'ServerCommands'} -- RedMew-specific globals
+globals = {'print', 'math', 'table', '_DEBUG', '_CHEATS', 'ServerCommands'} -- RedMew-specific globals
 max_line_length = LINE_LENGTH
 
 not_globals = NOT_GLOBALS
--- a/config.lua
+++ b/config.lua
@@ -1,6 +1,5 @@
 _DEBUG = false
 _CHEATS = false
-MARKET_ITEM = 'coin'
 
 global.config = {
     -- The title of the map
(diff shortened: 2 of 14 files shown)


grilledham:
This is a good idea and something we needed for some time now.

I think it would make sense to split config into two parts:
1. configs that can be changed at runtime.
2. configs that have to be set before runtime. (These shouldn't be inside the global table)

This will make it obvious what we can safety change at runtime with say a command.

linaori:
I tried implementing something like that for the meltdown. You can either enable or disable the feature and then sort of "lazy" load it if you wish. I'm not sure if the structure for that is the best the way I defined it.

linaori:
@grilledham anything you want to see changed for this PR or should we merge this as is and improve on it?

mheguy:
Splitting it into two parts like top and bottom of the file? Or do you mean in terms of table structure like `global.config.runtime` and `global.config.not_runtime` (don't know what word to use for pre-init)?

Valansch:
@plague006 How about "finals", "final", "readonlys" or "readonly"?
But this should be in the global table.

linaori:
What if I'd take over the structure of Diggy for this?
```lua
return {
    -- this feature can only be enabled or disabled completely
    autodeconstruct = {
        enabled = true,
        runtime = false, -- default, means you can't toggle the state 
    },
    nuke_control = {
        enabled = true,
        runtime = true, -- meaning you can turn this feature on or off
    },
    nuke_control_new = {
        enabled = false, -- it won't be loaded at all
        runtime = true, -- just a useless config option
    }
}
```

In a feature we can then do something like
```lua
local Feature = 'util.feature'
local Event = 'util.event'

function nuke_control.register(config)
    -- config = the contents of the above nuke_control table
    -- only called if enabled = true, otherwise this file isn't even loaded

    Event.add(..., function (event)
        if not Feature.enabled('nuke_control') then
            return
        end
    end)
end
```

Perhaps it can be made even fancier than this, just a quick example based on how I did it for Diggy.

Valansch:
I don't think it needs to be specified in the config whether a feature can be disabled on runtime. This is not a variable but a fact. A feature that can be disable on runtime should simply implement a `Module.enable(enabled)`.

I like the diggy approach. I like how diggy handles it, but diggy is more or less one cohesive feature so having a central config and control all subfeatures makes sense.

I'm not sure this is the best idea for the rest of our features. If the features modules need to interact with each other, they should require each other. If we have features writing and reading from config we break this directionality. In this case we would have to think about the order of features loaded.
Also features should be as independent as possible for two reasons:
1. Keep the learning curve for new players as low as possible.
2. Make importing existing modules from new people/other projects easier (without having to go convert the whole module)

I think maps should be able to disable certain features/change configs, while other features should be forbidden from doing so, unless the feature provides this functionality directly.

Sorry for my rambling, I'm not sure if I am stating the obvious at this point or repeat something that was already said but this is my proposal:
- We use the diggy approach for configs
- Utils/Lib are always to be kept independent from the rest of the codebase. They don't read or write to config.
- map_gen/ writes to config, never reads from config. Config is loaded before map_layout.lua.
- features/ reads from config, never writes to config.

linaori:
That makes sense, but what if features are disabled while another feature depends on it? That's currently happening

Valansch:
Depends on what disabled in this context means. 


1. A feature is disabled in the config (by map_gen/):
 - It is not loaded by config.lua
 - It is loaded by the feature that requires it
 - Everything working as intended.

2. A feature is disabled on runtime:
  - It is loaded by config.lua and the referencing feature
  - Module.enable(false) takes care of the complexity that comes with disabling such a feature



If you need more complex behavior from a feature this should not be handled by the config but by Module.change_something_about_module_behaviour.

PR #504, opened 2018-12-02
Added particle limits and scales

### LINKED ISSUE #495: System to limit the amount of particles
Currently create_particles.lua is just spawning particle after particle. However, end-game this becomes a bit more expensive. I would like to propose a command to alter the amount of particles the same way as gamespeed works. Example: `/particles 1` would set it to 100 where as `/particles 0.5` would set it to 50%.

linaori:
This PR adds the following:
 - the option to scale the total particles
 - an upper limit per 191 ticks of `12000 * (scale + 0.1)` particles
 - the ability to (run-time) change the scale of particles in case of performance issues (admin/server only)

While this doesn't fix the entirety of nukes in Diggy, it does no longer crash the server. Previously it would spawn up to 450k particles, which is now limited to 13~26k (depending on when it resets).

Some particles follow a ceil instead of floor rounding, like ceiling crumbling, as this is a game mechanic rather than a visual enhancement.

Fixes #495

### DIFF
--- a/features/create_particles.lua
+++ b/features/create_particles.lua
@@ -1,8 +1,120 @@
 local Task = require 'utils.Task'
+local Global = require 'utils.global'
 local Token = require 'utils.token'
+local Command = require 'utils.command'
+local Event = require 'utils.event'
 local random = math.random
+local ceil = math.ceil
+local floor = math.floor
+local format = string.format
 
 local CreateParticles = {}
+
+local settings = {
+    scale = 1.0,
+    particles_spawned_buffer = 0,
+    max_particles_in_three_seconds = 12000,
+}
+
+Global.register({
+    settings = settings,
+}, function (tbl)
+    settings = tbl.settings
+end)
+
+---sets the scale of particles. 1.0 means 100%, 0.5 would mean spawn only 50% of the particles.
+---@param scale number
+function CreateParticles.set_scale(scale)
+    if scale < 0 or scale > 1 then
+        error(format('Scale must range from 0 to 1'))
+    end
+
+    settings.scale = scale
+end
+
+---Returns the current scale
+function CreateParticles.get_scale()
+    return settings.scale
+end
+
+local function get_particle_cap()
+    return settings.max_particles_in_three_seconds * (settings.scale + 0.1)
+end
+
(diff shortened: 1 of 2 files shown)


features/create_particles.lua
```diff
@@ -1,9 +1,121 @@
...
+local function may_spawn_particles()
+    return settings.particles_spawned_buffer < get_particle_cap()
+end
+
+--- resets the amount of particles in the past 3 seconds so new ones may spawn
+Event.on_nth_tick(191, function ()
+    settings.particles_spawned_buffer = 0
+end)
+
+Command.add('set-particle-scale', {
```
mheguy: Essentially the same as in 505:
I would feel much better about a single command. `particle-scale <get/set>`
We're adding commands regularly and not really removing any so it's creating a sort of bloat in the number commands listed in /help and the like.

linaori: I'm not fond of conditional commands based on a second argument, as it provides no auto-complete. What I _can_ do, is make it like this: `/particle-scale <new_scale:optional>`. Omitting = get, passing = set. Not sure how common this is, but `jQuery`, an (in)famous javascript library, has done this for ages. `attr.html()` is a getter while `attr.html('foo')` is a setter.

Valansch: How about you compromise? :P 
We can do it like with /regular.
`/particle-scale get` getter
`/particle-scale` also getter
`/particle-scale set`

linaori: Because then I'd add complex logic and the help would not be descriptive as I'd have to call my parameter `get_or_set` or something

Valansch: Just hide this from the description. just treat `/particle-scale` as an alias to the real command `/particle-scale set`

linaori: I'll figure something out when I get home 😄

mheguy: Another argument for being able to pass custom /help text :p

linaori:
The command can now be used as:
 - `/particle-scale 0.1` -> sets the particle scale to 0.1
 - `/particle-scale` -> prints the current settings

Valansch (approved):

mheguy (approved):
Love it, will merge this shortly.

PR #505, opened 2018-12-02
Added a scale to change game performance and running speed accordingly

linaori:
It's a pretty basic script that assumes nobody has messed with either the game.speed or force speed movement modifiers. I recommend to turn this feature off when you're in a scenario that messes with those values. 

image


linaori:
Note that if #504 is merged, I can update this function also take the particles along.

features/performance.lua
```diff
@@ -0,0 +1,59 @@
...
+
+Command.add('set-performance-scale', {
+    description = 'Sets the performance scale between 0.05 and 1. Will alter the game speed and character running speed per force.',
+    arguments = {'scale'},
+    admin_only = true,
+    allowed_by_server = true,
+}, function (arguments, player)
+    local scale = tonumber(arguments.scale)
+    if scale == nil or scale < 0.05 or scale > 1 then
+        player.print('Scale must be a valid number ranging from 0.05 to 1')
```
grilledham, line 40: If this command is used on the server won't `player` be nil?

features/performance.lua
```diff
@@ -0,0 +1,59 @@
...
+    local p = game.print
+    p('## - Changed the game speed and running speed.')
+    p(format('## - Game speed: %.2f', Performance.get_scale()))
+    p(format('## - Force running speed: %.2f', Performance.get_running_speed_modifier()))
+end)
+
+Command.add('get-performance-scale', {
+    description = 'Shows the current performance scale.',
+}, function (_, player)
+    local p = player.print
```
grilledham, line 54: same as above.
If this command is used on the server won't `player` be nil?

linaori: This one is not executable by the server

features/performance.lua
linaori: True, but then it will cascade as error and you will see it anyway (I can fix it though)

features/performance.lua
grilledham: Is there a reason why a user entering the command at the server console shouldn't be allowed?

linaori: No, it's the default config and I haven't had a reason to add it so far. Now that I think of it, it might be desired to use this on the server so nobody has to log in to the server to see/execute it.

features/performance.lua
```diff
@@ -0,0 +1,59 @@
...
+end
+
+---Returns the current scale
+function Performance.get_scale()
+    return game.speed
+end
+
+---Returns the running speed modifier
+function Performance.get_running_speed_modifier()
+    return 1 / game.speed
```
Valansch, line 29: This will break any other feature that modifies character running speed.
There is two issues:
1. Force vs Player:
    If the player has a running speed buff (for example from the market) this buff will be removed if the force modifier is    changed.
2. Modifiers of other features:
    If walking speed was, say 120% before chaning the scale, it will be set to 100% (adjusted for game speed) after the scale is set.

One possible fix could be:
```lua
--Register as global!
local player_running_speed_modifiers = {}
local force_running_speed_modifiers = {}

local function apply_running_speed_modifier_scale()
    for name, force in pairs(game.forces) do 
        force.character_running_speed_modifier = (force_running_speed_modifiers[name] or 0) / game.speed
    end
    for index, player in ipairs(game.players) do 
        local modifier = player_running_speed_modifiers[index]
        if modifier then 
            player.character_running_speed_modifier =modifier / game.speed
        end
    end
end

function Performance.get_force_running_speed_modifier(force)
    return force_running_speed_modifiers[force.name] or 0
end

function Performance.set_force_running_speed_modifier(force, modifier)
    force_running_speed_modifiers[force.name] = modifier
    apply_running_speed_modifier_scale()
end

function Performance.get_player_running_speed_modifier(player)
    return player_running_speed_modifiers[player.index] or 0
end

function Performance.get_player_running_speed_modifier(player, modifier)
    player_running_speed_modifiers[player.index] = modifier
    apply_running_speed_modifier_scale() -- this can be optimized. Don't need to set ALL modifiers again
end

```
All features need to use this interface. I don't see another way this could work without either "Performance"breaking other features or other features breaking "Performance"

Valansch (dismissed):
This would break other features modifiers.

We should use this to increase the mining speed too. With my suggested changes this wouldn't break Diggys modifiers.

features/performance.lua
linaori: It's a bit of a "last-resort" command. Generally speaking there are no player movement speed buffs though. I'm not sure if it really has to be _that_ complex.

Valansch: I can accept that. No need to over do it. But on the other hand its already half implemented.
You can dismiss my review if you like then.

What about mining speed? You want to not touch it because of diggy?

linaori: When we're at the point where we need to change the game speed, people have bots anyway

Valansch: Okay

mheguy:
>We should use this to increase the mining speed too. With my suggested changes this wouldn't break Diggys modifiers.

And bot speed.

PR #506, opened 2018-12-03
Ensure donators are treated as regulars

mheguy:


linaori (approved):

Valansch:
No need to make it have it backwards compatible anymore. You can remove  `or global.regulars[player_name:lower()])`

grilledham:
I personally don't like this. 

I don't think it's that hard to remember to add donators to the regular list when we add them, and if we miss any we can simply promote them to have the problem fixed

The current system gives us the choice to demote rogue donators, instead of having to resort to a ban.
Not that I think donators going rogue is a common problem, but it's nice to have the choice.

mheguy:
I'm not attached to it either way/it's not something I would fight for. If you feel passionately about it I'll happily go your way.

PR #508, opened 2018-12-03
Added spiral_crossings map preset.

Jayefuu:


mheguy:
Tested fine. Merging with my thanks.

PR #509, opened 2018-12-05
Add shuffle_table to table utilities

mheguy:


utils/table.lua
```diff
@@ -102,6 +106,20 @@ table.get_random_weighted = function(weighted_table, item_index, weight_index)
     end
 end
 
+--- Creates a fisher-yates shuffle of a sequential number-indexed table
+-- from: http://www.sdknews.com/cross-platform/corona/tutorial-how-to-shuffle-table-items
+-- @param table takes a table
+table.shuffle_table = function(t)
+    local rand = math.random
```
linaori, line 113: This one can be replaced with `random`, which is already imported

utils/table.lua
```diff
@@ -102,6 +106,20 @@ table.get_random_weighted = function(weighted_table, item_index, weight_index)
...
+    local iterations = #t
```
linaori, line 114: Can you `error()` if size is 0? This in case someone uses it with a non-sequentially indexed array

linaori (commented):
Nice! I can probably put this to use somewhere in the future

Valansch:
I suggest this change:

```lua
table.shuffle_table = function(t, rng)
    local rand = rng or math.random
```

utils/table.lua
mheguy: Yeah that was silly of me. =)

utils/table.lua
mheguy: Can do.

mheguy:
> I suggest this change

Done

mheguy:
Closing this PR as it will be included as part of another. The suggested changes have all been made.

PR #510, opened 2018-12-05
Docs into wiki

mheguy:
The contents of all these docs are now in the wiki.


linaori:
Can you keep the files but instead put a link to the corresponding page on the wiki? If we don't do this, people will arrive on a 404 page on github, which would hurt every existing link out there atm, both from valansch's old repository and any direct links 😅 

Can also just refer to the wiki index instead if it's too much effort to keep it in sync.

mheguy:
Makes sense to me. Force-pushed the change since undoing the commit was the easiest way to get the right files back in the right places. :D

linaori (approved):
Nice!

Issue #298, opened 2018-11-11
Codify the terminology we use for maps


plague006:
In a perfect world, anyone could open up `map_layout.lua` and actually make sense of what can be combined with what. With that goal in mind I have a few interconnected proposals.

1. Presently in `map_layout.lua` we use some terms but don't define them:
`combined`, `presets`, `shapes`, `terrain`, `entity_modules`, `terrain_modules`, `everything else`
It would be nice to have these, and some other terms and definitions codified so that we can better categorize the files/modules in the repository and create clearer documentation.

2. I would like to have `map_layout.lua` split into 2 files. One which allows a user to create their own map by selecting modules, and one that only has preset maps and scenarios to allow the user to easily load up their favourite map/scenario in the easiest way possible.

So as a jumping off point to creating a map building vocabulary we can all share and agree on I have the following:

`the map`
A collective term for the tiles, entities, and resources which make up the game surface.

`resources`
A subset of entities which include coal, iron, copper, oil, stone, and uranium.

`significant gameplay features`
Gameplay changes beyond the enabling/disabling of recipes/techs and altering the market.

`shape_module`
Creates shapes only by turning tiles to void. Conflicts with `terrain_module`s. May conflict with other `shape_module`s. Does not affect entities or resources. (ex. x_shape)

`terrain_module`
Creates specific designs by changing tiles to other types of tiles. Conflicts with `shape_module`s and other `terrain_module`s. Does not affect entities, resources, or gameplay.  (ex. )

`ore_module`
Creates specific designs or patterns with resources. May replace vanilla ore generation. Does not affect tiles or gameplay. May conflict with other `ore_module`s. Should not, but may, affect entities. Whether a module removes vanilla ore generation or not should be included in a comment at the top of the file. (ex. glitter_ores)

`entity_module`
Creates specific designs or patterns with entity placement. Does not affect resources, tiles, or gameplay. May conflict with other `entity_module`s. May affect vanilla generation of entities. Whether a module removes vanilla entity generation or not should be included in a comment at the top of the file. (ex. worms)

`gameplay_module`
Only adds gameplay elements. May not affect the map. (ex. danger ores)

`day_night_cycle_module` 
Only sets the day/night cycles of a map. May not affect the map or gameplay.

----

`preset_map`
A lua file that is meant to be run with no other modules. May affects the map, resource, and/or entity generation. Does not introduce significant gameplay features/changes.

`scenario`
[...]

PR #516, opened 2018-12-06
Change map loading method

mheguy:
This change is primarily for end users, but has a couple nice effects for us.

This changes the way we load maps. Instead of commenting or uncommenting lines in a file, we plop the name into a file. 

For end users: no more looking at some ungainly 150+ line file. Easier process to select the map. Users don't even need to understand what a comment is.

For us: when we need a map we no longer have to change a file undergoing actual development. Instead we can add `map_selection.lua` (which will likely remain static) to our git exclude and never have to discard changes to it/accidentally commit changes to it. In addition, because of the simplicity of `map_selection` and `map_names`, it opens the door to creating a front-end at some point. Lastly, I think it's roughly 2% easier and quicker to add to the table instead of copy-pasting the whole `shape = require` line.

~~I intentionally have CRLF line endings in map_selection to maximize accessibility. (Because while notepad needs to die in a fire, it's still the default on most windows machines)~~ I forgot that git would eat those during the upload. Forget that.

This also begins the process of structuring the maps a bit.


map_selection.lua
```diff
@@ -0,0 +1,21 @@
+local map_name = 'default'
+return map_name
```
linaori: Can this one be moved to below the comment? Otherwise `return 'default'` would suffice I think

map_selection.lua
```diff
@@ -0,0 +1,21 @@
...
+
+--[[
+Replace the word default in the quotes above with the name of the map you want to play then save this file.
+Names of some popular maps:
```
linaori, line 8: Could we make a full list of possible maps here? That would make things a lot easier

linaori (commented):
This could work very nicely in combination with the scenario changes I proposed in the comment.

As a reminder, we have to update the Diggy docs example regarding the map_layout when merging this.

map_gen/map_loader.lua
```diff
@@ -0,0 +1,26 @@
...
+local regen_decoratives = false
+local tiles_per_tick = 32
+local shape
+
+local maps = require('map_names')
+
+if map_name == 'default' then
+    return
+elseif table.contains(maps['shape_maps'], map_name) then
+    shape = require('map_gen.presets.' .. map_name)
```
Valansch: This *will* break the delfate script. All presets will be loaded.

map_gen/map_loader.lua
```diff
@@ -0,0 +1,26 @@
...
+local map_name = require 'map_selection'
+local regen_decoratives = false
+local tiles_per_tick = 32
+local shape
+
+local maps = require('map_names')
+
+if map_name == 'default' then
+    return
+elseif table.contains(maps['shape_maps'], map_name) then
```
Valansch: This will make it impossible to combine shape maps with any other map type. Why do you split it into shape non_shape?

map_gen/map_loader.lua
Valansch: Delfate, heh. The fate of Del. Can you even spell @Valansch?!
I will change the deflate script instead.

map_gen/map_loader.lua
mheguy: I'm not removing any functionality that I'm aware of. `map_layout` is explicitly an option for that reason.

Further, it's incredibly unclear to an end user which maps (if any) can actually be used together in any rational way. Even as one of the devs I can't look at `map_layout` and tell you if two pieces will go together.

map_gen/map_loader.lua
mheguy: In that case, this can wait for https://discordapp.com/channels/432567222481846283/432570869039104012/520218000243294209

It's no rush in my book.

map_selection.lua
mheguy: We could plop all the names, but the vast majority go unused. I opted to include the maps with the highest hour counts. Are there any you think should be included that are missing?

map_selection.lua
mheguy: Having the return statement after the comment makes sense to me just to make it that little bit easier to read. I've made that change.

map_selection.lua
mheguy: I added more maps to the list, let me know if you think I missed any popular ones.

linaori: Nice, I think this will do. Though I wonder, where could one find the full list?

map_gen/map_loader.lua
Valansch: Thats why it's explained in the file head (I think^^)
So you want to have two ways to load a shape map? With map layout and without it? This seems counterproductive. We should choose one.
I thought you only left it until every map is converted.

map_selection.lua
mheguy: I'll move that line from lower to higher up: https://github.com/Refactorio/RedMew/pull/516/commits/a83751eda2567df05e06064e57c35dfaaec172cc

map_gen/map_loader.lua
mheguy: Map layout:
I want to leave `map_layout` in place for now in case someone wants some of the options currently only offered there (ex. combining maps, including `entity_modules`, etc.)

Shape/non-shape loading:
Having all maps use the same methods would be cool, but is that possible/practical? Could diggy ever return a shape?

Shape/non-shape being in 2 tables:
I figured out a way to get shape and non-shape maps into the same table, that should make for easier reading. I tested 2 of the non-shape maps and they seem to load fine so I don't *think* I created any issues with that.

Valansch: This is how its indended to work. This is why there are different categories. This is what the current implementation breaks.

mheguy:
> As a reminder, we have to update the Diggy docs example regarding the map_layout when merging this.

Nope. The diggy guide uses the release system, so extract and play. I updated the default user guide in the wiki, as well as tetris which will leave them out of sync until the next release, hard to avoid that. https://github.com/Refactorio/RedMew/wiki/New-scenario does need changing, but that will depend heavily on #298

map_gen/map_loader.lua
mheguy: >**This** is how its indended[sic] to work. This is why there are different categories. **This** is what the current implementation breaks.

I'm not sure what 'this' you're referring to, so I'm not sure what I've broken or how to fix it.

Valansch: I will make a review.

map_gen/map_loader.lua
```diff
@@ -0,0 +1,24 @@
...
+local regen_decoratives = false
+local tiles_per_tick = 32
+local shape
+
+local maps = require('map_names')
+
+if map_name == 'default' then
+    return
+elseif table.contains(maps, map_name) then
+    shape = require('map_gen.presets.' .. map_name)
```
Valansch: There are 86 entries in the map_names. Presets contains 70 files. This cause a file not found error for the remaining 16 names.

Valansch (changes requested):
I have a few problems with these changes. 

Now there are 2 ways to do the same thing, 2 places to maintain and keep in sync. This is bad practice and in no way makes it easier for us.

As you state this is to make it easier for end users. But our end users are the players on the RedMew server. I disagree with compromising compromising for making it marginally easier for people unaffiliated with RedMew to select a map.

>This changes the way we load maps. Instead of commenting or uncommenting lines in a file, we plop the name into a file.
> ...
>Lastly, I think it's roughly 2% easier and quicker to add to the table instead of copy-pasting the whole shape = require line.

I disagree. Now we need to add it in both places. Removing comments is easier 

>For us: when we need a map we no longer have to change a file undergoing actual development. Instead we can add map_selection.lua (which will likely remain static) to our git exclude and never have to discard changes to it/accidentally commit changes to it. 

If this is an issue, we can add map_layout.lua to .gitignore. This will force you to add it manually, no accidental commit possible.

*Unless* you want to remove map_layout.lua entirely. In this case we cannot load two different map modules at the same time. Completly breaking this feature.
I also find the distinction between the modules very usefull. Simply putting them into one big table makes it IMO harder to read, harder to find modules im looking for and harder to understand which modules are compatible with each other, as this information is completly discarded.

mheguy:
> As you state this is to make it easier for end users. But our end users are the players on the RedMew server. I disagree with compromising compromising for making it marginally easier for people unaffiliated with RedMew to select a map.

I'm in favor of this. I would save a lot of time on keeping the changelog up to date, on issuing releases, and on documenting how to use RedMew scenario code.

> Unless you want to remove map_layout.lua entirely. In this case we cannot load two different map modules at the same time. Completly breaking this feature.

Yeah let's do that too actually. Loading map modules outside of map presets isn't as much a feature as is it a hindrance. I don't want to guess at what modules are required by map presets and I definitely don't want to have to toggle them every time I generate the map, let them be required in the presets themselves.

map_gen/map_loader.lua
mheguy: Fixed in next commit. 90 files in preset, 90 entries in table. I had neglected to move the `shape` maps from their folder.

Valansch:
>I'm in favor of this. I would save a lot of time on keeping the changelog up to date, on issuing releases, and on documenting how to use RedMew scenario code.
I don't understand your point, what does that have to do with this issue?

ALL modules are standalone. The presets have nothing to do with them. Presets are only special kinds of map modules that use grilleds shape builders. Why would you need to remember anything? You pick and choose what you want. Don't you know what you want?
If we remove map layouts i can no longer make a map that loads terrain module AND a misc module. How is that ability a hindrance?

mheguy:
> Don't you know what you want?

If I'm creating a map? Sure I know what I want and I make sure it's included in the preset. If I'm trying to use someone else's map? No, I don't know what they wanted. If I want to load up the danger_ores preset or beach or triangle_of_death I don't know what modules are or aren't explicitly needed (or compatible). I have to guess.

Meanwhile with crashsite, diggy, venus, etc. I don't have to worry about it: I just load the preset and it's done.

> If we remove map layouts i can no longer make a map that loads terrain module AND a misc module. How is that ability a hindrance?

That is just incorrect. Misc modules have nothing to do with the builders, they don't return anything for the builders.

I think you meant is loading a preset map and a terrain/entity module. I'll add instructions for how to include them in the files themselves, as well as the [creating a new scenario](https://github.com/Refactorio/RedMew/wiki/New-scenario) wiki entry.

As you said yourself, our end users are the players on the server. Therefore this creates a "canonical" version of map presets making it easier to serve those maps to our users. If map X needs infinite chests to be played, it should be included in the preset. If another map-maker wants to create a version of map X without infinite chests all they need do is open the preset and comment out the require line for the infinite chests.

mheguy:
Closing as this has gone stale and #555 will cause more merge conflicts. Will create a new PR after #555 is merged.

PR #522, opened 2018-12-07
Cleaned up some diggy commands and the left-overs register in _DEBUG

linaori:
Not much to add, just some cleanup and making things less complex


mheguy:

mheguy (approved):

PR #523, opened 2018-12-07
Added the crash-site GUI features to the retailer

linaori:
It's pretty much an "as is" port, no added features, no removed features. The Retailer hides the implementation details of the GUI so not much changed for Diggy, except not having to ship items anymore as markets no longer hold the items. A minor difference in setting an item, no longer needs a complex price array as it only works with coins now.

```lua
local Retailer = require 'features.retailer'
local group_name = 'default'

Retailer.set_item(group_name, {name = 'steel-axe', price = 10})
-- Retailer.remove_item(group_name, 'steel-axe')

local market_entity = game.surfaces.nauvis.create_entity({name = 'market', position = {0, 0}})
-- add a market
Retailer.add_market(group_name, market_entity)

-- shows all items for a specific group
game.print(require 'utils.debug'.print(Retailer.get_items(group_name)))
```

I haven't tested if the GUI updates during a level up, but that's something for later. I was also thinking of adding/removing items via a command for admins only in a later stage, adding descriptions instead of the raw item name,the possibility to add your own icon and eventually also events that trigger when you buy something.

Starting out with nothing
image

First level
image

Level 70
image

### DIFF
--- a/features/retailer.lua
+++ b/features/retailer.lua
@@ -1,10 +1,23 @@
 local Global = require 'utils.global'
-local insert = table.insert
+local Gui = require 'utils.gui'
+local Event = require 'utils.event'
+local PlayerStats = require 'features.player_stats'
+local Game = require 'utils.game'
+local math = require 'utils.math'
+local format = string.format
+local concat = table.concat
+local clamp = math.clamp
+local floor = math.floor
+local ceil = math.ceil
+local market_frame_name = Gui.uid_name()
+local item_button_name = Gui.uid_name()
+local count_slider_name = Gui.uid_name()
+local count_text_name = Gui.uid_name()
 
 local Retailer = {}
 
 ---Global storage
----Markets are indexed by the group_name and is a sequential list of LuaEntities
+---Markets are indexed by the position "x,y" and contains the group it belongs to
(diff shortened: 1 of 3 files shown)


grilledham:
Looks good.

We should probably add support for a currency item that isn't coins, or even for multiple items.

I'm excited for the new features you suggested to add. 

I'm pretty sure the UI only updates when it is first opened or when the values in the slider or textfield change. Fixing that would require keeping track of which players have the market gui opened.

linaori:
I think those would be nice additions indeed, I'm not sure if I'll manage though, GUI looks like voodoo to me. Only reason I got it to work so easily was because by some chance, the structure was fairly similar to what I implemented in the Retailer

grilledham:
I could do the UI stuff, but it'd have to wait until after the 17th

linaori:
No rush. This can be merged as is (once the review is approved), as it remains status quo

mheguy (approved):
This looks good, I like the feature set, I look forward to it being available on all maps at some point. :)

mheguy:
As this works with Diggy now and doesn't touch any other maps, I'll merge this with my thanks.

PR #524, opened 2018-12-08
Overhauled day_night_cycle, creep_spread, and nightfall. Add Venus map.

mheguy:
There are still optimizations to be made, but I felt I needed to PR this sooner rather than later because of how bloated the branch became.

In particular: better threading for both nightfall and terraforming and creating a generic feature out of Venus::world_settings (and likely at the same time pushing all maps to start using a second surface and having `nauvis` become a kind of lobby.


map_gen/misc/day_night.lua
```diff
@@ -0,0 +1,92 @@
...
+    end
+end
+
+--- Sets the day/night cycle according to the table it is given.
+-- @param day_night_cycle table containing keys: ticks_per_day, dusk, evening, morning, dawn
+-- @param surface the LuaSurface to set the day/night cycle of
+-- @returns boolean true if set properly
+Public.set_cycle = function(day_night_cycle, surface)
+    if not check_cycle_validity(day_night_cycle) then
+        error('[DAY_NIGHT] Provided day/night cycle is invalid')
```
linaori: What's `[DAY_NIGHT]` referring to?

map_gen/misc/day_night.lua
```diff
@@ -0,0 +1,92 @@
...
+    local breakpoint = (daylight - y_intercept) / brightness_slope
+    -- Set the daytime to the point we just calculated
+    surface.daytime = breakpoint
+    -- Freeze the day/night cycle
+    surface.freeze_daytime = true
+
+    if _DEBUG then
+        game.print('[DAY_NIGHT] breakpoint/surface_daytime: ' .. breakpoint .. '/' .. surface.daytime)
+    end
+    if _DEBUG then
```
linaori: These 2 could be merged, or instead, `local Debug = 'utils.debug'` and then `Debug.print('your message')` works too by now. Advantage here is that each message gets a uniquely incremented number in front of it, so that the console doesn't merge the message if posted frequently, and only triggers if `_DEBUG` is true

map_gen/misc/day_night.lua
```diff
@@ -0,0 +1,92 @@
...
+        game.print('[DAY_NIGHT] brightness/surface_brightness: ' .. math.round(daylight, 2) .. '/' .. math.round((1 - surface.darkness), 2))
+    end
+
+    if math.round(daylight, 2) == math.round((1 - surface.darkness), 2) then
```
linaori, line 85: Minor thing, referencing `local round = math.round` in the top of the file will give a raw 30% performance boost in those calls

map_gen/misc/nightfall.lua
```diff
@@ -10,150 +10,166 @@ The above copyright notice and this permission notice shall be included in all c
...
 local ATTACKING = 3
 
-local random = math.random
-local insert = table.insert
+-- create globals
+local bases = {}
+local c_index = {1}
+local chunklist = {}
+local state = {1}
+local lastattack = {0}
```
linaori: You could also do something like:
```lua
local memory = {
    base = {},
    c_index = 1,
    chunklist = {},
    state = 1,
    lastattack = 0,
}
```
This would make it less cumbersome to store scalar values in globals

map_gen/misc/nightfall.lua
```diff
@@ -10,150 +10,166 @@ The above copyright notice and this permission notice shall be included in all c
...
         local chunk = {}
         local coords = event.area.left_top
         chunk.x = coords.x + 16
         chunk.y = coords.y + 16
-        insert(global.chunklist, chunk)
+        insert(chunklist, chunk)
     end
 end
 
+--- Every tick, choose between searching for bases, preparing an attack, or doing nothing
```
linaori, line 146: This sounds very heavy, but in fact it looks like it's 1 chunk per tick. Can you add this to the comment so the next person won't be like "oh wow, this is inefficient!" 😄

linaori (changes requested):

map_gen/misc/day_night.lua
mheguy: Just the module that the error is coming from. I guess with `error` it would show that. It's more a holdover from other debug messages where I'm printing.

map_gen/misc/day_night.lua
mheguy: Debug.print it is! Awesome. 😄

map_gen/misc/day_night.lua
mheguy: Removed the indicator.

map_gen/misc/day_night.lua
mheguy: I considered that, but you only set a fixed brightness once a map. If someone ends up abusing it and calling it multiple times I'll happily refactor for performance.

If you feel strongly about it I'll change it, or if you want to commit the change I'll be happy to have it.

map_gen/misc/nightfall.lua
mheguy: Funny enough, this would have saved us the desyncs. :D
Covered in bf799718d28e7bd710ef0709c3756509ab2b8b06

map_gen/misc/nightfall.lua
mheguy: I'm not sure the documentation here is inaccurate. All it does in the `on_tick` call is choose the path. `find_bases` and `biter_attack` are fairly well documented so I would rather not copy+paste the full documentation for each, but I can add a line telling the reader check the definitions of those functions for further documentation?

map_gen/misc/day_night.lua
linaori: It doesn't matter if it gets called once, I thought it was called more often. Carry on! 👍

map_gen/misc/nightfall.lua
linaori: That could help already yes! 👍

Valansch:
Thank you. I will merge this and launch another test soon

PR #526, opened 2018-12-08
refactored globals

grilledham:
I took the code from the current live map and edited it.
Hopefully it fixes the desync issue.


Valansch (approved):

mheguy:
You can just commit to the open PR :p

grilledham:
This way is more fun

PR #528, opened 2018-12-08
Ore tendril pattern with slight impurities

BlueRock2:
Instead of the heavily mixed ore pattern currently used with Diggy, this alternative ore pattern contains ore tendrils that are about 80% pure ore with lower impurities.   Feedback encouraged.


linaori (approved):
I can't see anything wrong with this, as it's not enabled by default, it looks safe to merge

mheguy:
Looks good in testing. I think you should add it as an option on line 302 of diggy config

mheguy (approved):

mheguy:
image

linaori (approved):
I'm still afraid the tendrils are too big, but it looks really good like this already

PR #529, opened 2018-12-08
Gave the player_create feature the same as diggy SetupPlayer

linaori:
Enables config for:
 - starting items
 - join messages
 - weighted join message
 - whether or not to show the info at the start
 - cheats (combined with diggy cheats)
 - now only sets the force config once
 - Diggy's SetupPlayer now configures player_create instead

### DIFF
--- a/config.lua
+++ b/config.lua
@@ -14,6 +14,8 @@ global.config = {
     -- New Scenario Features, appears in the "What's new" tab
     new_info_key = 'Nothing is new. The world is at peace',
 
+    -- Adds a command to scale UPS and movement speed. Use with caution
+    -- as it might break scenarios that modify movement speed
     performance = {
         enabled = true,
     },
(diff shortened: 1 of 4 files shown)


config.lua
```diff
@@ -97,6 +99,66 @@ global.config = {
...
+            -- increase the amount of inventory slots for the player force
+            character_inventory_slots_bonus = 0,
+
+            -- increases the run speed of all characters for the player force
+            character_running_speed_modifier = 5,
+
+            -- a flat health bonus to the player force
+            character_health_bonus = 1000000,
+
+            -- unlock all research by default, only useful when testing
```
mheguy: When `_CHEAT` or `_DEBUG` we have the `/all-tech`. I would propose not automatically applying all techs since it's easily applied (via the command) but much less simple to reverse (speaking for myself, I need to check the wiki each time)

mheguy (approved):
Looks good to me, whether you make the change I suggested or not this is good to merge in my books.

config.lua
linaori: Hm yeah if we have this command, might as well remove this option

PR #530, opened 2018-12-08
Added a command-search command

linaori:
Lets you search for a command with a keyword, acting like a simple filter.

image

### DIFF
--- a/features/custom_commands.lua
+++ b/features/custom_commands.lua
@@ -8,7 +8,6 @@ local Report = require 'features.report'
 local Server = require 'features.server'
 local Timestamp = require 'utils.timestamp'
 local Command = require 'utils.command'
-local string_length = string.len
 local format = string.format
 
 --local Antigrief = require 'features.antigrief'
@@ -605,7 +604,7 @@ Command.add('search-command', {
 }, function (arguments, player)
     local keyword = arguments.keyword
     local p = player.print
-    if string_length(keyword) < 2 then
+    if #keyword < 2 then
         p('Keyword should be 2 characters or more')
         return
     end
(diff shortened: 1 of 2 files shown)


features/custom_commands.lua
```diff
@@ -8,6 +8,8 @@ local Report = require 'features.report'
 local Server = require 'features.server'
 local Timestamp = require 'utils.timestamp'
 local Command = require 'utils.command'
+local string_length = string.len
```
grilledham: the # operator works the same as string.len on strings, but that's just my personal preference.

grilledham:
I take it you know that if you type in part of a command and press tab the game tells you which commands match?

So the purpose of the search-command is that it also searches descriptions and / or provides a nicer output.

grilledham (approved):

features/custom_commands.lua
linaori: I was not aware of that. I will change this

linaori:
@grilledham exactly, in this setup you could search for all admin commands as an example.

mheguy:
I tried the example you cited of searching admin commands but it doesn't work as expected. It lists some number of commands but 15 are hidden. It suggests you refine your search but if you don't know what commands are available you don't know how to refine it.

I'm not sure how best to solve it but there are options. A system to change pages, or when there are too many results just showing a list of the commands and the suggestion of checking /help for each command. Alternatively: a hybrid system which shows x commands and their help and then a list of the remaining commands.

Vanilla commands don't display any text other than the command itself.

linaori:
I can tweak the number a bit, but if I show too many commands, it just becomes clutter on your screen. I could add paging, that's not really a problem: `/search-command <keyword> <page:optional>`

linaori:
image
image
image
image
image
image

mheguy:

PR #532, opened 2018-12-09
Made hail_hydra available to all redmew scenarios

linaori:
The configuration format isn't the most powerful, might have to change in the future as it was designed for Diggy. It's a good starting point to make this a generic RedMew feature though, making it easy to use everywhere. Tested in Diggy.

### DIFF
--- a/features/hail_hydra.lua
+++ b/features/hail_hydra.lua
@@ -1,0 +1,81 @@
+-- Cut off a limb, and two more shall take its place!
+local Event = require 'utils.event'
+local CreateParticles = require 'features.create_particles'
+local random = math.random
+local floor = math.floor
+local ceil = math.ceil
+local compound = defines.command.compound
+local logical_or = defines.compound_command.logical_or
+local attack = defines.command.attack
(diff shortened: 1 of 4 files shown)


features/hail_hydra.lua
```diff
@@ -0,0 +1,81 @@
...
+            if spawned and spawned.type == 'unit' then
+                spawned.set_command(command)
+            elseif spawned and cause then
+                spawned.shooting_target = cause
+            end
+        end
+    end
+end)
+
+return HailHydra
```
mheguy: Is this just returning the empty table created on L12?

linaori: Actually it is, I was thinking of adding some sort of a unit command builder in this thing, but decided to keep it as is for now. I can remove it if you want

linaori: I was just thinking though, if you require a hail hydra, you should probably also return a hail hydra 🤔 



_But I'll remove it_ 😞

mheguy (approved):

Issue #452, opened 2018-11-24
Change Task.lua, Queue.lua PriorityQueue.lua to conform to the naming convention of other files


mheguy:
`Task`, `Queue`, and `PriorityQueue` should conform with the other utilities' naming conventions. Non-caps, underscores and not camelcase.

grilledham:
Can we also change `global_token` to just `token`?
When I first implemented it we couldn't decide on the name which is why most code is
`Token = require 'utils.global_token'

linaori:
A few more points: 
 - `utils.list_utils` -> `utils.table`
 - `utils.utils`, not sure what to change it to though
 - make snake_case the convention to avoid windows/linux case sensitivity errors

mheguy:
Most of these will be addressed in #464.

Changing `Queue` to `queue` and `Task` to `task` will not be trivial as neither linux nor windows systems will automatically rename a file's casing by default.

I can think of 2 possible solutions:
* Rename the files to something intermediate (ex. tasks or task_rename) and then rename them a few weeks from now, once the intermediate name has propagated to all devs/forks.
* Rename the files to something else permanently.

Option 2 is cleaner, but requires coming up with new names.

linaori:
Git should be able to handle this: https://stackoverflow.com/questions/1793735/change-case-of-a-file-on-windows

Haven't looked into depth, but it seems possible, one way or another.

mheguy:
The problem is not pushing the change, it's the pull. By default git doesn't seem to change filenames when only the casing has changed.

I created a commit with an intermediate name and then another to name it back to be sure the name was changed. I pulled the commits onto the linux server that we host on, and the file names did not change.

PR #535, opened 2018-12-10
Rename task and queue

mheguy:
This can either be a step towards https://github.com/Refactorio/RedMew/issues/452 or close it fully if people find the names acceptable as long-term replacements. I personally don't like `processing_queue` and would want to change that back after 3-4 weeks when the change propagates to all (or most) forks.


mheguy:
After a quick look at current open PRs, this should not create issues with any.

linaori (approved):
This is a nice solution, I like it!

You're not actually threading anything, it's more like scheduling, but it feels like threading so I'm giving it a 👍  anyway

Valansch:
This used to be calles threading. :D We renamed it task as that was uhm more approriate? Task isnt good either, but its certainly not threading either. How about "async" or like lin proposed "scheduling".

linaori (approved):

grilledham:
I don't approve of renaming Queue to processing_queue, Queue is just a standard FIFO data structure. It has nothing to do with the scheduler, it just happens to be part of its implementation.
We should rename it to lower case however to be consistent.

linaori:
@grilledham that causes issues on windows

mheguy:
> We should rename it to lower case however to be consistent.

@grilledham @iltar that causes issues on linux as well. I detailed that in https://github.com/Refactorio/RedMew/issues/452#issuecomment-442008290 but as a tl;dr: unless set up to do otherwise, git will not rename files just for case. We just can't go from A to B. We need some kind of middle step.

What I was planning on doing originally with both `Task` and `Queue` was just renaming from from singular to plural forms. Would that be acceptable? To rename `Queue` to `queues` for a few weeks then change it to `queue` afterwards?

mheguy:
image

Valansch:
q... ffs. Well played.

PR #536, opened 2018-12-10
Change info icon to coin

mheguy:
"How do I use/get coins?"

Hopefully this reduces the frequency of this question.


linaori:
Isn't this the info screen though? It doesn't exactly make sense to me to have the scenario information under a coin button. Perhaps there are other ways we can explore for this?

mheguy:
Sure

PR #539, opened 2018-12-10
Fixed coin mining and behavior in Diggy

### LINKED ISSUE #527: Mining coins won't add to the sum when tracking them
**Describe the bug**
Right clicking a coin doesn't trigger it being added to the statistics

**Expected behavior**
It should be added

**Steps to reproduce**
1. Open the Playerlist GUI
2. Right click a coin on the map

**Screenshots**
n/a

**Map**
n/a

**Additional context**
Can be fixed by listening to the `on_player_mined_entity` event. I'm not sure if the robot loot has been added either, haven't tested it.

linaori:
Fixes #527 

Makes mining rocks in Diggy drop coins instead of directly inserting them into the inventory of players. This makes behavior for coin tracking easier. Additionally the coin per player has been enabled in the stats.

### DIFF
--- a/features/player_stats.lua
+++ b/features/player_stats.lua
@@ -1,6 +1,8 @@
 local Event = require 'utils.event'
 local Global = require 'utils.global'
 local Game = require 'utils.game'
+local pairs = pairs
+local sqrt = math.sqrt
 
 local player_last_position = {}
 local player_walk_distances = {}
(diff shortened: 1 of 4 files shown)


mheguy:
Lin brought up in discord that this doesn't cover the case of bots picking up coins: but that's fine. We don't need that level of accuracy since it's just a fun stat we track on the scoreboard and as it stands if the issue is raised it's "wontfix".

mheguy (approved):

PR #543, opened 2018-12-13
Diggy fixes, performance improvements and balance tweaks

linaori:
This change turned out a bit bigger than I hoped for. However, it fixes a bunch of things:
 - Reduced table access in frequently called places
 - Reduced calls to global functions (math, pairs, etc)
 - Added a config option for Hydra Mode that lets you tweak the modifier applied by evolution_factor
 - Starting with 2 axes instead of 1 and 12 walls instead of 10
 - Added train stuff to the market
 - Removed `surface.get_tile` calls when water was placed in room generation
 - Optimized some loops (pairs -> counts) that were frequently called
 - Increased reward from the chest raffle, added ammo, turrets, modules, rail stuff and some other goodies


mheguy:
The "chest" marker persists after mining the chest. I imagine that's not supposed to happen?

mheguy:
Chests will try to spawn in water:
image

mheguy (approved):
Code looks alright, issues in comments about chests seem to be beyond the scope of this PR. Testing showed no issues.

linaori:
@plague006 if you have heavy mode on, you'll also see biters spawn in water and then get removed, it's one of those artifacts that happen if you have 2 different features that do something with noise and are not working together. This will be solved once we have a single point of entry for all noise related features (which @Rdjrjqouqcu had some ideas for)

mheguy:
Since you're aware of the unrelated issues and this PR itself showed no new issues: merging with my thanks.

PR #545, opened 2018-12-14
Added electric poles to Support List

R-Nukem:
Added the three electric poles to map_gen\Diggy\Config.lua as supports with some initial values.

Copied stress_map_add to stress_map_add_pole in map_gen\Diggy\Feature\DiggyCaveCollapse.Lua.
     - Set the furthest support value to 0. This means all power poles only provide support in the center and disc regions, not the ring region.

Added a check to each entity add/remove event to check if the entity is a power pole. If yes proceeds to stress_map_add_pole, otherwise continues to stress_map_add as before.

### DIFF
--- a/map_gen/Diggy/Config.lua
+++ b/map_gen/Diggy/Config.lua
@@ -99,6 +99,9 @@ local Config = {
                 ['concrete'] = 0.04,
                 ['hazard-concrete'] = 0.04,
                 ['refined-concrete'] = 0.06,
+				['small-electric-pole'] = .25,
+				['medium-electric-pole'] = .5,
+				['big-electric-pole'] = 2
             },
             cracking_sounds = {
                 ' R U NY O UF O O L S !',
--- a/map_gen/Diggy/Feature/DiggyCaveCollapse.lua
+++ b/map_gen/Diggy/Feature/DiggyCaveCollapse.lua
@@ -41,6 +41,7 @@ local ring_value = 0
 
 local enable_stress_grid = 0
 local stress_map_add
+local stress_map_add_pole
 local mask_disc_blur
 local stress_map_check_stress_in_threshold
 local support_beam_entities
@@ -246,7 +247,14 @@ local function on_mined_entity(event)
         if not Template.is_diggy_rock(name) then
             player_index = event.player_index
         end
-        stress_map_add(entity.surface, entity.position, strength, false, player_index)
+		
+		if ( name == 'small-electric-pole' or name == 'medium-electric-pole' or name == 'big-electric-pole' ) then
+			stress_map_add_pole(entity.surface, entity.position, strength, false, player_index)
+		else
+			stress_map_add(entity.surface, entity.position, strength, false, player_index)
+		end
+
+        -- stress_map_add(entity.surface, entity.position, strength, false, player_index)
     end
 end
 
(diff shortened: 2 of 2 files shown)


map_gen/Diggy/Config.lua
```diff
@@ -99,6 +99,9 @@ local Config = {
                 ['concrete'] = 0.04,
                 ['hazard-concrete'] = 0.04,
                 ['refined-concrete'] = 0.06,
+				['small-electric-pole'] = .25,
+				['medium-electric-pole'] = .5,
+				['big-electric-pole'] = 2
```
linaori, line 104: Can you replace the tabs (in all your changes) to spaces? That will fix the indenting issues. We should probably add an editor config for this 🤔

map_gen/Diggy/Feature/DiggyCaveCollapse.lua
```diff
@@ -260,24 +265,39 @@ local function on_entity_died(event)
...
 local function on_placed_entity(event)
-    local strength = support_beam_entities[event.entity.name]
+	local entity = event.entity
+    local name = entity.name
+    local strength = support_beam_entities[name]
 
     if strength then
-        stress_map_add(event.entity.surface, event.entity.position, -1 * strength)
+		if ( name == 'small-electric-pole' or name == 'medium-electric-pole' or name == 'big-electric-pole' ) then
+			stress_map_add_pole(entity.surface, entity.position, -1 * strength)
```
linaori, line 297: This is a very fragile setup, as it will not be working well for mods that add different power poles

map_gen/Diggy/Feature/DiggyCaveCollapse.lua
```diff
@@ -533,6 +553,40 @@ end
...
+    local x_start = floor(position.x)
+    local y_start = floor(position.y)
+
+    local stress_map = stress_map_storage[surface.index]
+    if not stress_map then
+        return
+    end
+
+    if no_blur then
+        add_fraction(stress_map, x_start, y_start, factor, player_index, surface)
```
linaori, line 566: What if we allow setting a radius per entity in addition to a strength level? Meaning we could do something like this:
```lua
-- {strength, radius}
['stone-wall'] = {3, 4}
['stone-path'] = {0.03, 1},
['refined-concrete'] = {0.06, 1},
['small-electric-pole'] = {0.25, 2,
['medium-electric-pole'] = {0.5, 2},
['big-electric-pole'] = {2, 2}
```
This is a slightly bigger change, but will make it easier for mods to alter this and changes the `no_blur` to be a bit more useful. @Valansch what do you think?

Did you get to test if fast-replace works?

linaori (changes requested):
Thanks for looking into this! I have a few remarks as I think the current implementation is a tad fragile

map_gen/Diggy/Feature/DiggyCaveCollapse.lua
```diff
@@ -533,6 +553,40 @@ end
...
+        return
+    end
+
+    for x = -radius, radius do
+        for y = -radius, radius do
+            local value = 0
+            local distance_sq = x * x + y * y
+            if distance_sq <= center_radius_sq then
+                value = ((center_value*6)/4)
```
Valansch, line 575: These calculations should be in init anyways.

Valansch (changes requested):
I don't think this change is a good idea. Reasons in decending order of importance.

- Why do powerpoles need another mask? Whats wrong with using the same as for other entities?


- Power poles cost cpu. A lone power pole standing around being connected or not will cost cpu. The game checks each network each tick, so even unconnected poles add another loop network for the game to check. Connected poles are even worse as they grow the network, making it harder to add more poles with each additional pole.

- This will cause the matrix to no longer be normalized. different pole types will have different total strength added to the map only related to their strengh as state in the config.
If you have a support structure with strengt `x`, then total stress will be reduced by `x` will be applied to the stress map. This change will make it so poles with streng `x` reduce stress by `x+y`.

PR #546, opened 2018-12-14
Add player_reward, add coins as reward for looking at info

mheguy:
Creates a generic player reward utility: pretty straightforward, it's little more than player insert and an optional player print.

Ran into problems with the config since I want `player_rewards_token` to be the market item by default but can't reference the market item since it hasn't been defined yet. Not sure if this is the best way to handle it, I considered setting `player_rewards_token` after the `global.config` table was created, but no matter what it all looks messy.

Info.lua:
Added a mechanic that rewards players for looking at each tab by giving them coins.

Removed the admin/regular check since everyone will want the coins. Removed `draw_main_frame` call from `info::player_create` since it conflicts with recent changed to `player_create.lua`. Removed `welcomed_players` since it wasn't actually called by anything. 

Changed fish_market to keep the coin count the same.


utils/player_rewards.lua
```diff
@@ -0,0 +1,70 @@
...
+        player = Game.get_player_by_index(player)
+    end
+    local reward = {name = reward_token[1], count = amount}
+    if not player.can_insert(reward) then
+        return 0
+    end
+    if message then
+        player.print(message)
+    end
+    return player.insert(reward)
```
linaori: Should this feature add coins the same way the stats tracking does? Same question goes for coins as starting items. It's not an important thing to add, more of a "nice to have"

config.lua
```diff
@@ -181,6 +182,13 @@ global.config = {
             ['big-worm-turret'] = {['small-biter'] = 3.8, ['medium-biter'] = 1.3, ['big-biter'] = 1.1},
         },
     },
+
+    -- grants reward coins for certain actions
+    player_rewards_enabled = true,
```
linaori: Can you put this in the following structure?
```lua
player_rewards = {
    enabled = true,
    -- the token to use for rewards
    player_rewards_token = market_item,
    -- rewards players for looking through the info tabs
    info_player_reward = true,
},
```
I've omitted the `if true, will override player_rewards_enabled to true`, as it makes no sense to disable a feature, still configure it, and end up loading it anyway because of this value

features/fish_market.lua
```diff
@@ -273,7 +273,8 @@ local function player_created(event)
         return
     end
 
-    player.insert {name = market_item, count = 10}
+    local count = global.config.info_player_reward and 0 or 10
+    player.insert {name = market_item, count = count}
```
linaori, line 277: Should this increase the coin count in the stats tracking?

features/gui/info.lua
```diff
@@ -29,11 +46,11 @@ local editable_info = {
...
 Global.register(
     {
-        welcomed_players = welcomed_players,
-        editable_info = editable_info
+        rewarded_players = rewarded_players,
+        editable_info = editable_info,
     },
     function(tbl)
-        welcomed_players = tbl.welcomed_players
+        rewarded_players = tbl.rewarded_players
```
linaori, line 53: Can you change this behavior to be:
 - on player created -> add `[player_index] = true`
 - on awarded -> remove `[player_index] = nil`

By keeping track of only those who still need rewards, we effectively reduce the save size by quite a lot. Storing ~40bytes per item when we have 800 players on a map in total, becomes quite a bit of data to serialize, vs perhaps 200 players that haven't been rewarded yet.

utils/player_rewards.lua
```diff
@@ -0,0 +1,70 @@
...
+Public.get_reward = function()
+    return reward_token[1]
+end
+
+--- Gives the player the quantity of reward
+-- @param player player_index number or LuaPlayer table -- the player to reward
+-- @param amount number - the amount of reward tokens to give
+-- @param message string - an optional message to send to the affected player
+-- @return number - indicating how many were inserted or if operation failed
+Public.give_reward = function(player, amount, message)
```
linaori, line 38: This function is effectively returning how many coins have been inserted. In theory this will usually work, as coins can be up to 200k a stack. However, if there's no space, the reward will still be given, despite not being inserted into the inventory

linaori (changes requested):

utils/player_rewards.lua
mheguy: Really good suggestion (and implemented in the next commit). I had the specific info box use-case in mind when I was designing `player_rewards` and that restricted the rewards I was considering. If we use it for something like quests, listing the earnings in stats becomes important.

config.lua
mheguy: Your structure is much better/makes more sense and is now in place. I also agree with the whole override thing. I've correspondingly changed code in info/player_rewards to 1) not override the setting and 2) removed code specific to info_rewards from player_rewards itself.

features/fish_market.lua
mheguy: I think because it is such a small, fixed, and universal amount it's not worth bothering with. But if someone else wants to write the commit I also wouldn't oppose it.

mheguy:
In testing I noticed coins earned (`player_stats::player_coin_earned`) can be negative. I don't know if we want this behaviour or not.

utils/player_rewards.lua
mheguy: The function intentionally returns the amount of coins inserted, so that any call to it can keep track of how effective the function was.

Let's say we have a reward system for quest rewards:
```
function quest_complete(event)
  actual_reward = give_reward(event.player, reward_quant)
  outstanding_reward = actual_reward < reward_quant
  if outstanding_reward  then
    event.player.print('will try again in 30 secs to give you your reward')
    outstanding_rewards[event.player.index]=outstanding_reward 
    timeout 30 secs
    give_reward(event.player, outstanding_reward, 'giving you the rest of your reward')
  end
end
```

mheguy: I would rather not handle those cases in the util itself because I feel like rewards can vary greatly. Some you might want the difference spilled, some you might want to try again later, some you might want to ignore.

linaori: Spilling loot at the player location would be an option

linaori:
@plague006 if you remove them, it kind of makes sense. You can get coins from someone and then have them removed

utils/player_rewards.lua
mheguy: Yes, I just would rather let each application choose their own method of handling it.

mheguy:
Everything except subtracting bits is in the last commit, implementing that will come in the next commit.

Also added an admin command to manually give rewards (added with the reward for blank map text in mind)

mheguy:
Should be good to go.

features/gui/info.lua
mheguy: There is now a check in place such that when a player has gained all possible coins their entry is removed from the table.

PR #548, opened 2018-12-14
added command deprecated message

Valansch:


features/chat_triggers.lua
```diff
@@ -131,6 +131,15 @@ local function mentions(event)
     end
 end
 
+--warns admins that /c /command and /silent-command are deprecated
+local function notify_deprecated_command(event)
+    local command = event.command
+    if command == 'silent-command' or command == 'c' or command == 'command' then
```
linaori, line 137: What about a table to make this more generic?
```lua
local deprecated_commands = {
    ['silent-command'] = 'sc',
    ['c'] = 'sc',
    ['command'] = 'sc',
}
```

linaori (approved):

grilledham:
Why does `/sc` deprecate `/c` or `/command` they serve different purposes?

Valansch:
sc's only purpose is to deprecate /c and /command.
We dont want errors go to all players.

grilledham:
I understand `/sc` replacing`/silent-command` but I think it follows that it replaces `/c` or `/command`. If you use `/c ` it's because you want people to see that you are using the command, so wouldn't you want people to also see the errors?

For example you might use `/c game.speed = 0.8` so players know you have changed the game speed. You wouldn't want to then receive a message saying it's deprecated.

Valansch:
Yea you are right. I rest my case :)

Valansch:
Thinking about it... I remember why i wanted to deprecate /c too.
I cannot disable /silent-command without disabling /c and /command too. And that was my plan...

PR #549, opened 2018-12-14
Implemented deprecated commands

Valansch:


mheguy (dismissed):
In testing as a player, this logged the message rather than printing.

mheguy:
Stylistically, we should probably stick to only single newlines to break up code blocks.

Valansch:
@plague006 
>In testing as a player, this logged the message rather than printing.
no? As a player this will print to chat. As server this will print to console and log it.

grilledham:
`game.player` is only accessible inside a command. You are using it inside an event.

Valansch:
My bad, you are right. I thought of on_console_command as a command 🙄

features/custom_commands.lua
```diff
@@ -11,7 +11,21 @@ local Command = require 'utils.command'
...
+local deprecated_command_alternatives = {
+    ['silent-command'] = 'sc'
+}
+
+Event.add(defines.events.on_console_command, function (event)
+        local alternative = deprecated_command_alternatives[event.command]
+        if alternative then
+            local print = log
+            if event.player_index then
+                print = game.players[event.player_index].print
```
grilledham: We normally use `Game.get_player_by_index` when getting player by index

Valansch: Fixed

features/custom_commands.lua
```diff
@@ -11,7 +11,21 @@ local Command = require 'utils.command'
...
+}
+
+Event.add(defines.events.on_console_command, function (event)
+        local alternative = deprecated_command_alternatives[event.command]
+        if alternative then
+            local print = log
+            if event.player_index then
+                print = Game.get_player_by_index(event.player_index).print
+            end
+            print(string.format('Warning! Usage of the command /"%s" is deprecated. Please use "%s" instead.', event.command, alternative))
```
mheguy: `command /"%s" is` should probably be `command "/%s" is`, it looks weird to have the slash outside of the quotes. `use "%s" instead` should probably be `use "/%s" instead` to match the style of the previous.

mheguy (approved):
Looks good to me.

PR #553, opened 2018-12-16
allow player to toggle top buttons visibility

### LINKED ISSUE #383: Allow our GUI buttons to collapse
Our friends at explosive gaming have a feature I really want: a button that collapses all their GUI buttons. I think it's an amazing idea and would love if we had it as well.

Their code is under the GPL license so we can just take inspiration or if we want to attribute it back to them we can take as much as we need: https://github.com/explosivegaming/scenario/blob/release/4.0.0/modules/ExpGamingCore/Gui/src/toolbar.lua

image

grilledham:
image

image

image

How to use in code:
1. Let Gui know that you want the button to be toggle-able by the player 
```lua
Gui.allow_player_to_toggle_top_element_visibility(element_name)
```
That is all.

If you want to register a handler for when the button is toggled you can use
```lua
Gui.on_player_show_top(element_name, handler)
Gui.on_pre_player_hide_top(element_name, handler)
```

I think I've registered all the top buttons.

Elements that are opened by the top buttons will remain open when the player hides the top buttons, like so

image

We could use the callbacks above to control that.

Closes #383

### DIFF
--- a/features/gui/blueprint_helper.lua
+++ b/features/gui/blueprint_helper.lua
@@ -572,4 +572,6 @@ Gui.on_custom_close(
     end
 )
 
+Gui.allow_player_to_toggle_top_element_visibility(main_button_name)
+
 Event.add(defines.events.on_player_joined_game, player_joined)
--- a/features/gui/info.lua
+++ b/features/gui/info.lua
@@ -675,6 +675,8 @@ Gui.on_custom_close(
     end
 )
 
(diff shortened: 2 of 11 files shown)


Valansch (approved):
Thank you.
Finally i can read debug info again! :D
Code is good as usual. Your comment game is top notch :D

mheguy (approved):
It's always funny to me when I mention something in admin chat and a few hours later you have a PR for a great change. :)

Happily merging this.

PR #554, opened 2018-12-16
Diggy fixes from last run

linaori:
Some fixes done based on last run


mheguy (approved):
Looks reasonable to me

PR #555, opened 2018-12-16
Create a new surface for RedMew

mheguy:
This introduces redmew_surface (as always my names are terrible and I would love a name suggestion)

Maps are no longer generated on nauvis. The biggest benefit is that it gives us control in the code over  "New Game" parameters like map_gen, difficulty, and map_settings.

We no longer have to set anything when creating a world from the scenario, just hit "generate". Further, this means that maps requiring specific map_gen can be launched from the server without issue. Each map can explicitly define the type of world they want to start with. All options can be set via the Public table.

Nightfall and terraforming are untested, but both due for more tweaks and are on my todo list.

Dimensions received some special attention and seems to work as intended.


map_gen/combined/borg_planet.lua
```diff
@@ -89,7 +89,7 @@ local function find_tile_placement_spot_around_target_position(tilename, positio
     local x = position.x
     local y = position.y
     if not surface then
-        surface = game.surfaces[1]
+        surface = RS.get_surface()
```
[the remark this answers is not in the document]
mheguy: Not my monkeys, not my circus.

map_gen/misc/rusky_pvp.lua
```diff
@@ -137,15 +138,15 @@ Event.add(
 
 function neForceNear(pos)
     for k, v in pairs(game.forces) do
-        if dist(pos, v.get_spawn_position(game.surfaces[1])) <= 50 then
+        if dist(pos, v.get_spawn_position(RS.get_surface())) <= 50 then
```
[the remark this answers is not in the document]
mheguy: Not my monkeys, not my circus.

map_gen/shared/ent_functions.lua
```diff
@@ -213,9 +216,8 @@ function auto_place_entity_around_target(entity, scan_radius, mode, density, sur
     return false
 end
 
-function create_entitie_cluster(name, pos, amount)
-
-    local surface = game.surfaces[1]
+Public.create_entitie_cluster = function(name, pos, amount)
+    local surface = RS.get_surface()
```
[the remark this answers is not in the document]
mheguy: Not my monkeys, not my circus.

map_gen/shared/ent_functions.lua
```diff
@@ -288,10 +290,10 @@ function create_tree_cluster(pos, amount)
     return b, e
 end
 
-function find_tile_placement_spot_around_target_position(tilename, position, mode, density)
+Public.find_tile_placement_spot_around_target_position = function(tilename, position, mode, density)
     local x = position.x
     local y = position.y
-    if not surface then surface = game.surfaces[1] end
+    if not surface then surface = RS.get_surface() end
```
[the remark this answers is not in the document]
mheguy: Not my monkeys, not my circus.

map_gen/shared/ent_functions.lua
```diff
@@ -514,15 +516,15 @@ function find_tile_placement_spot_around_target_position(tilename, position, mod
     return false
 end
 
-function create_tile_cluster(tilename,position,amount)
+Public.create_tile_cluster = function(tilename,position,amount)
     local mode = "ball"
     local cluster_tiles = {}
-    local surface = game.surfaces[1]
+    local surface = RS.get_surface()
```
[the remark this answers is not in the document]
mheguy: Not my monkeys, not my circus.

map_gen/shared/ent_functions.lua
```diff
@@ -514,15 +516,15 @@ function find_tile_placement_spot_around_target_position(tilename, position, mod
...
     local pos = position
     local x = pos.x
     local y = pos.y
     for i = 1, amount, 1 do
-        local b,x,y = find_tile_placement_spot_around_target_position(tilename, pos, mode)
+        local b,x,y = Public.find_tile_placement_spot_around_target_position(tilename, pos, mode)
```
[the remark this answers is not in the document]
mheguy: Not my monkeys, not my circus.

mheguy:
All current lint errors are unrelated to the changes I've made.

Valansch:
I very much like these changes. Could we brainstorm a better name?

mheguy:
Yes please. Naming is not my forté.

Valansch:
- main 
- map
- alpha
- redmew
- earth
- grilledham
- azeroth
- draenor

mheguy:
Oh the surface name itself? I meant the filename sucks. ~~Sounds like we need `resources.surface_names` and to randomize it every game. Nothing could potentially go wrong by doing it that way. 😄~~

Turns out you can't call math.random outside of an event or during loading.

map_gen/shared/redmew_surface.lua
```diff
@@ -0,0 +1,293 @@
...
+        first_player_position_check_override = first_player_position_check_override,
+    },
+    function(tbl)
+        first_player_position_check_override = tbl.first_player_position_check_override
+    end
+)
+
+-- These settings are default except:
+-- cliffs disabled, enemy bases to high frequency and big size, and starting area to small
+Public.map_gen_settings = {
```
linaori: What if we move this to `resources/surfaces/redmew.lua`? Meaning we can easily set/swap/store presets

mheguy:
Bring on the stickler.. (FP to resolve conflicts)

mheguy:
I'll resolve the conflict when this is ready to merge.

features/antigrief.lua
```diff
@@ -173,7 +173,7 @@ Module.undo =
         if e.last_user == player then
             --Place removed entity IF no collision is detected
             local last_user = global.original_last_users_by_ent_pos[get_position_str(e.position)]
-            local new_entity = place_entity_on_surface(e, game.surfaces.nauvis, false, last_user)
+            local new_entity = place_entity_on_surface(e, game.surfaces.redmew, false, last_user)
```
linaori: Shouldn't this be `RS.get_surface()`?

mheguy: Yep, missed this one. Ty.

mheguy:
Another round stickler, let's go.

map_gen/shared/redmew_surface.lua
mheguy: Love the idea. Added.

mheguy:
I caved, I just couldn't deal with stickler's squawking.

map_gen/shared/redmew_surface.lua
```diff
@@ -0,0 +1,122 @@
...
+local Event = require 'utils.event'
+local Game = require 'utils.game'
+local Global = require 'utils.global'
+local map_gen_settings_presets = require 'resources.map_gen_settings'
+local map_settings_presets = require 'resources.map_settings'
+local difficulty_settings_presets = require 'resources.difficulty_settings'
+
+local Public = {}
+Public.surface_name = 'redmew'
+Public.first_player_position_check_override = false
```
grilledham: Do any other modules use `Public.first_player_position_check_override`?
I raise this because this doesn't do what you probably think it does. `Public.first_player_position_check_override` will stay false, as you never set it to anything else.

Valansch: Even if this is intended to be a read only public variable, it should probably be behind a getter. This suggest that the variable can be written, which will cause desyncs.

map_gen/shared/redmew_surface.lua
```diff
@@ -0,0 +1,122 @@
...
+                {name = 'lab-white', position = {0, -1}},
+                {name = 'lab-white', position = {0, 0}}
+            }
+        )
+        player.teleport({0, 0}, surface)
+        spawn_coords = {0, 0}
+        first_player_position_check_override[1] = false
+    end
+    game.forces.player.set_spawn_position(spawn_coords, surface)
+end
```
grilledham, line 98: This function confuses me. Setting a starting island in case there isn't one is a nice touch. But do we really have to find a non_colliding_position? Doesn't the game do that automatically.

Also you always set the force spawn position every time a new player is created, which doesn't seem necessary.

map_gen/shared/redmew_surface.lua
mheguy: The specific use-case is diggy. It needs to check for the first player in. But you're right this should be a getter and setter function. Will make that change.

grilledham:
So my understanding is the redmew surface either uses the user settings or the map preset settings. 

It would be a bit of work, but I think it'd be better if the redmew surface used both, with the map preset settings overrides the user settings. 

That way a map could say that water most be turned off, but the user can still change the ore settings.

map_gen/shared/redmew_surface.lua
mheguy: You're right, will move the spawn position setting to the `create_redmew_surface` function.

As for the weird teleport, it's a diggy thing. Without it, I often ended up spawning inside rocks and dying.

mheguy:
Actually it completely ignores the user settings. The rational is that "the map is the map".

By default it uses settings from the resources files.
Edit:
But you can also enable/disable `map_gen_settings` `map_settings` and `difficulty` via the config.
That was my easy route to compromise when I wanted users to have *some* control over things.

grilledham:
Ignoring user settings is a non go for some presets that rely on user settings for ores or bitter settings.

mheguy:
Those presets could write to the `global.config.redmew_surface` table to disable `redmew_surface` taking control of those aspects of worldgen. Would that work for you?

grilledham:
In principle if we are changing the default behaviour of the game, it should be opt in not opt out.

mheguy:
> In principle if we are changing the default behaviour of the game, it should be opt in not opt out.

I think that depends how you see redmew.

The 1.1.0 release, released roughly a month ago has 260 downloads across the 3 files available. `Silly Names for the Holidays` which is a map that was just run 11 days ago has 145 players in 5 1/2 hours.

That to say, I think our time/effort should be on internal development. Making things as easy as possible for map makers to get their maps exactly how they want them.

Further, I think if users are downloading our scenario they don't expect default behaviour. The moment they start the game a lot is different from vanilla and they have a big info box telling them know about some of the differences. They have coins in their hotbar, and a bunch of GUI buttons in the top left. All of which are opt-out. If they're playing with their friends `nuke_control` might kick or ban them. If they open the player list and click the little flag icon they could well jail themselves and have no idea how to undo it. So I think as it stands RedMew can be hostile to an end-user. And for a time I tried to advocate for end-users but the numbers just don't support it.

But I do agree with you in that: if someone takes the time to download the scenario and they don't want to start a map then we shouldn't force any of `map_gen` down their throats. Something like #516 would change that to where loading (the equivalent of) `map_layout` is opt-in, not opt-out. Of the 36 files that require `redmew_surface` only `antigrief` is outside of `map_gen` and it presently is not loaded. So if we were to split the base scenario and map_gen properly, then redmew_surface can be made opt-in along with loading one of our maps.

As far as each map opting in or out, when we redo map loading (again, something like #516) it would be nice to go through existing maps to see which are actually functional, and which need to opt in or out of custom map gen settings.

mheguy:
Have big changes to make, as a new way of doing things came to me.

mheguy:
Will recreate PR when I'm done rewriting.

PR #558, opened 2018-12-17
Add dump_env

mheguy:
When `_DEBUG` is turned on, it dumps `_ENV` into a file in the tick that the game is loaded. Nice for catching any potential diffs that can get rid of themselves over time.

In terms of workflow: open the desync report and extract the ref and desync save files, open both to edit config.lua to _DEBUG = true. Load save 1, rename dump_env.lua. Load save 2 and you can compare the two output files.

How this can be improved: automatically stripping out strings matching `<.*>` since they differ between files.

### DIFF
--- a/features/player_create.lua
+++ b/features/player_create.lua
@@ -48,7 +48,7 @@ local function player_created(event)
         p(get_random_weighted(random_messages))
     end
 
-    if config.show_info_at_start and not _DUMP_ENV then
+    if config.show_info_at_start then
         if Info ~= nil then
             Info.show_info({player = player})
         end
--- a/features/player_create.lua
+++ b/features/player_create.lua
@@ -48,7 +48,7 @@ local function player_created(event)
         p(get_random_weighted(random_messages))
     end
 
-    if config.show_info_at_start then
+    if config.show_info_at_start and not _DUMP_ENV then
         if Info ~= nil then
             Info.show_info({player = player})
         end
--- a/.luacheckrc
+++ b/.luacheckrc
@@ -56,7 +56,7 @@ local STD_BASE_CONTROL = 'lua52c+factorio+factorio_control+factorio_defines+fact
 --[Assume Factorio Control stage as default]--
 -------------------------------------------------------------------------------
 std = STD_CONTROL
-globals = {'print', 'math', 'table', '_DEBUG', '_CHEATS', 'ServerCommands'} -- RedMew-specific globals
+globals = {'print', 'math', 'table', '_DEBUG', '_CHEATS', '_DUMP_ENV', 'ServerCommands'} -- RedMew-specific globals
 max_line_length = LINE_LENGTH
 
 not_globals = NOT_GLOBALS
--- a/config.lua
+++ b/config.lua
@@ -1,5 +1,6 @@
 _DEBUG = false
 _CHEATS = false
+_DUMP_ENV = false
 local market_item = 'coin'
 
 global.config = {
--- a/control.lua
+++ b/control.lua
@@ -91,6 +91,6 @@ end
 if config.camera.enabled then
     require 'features.gui.camera'
 end
-if _DEBUG then
+if _DUMP_ENV then
     require 'utils.dump_env'
 end
--- a/control.lua
+++ b/control.lua
@@ -91,3 +91,6 @@ end
 if config.camera.enabled then
     require 'features.gui.camera'
 end
+if _DEBUG then
+    require 'utils.dump_env'
+end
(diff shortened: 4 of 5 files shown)


linaori (approved):
Nice!

grilledham:
This is a nice idea.

A few points come to mind.

 - I don't think we should use `_DEBUG` for this. You might want to use `_DEBUG` for online test runs without dumping files every time someone joins. I suggest we make a new variable called `_DUMP_ON_JOIN` or something.
 - How does `inspect` differ from `serpent`. More just curious, it seems to work well.
 - When dumping files we probably want to ignore some tables. Normally I don't dump the content of the `package` table as that can be quite big.

mheguy:
You're right about not using debug, I debated it but honestly was shy about introducing a new global.

I don't know exactly how inspect and serpent differ. `serpent.dump` does a fine job but isn't human-readable (returns a block of text with no line breaks). `serpent.block` errors out so I used `inspect`.

Serpent: Lua serializer and pretty-printer that implements compact and robust serialization that still produces a readable code.

Inspect: Not a serialization library - its output is not valid Lua code, but it's extremely human-readable. It's supposed to be used for debugging. Keys are sorted out alphabetically, nested tables are properly indented, table loops are detected and signaled, and metatable information is also included.

mheguy:
For ignoring some tables, I'll have to implement that in another commit.

features/player_create.lua
```diff
@@ -48,7 +48,7 @@ local function player_created(event)
         p(get_random_weighted(random_messages))
     end
 
-    if config.show_info_at_start then
+    if config.show_info_at_start and not _DUMP_ENV then
```
mheguy: Without this check, the game (local or on server) will hard crash after a few seconds and I have no clue why.

It will only happen with the changes made to dump_env, if we don't process the table there is no crash. In addition, I was unable to reproduce the crash by opening the info window manually.

I had a hard time convincing myself this was the cause of the crash, but after a bunch of testing I'm fairly convinced this is the only source of the issue. The conclusive test for me was:

Testing without the additional check (so only `if config.show_info_at_start then`):
Set `config.show_info_at_start` to `false` in config.lua. Launch on server. Join server and hang out for a minute or two and observed no crash. Changed `player.name` and set `global.config.player_create.show_info_at_start = true`. Disconnect and rejoin (because of the name change, the server recognizes you as a new player). Upon rejoin, the server crashes within seconds.

I don't know if we want to leave this weird interaction in the code or if we want to revert this commit.

mheguy:
Had a look at a couple desync reports: I'll leave the filter in if you feel strongly about it, but package only seems to be ~1000 lines. The reports I'm looking at are about 63,000 lines long.

Edit: Seems like about 40,000 of those lines are terraforming and nightfall each recording every chunk generated 😁

mheguy:
I created a branch to experiment on. I've used it to narrow down the possible interactions causing the issue. At this point I've run up against the limit of what I can knowledgeably remove.

Everything remains as described above: in dump_env if line 13 is
```
local dump_string = inspect(_ENV, {process = remove_package})
```
it will crash and if it's
```
local dump_string = inspect(_ENV)
```
it will not.

The branch: https://github.com/Refactorio/RedMew/tree/inspect_filter_and_info_gui_crash
Info on inspect: https://github.com/kikito/inspect.lua

In the course of testing, a crash file was generated once: [crash_report.zip](https://github.com/Refactorio/RedMew/files/2704402/crash_report.zip)

plague006:
Makes sense I guess, but you can't pass a factorio Lua metatable or its elements or whatever through a function and not expect factorio to be upset.

mheguy:
Makes sense I guess, but you can't pass a factorio LuaObject through a function and not expect factorio to be upset.

This is now stable, filters metatables, and `package`. Should be good to go.

Valansch:
Thank you for your work and time. Hope you can finally sleep :D

Valansch:
Do you want to make a bug report? The crash does not make sense. A crash is never expected behavior, even if the Microsoft bluescreen team might disagree.

mheguy:
I did: https://forums.factorio.com/viewtopic.php?f=218&t=64019&p=390678#p390678

RSeding replied saying they'll remove inspect from 0.17 and he's not quite sure why it happened.

Valansch:
Welp, so this deprecates this.

mheguy:
Not really, we can just include inspect in utils

mheguy:
`serpent.block` has been unable to cope with some dumps in the past.
image
`serpent.dump` works but only spits to a single line.
So I'm not sure if we have other options.

PR #559, opened 2018-12-18
Fix issues with silly names

mheguy:
The 2 major issues were: a new LuaPlayer being created when someone rejoined and incompatibility with user groups.

This solves those by storing the player's original name and reverting to that when they leave, and by creating a second global table for regulars. I is the lowest overhead implementation I could think of. When silly_names is not loaded, it will add 1 check for `global.silly_regulars` to user_groups::is_regular for guests.

tweaks info.lua to allow extending add_map_extra_info rather than outright overwriting it, and tweaks table.clear_table to give a faster option for sorted tables.

Edit: This also changes the naming scheme since @Jayefuu strongly urged to have the player's real name included. In the future I would like to make the naming scheme configurable.


Valansch (approved):
Looks good.

mheguy:
Force-push after Val's review to resolve merge conflict

PR #561, opened 2018-12-18
Add regulars to potential train station names

mheguy:
Also.. we now increment the counter in the non-sorted `get_random` path.

### DIFF
--- a/features/train_station_names.lua
+++ b/features/train_station_names.lua
@@ -2,10 +2,10 @@ local Event = require 'utils.event'
 
 local function pick_name(event)
     -- Create a weight table comprised of the backer name, a player's name, and a regular's name
-    local player = table.get_random(game.players, true)
+    local random_player = table.get_random(game.players, true)
     local name_table = {
         {event.created_entity.backer_name, 8},
-        {player.name, 1},
+        {random_player.name, 1},
         {table.get_random(global.regulars, false, true), 1},
     }
     return table.get_random_weighted(name_table)
--- a/features/train_station_names.lua
+++ b/features/train_station_names.lua
@@ -3,6 +3,9 @@ local Event = require 'utils.event'
 local function pick_name(event)
     -- Create a weight table comprised of the backer name, a player's name, and a regular's name
     local random_player = table.get_random(game.players, true)
+    if not random_player then
+        return
+    end
     local name_table = {
         {event.created_entity.backer_name, 8},
         {random_player.name, 1},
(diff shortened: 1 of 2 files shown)


mheguy:
I really enjoy when people are surprised to see a name they recognize on stations, so let's make the odds a bunch lower. :)

features/train_station_names.lua
```diff
@@ -1,17 +1,24 @@
 local Event = require 'utils.event'
 
+local function pick_name(event)
+    -- Create a weight table comprised of the backer name, a player's name, and a regular's name
+    local random_player = table.get_random(game.players, true)
```
grilledham, line 5: Really minor point, but technically game.players isn't guaranteed to be a 'sorted'* table. There could be gaps in the indexes if we every use `game.remove_offline_players` which would mean `random_player` could be `nil`.

As that is extremely rare, I'd suggest just returning if `random_player` is nil

*by sorted I mean what `table.get_random` considers to be sorted.

utils/table.lua
```diff
@@ -73,28 +73,34 @@ table.set = function(t, index, element)
...
 --- Chooses a random entry from a table
---@param t table to select an element from
---@param sorted boolean to indicate whether the table is sorted by numerical index or not
---@return a random element of table t
-table.get_random = function(t, sorted)
+-- @param t table to select an element from
+-- @param sorted boolean to indicate whether the table is sorted by numerical index or not
+-- @param key boolean to indicate whether to return the key or value
+-- @return a random element of table t
+table.get_random = function(t, sorted, key)
```
grilledham, line 81: I know this PR doesn't change this. But using `math.random` inside this function means the function can't be used outside of events. That should probably be mentioned in the comments.

utils/table.lua
```diff
@@ -73,28 +73,34 @@ table.set = function(t, index, element)
...
--- @param item_index the index of the items, defaults to 1
--- @param weight_index the index of the weights, defaults to 2
--- @returns a random item with weighting
--- @see features.chat_triggers.hodor
+-- @param weight_table table of tables with items and their weights
+-- @param item_index number of the index of items, defaults to 1
+-- @param weight_index number of the index of the weights, defaults to 2
+-- @returns a table entry
+-- @see features.chat_triggers::hodor
 table.get_random_weighted = function(weighted_table, item_index, weight_index)
```
grilledham, line 106: same as above

features/train_station_names.lua
mheguy: Makes sense to me. And yes 'sorted' could use a better name.

utils/table.lua
mheguy: Added

utils/table.lua
mheguy: same as above

grilledham (approved):

PR #563, opened 2018-12-20
Fixed possible gui uid collision

Valansch:
The gui uid was of the form:
```lua
filename .. ',' .. x
```

This change makes it:

```
filepath .. ',' .. x
```

Making collisions of different guis from different files with the same name impossible


grilledham (approved):
LGTM 👍

PR #565, opened 2018-12-21
Event

grilledham:
This PR adds 3 things
 - Extra methods for registering function, that can be registered at runtime and removed without causing desyncs.
For example this is now possible:
```lua
/c
Event.add_removable_nth_tick_function(
    60,
    function(event)
        game.print(event.tick)
    end
)
```
- Safety checks to stop the register methods being called at times when they would cause deyncs.
Using Event.on_nth_tick at console.
image
Using a closure in Event.add_removable_nth_tick_function.
image
image

- Documentation for how to use events.

I thought about adding type checks, but thought it would be unlikely that someone wouldn't pass a function to the register methods. With the exception of the multiple add_removable methods where some take a token and some take a function, so those have type checks.

### DIFF
--- a/utils/event.lua
+++ b/utils/event.lua
@@ -4,7 +4,7 @@
 --
 -- Handlers added with Event.add must be added at the control stage or in Event.on_init or Event.on_load.
 -- Remeber that for each player, on_init or on_load is run, never both. So if you can't add the handler in the
--- contorl stage add the handler in both on_init and on_load.
+-- control stage add the handler in both on_init and on_load.
 -- Handlers added with Event.add cannot be removed.
 -- For handlers that need to be removed or added at runtime use Event.add_removable.
 -- @usage
@@ -20,7 +20,7 @@
 --
 -- For conditional event handlers. Event.add_removable can be safely called at runtime without desync risk.
 -- Only use this if you need to add the handler at runtime or need to remove the handler, other wise use Event.add
--- Token is used becasue it's a desync risk to store closures inside the global table.
+-- Token is used because it's a desync risk to store closures inside the global table.
 --
 -- @usage
 -- local Token = require 'utils.token'
@@ -52,7 +52,7 @@
 -- @usage
 -- local Event = require 'utils.event'
 --
--- If you want to remove the handler you will need to keep a refernece to it.
+-- If you want to remove the handler you will need to keep a reference to it.
 -- global.handler = function(event)
 --     game.print(serpent.block(event)) -- prints the content of the event table to console.
 -- end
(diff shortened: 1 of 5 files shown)


utils/event.lua
```diff
@@ -1,152 +1,418 @@
+--- This Module allows for registering multiple handlers to the same event, overcoming the limitation of script.register.
+--
+-- ** Event.add(event_name, handler) **
+--
+-- Handlers added with Event.add must be added at the control stage or in Event.on_init or Event.on_load.
+-- Remeber that for each player, on_init or on_load is run, never both. So if you can't add the handler in the
```
linaori: typo: Remeber -> remember

utils/event.lua
```diff
@@ -1,152 +1,418 @@
...
+-- contorl stage add the handler in both on_init and on_load.
```
linaori: typo: contorl -> control

utils/event.lua
```diff
@@ -1,152 +1,418 @@
...
+-- ** Event.add_removable_function(event_name, func) **
+--
+-- Only use this function if you can't use Event.add_removable. i.e you are registering the handler at the console.
+-- func cannot be a closure in this case, as there is no safe way to store closures in the global table.
+-- A closure is a function that uses a local variable not defined in the function.
+--
+-- @usage
+-- local Event = require 'utils.event'
+--
+-- If you want to remove the handler you will need to keep a refernece to it.
```
linaori: typo: refernece -> reference

utils/event.lua
```diff
@@ -1,152 +1,418 @@
...
+-- Use Event.on_nth_tick(tick, handler) for script.on_nth_tick(tick, handler)
+-- Favour this event over Event.add(defines.events.on_tick, handler)
+-- There are also Event.add_removable_nth_tick(tick, token) and Event.add_removable_nth_tick_function(tick, func)
+-- That work the same as above.
+--
+-- ** Custom Scenario Events **
+--
+-- local Event = require 'utils.event'
+--
+-- local event_id = script.get_event_handler()
```
linaori: Shouldn't this be: `script.generate_event_name()`?

linaori (commented):
What's the difference between EventCore and Event? which should we use when?

grilledham:
I'll fix the typos later today. I should really get a word check. 

event_core only exists to break a circular dependency between global.lua and event.lua.
I don't expect anyone to use event_core.
There is a comment at the top of the file saying that.

mheguy:
Typos corrected

mheguy:
I was able to test a bit of this. The only issue I ran into was addressed by 0a1a431

mheguy:
I love the potential to disable events during runtime. This lets us enable/disable tons of modules/features during runtime.

As a sample: https://github.com/plague006/RedMew/blob/91c39af47059a2c6417987a67a213f8212a583cd/features/redmew_qol.lua

linaori (approved):

PR #566, opened 2018-12-21
Implement server-side color saving

mheguy:
When joining server, sets your saved color.
Adds a couple commands for setting and resetting your color server-side. Also enables getting a random color.

Unrelated changes:
If you're running with `_DEBUG` and you're an admin, gain regular status so that you can test commands that require regular status. -_-
Extend `command` to make commands regular_only.
Add `set_and_return` to core utils.


utils/command.lua
```diff
@@ -62,6 +65,7 @@ function Command.add(command_name, options, callback)
     local description = options.description or '[Undocumented command]'
     local arguments = options.arguments or {}
     local default_values = options.default_values or {}
+    local regular_only = options.regular_only or false
     local admin_only = options.admin_only or false
```
linaori, line 69: What if we change this to "minimum permission level required" of sorts and default it to guest? That would solve the problem and we only have to check if your permissions aren't too low. Might have to change the info added to something like "(Regular or higher)" and "(Admin only)" and not showing anything if everyone (regardless of server) may execute it.

linaori (dismissed):

grilledham:
The first time a player joins the welcome message might not have the correct color. We could wait 60 ticks before showing the welcome message to give time for the color to be fetched from the server. Or we print welcome messages in white.
https://github.com/Refactorio/RedMew/blob/d4e7849cd20ef9342f09b231ae39f1110d3b05e1/features/donator_messages.lua#L24

linaori:
I don't mind a delayed message

mheguy:
Right right, because retrieving from the server isn't same-tick. Yeah a delay would be totally cool in my book.

utils/command.lua
mheguy: That would seem to contradict https://github.com/Refactorio/RedMew/pull/506 where it was established that each rank should be modular and not hierarchical.

mheguy:
No luacheck on mac laptop so stickler is getting a workout. :D

utils/command.lua
```diff
@@ -12,6 +13,7 @@ local option_names = {
     ['description'] = 'A description of the command',
     ['arguments'] = 'A table of arguments, example: {"foo", "bar"} would map the first 2 arguments to foo and bar',
     ['default_values'] = 'A default value for a given argument when omitted, example: {bar = false}',
+    ['regular_only'] = 'Set this to true if only regulars may execute this command',
```
Valansch, line 16: I would change lines 16 and 17 to:

['requires_regular'] = 'Set this to true if only players with rank regulars may execute this command'
['requires_admin'] = 'Set this to true if only players with rank admin may execute this command'

Equally at all occurrences. This way we don't violate #506. The word "only" is a bit confusing here. I know what you mean by "regular" and "non-regular", but i don't think its 100% clear that and admin for example is also always a regular and that someone with donator and regular rank is indeed also a regular.

Valansch (changes requested):

utils/command.lua
mheguy: https://github.com/Refactorio/RedMew/issues/610 aims to do just that. But because #576 and #566 require the same changes it would be much easier for me to work on `command` once those two PRs are merged. Or to put it a different way: if I try to refactor `plague:color`, `plague:overhaul_commands`, and `refactorio:dev` at the same time I'm quite certain I'll make a mess of one, if not all, of them.

Valansch: Well good then.

Valansch (approved):

PR #571, opened 2018-12-22
Issue: Using inspect with a filter causes something in the GUI to crash the game.

mheguy:
So this is technically an issue that I'm opening as a PR to help keep track of changes and states that have been tested.
This issue first arose in  #558 and I will summarize below:
If we use `inspect` to dump the contents of `_ENV`, the game will crash when you close the info GUI.

Testing has revealed: it's not specific to `_ENV`. It will crash if dumping `global` as well. I tested with an empty table `local test_table = {}` and it did not crash. Further, it has nothing to do with closing the GUI per se, as you can replicate the crash in other ways.

The contents of the branch in this PR is the most I could reduce the codebase while
1) still observing the crash
2) not touching functional code that I don't understand at all

Presently:
You can open the scenario, but when you try to quit or reload the game *should* crash to desktop. Typically not even generating a log in the process.

I've reached the end of what I can honestly understand, so I'm hoping someone can finish this relay race and take it over the finish line. :)

This PR is into the live_testing branch, so if someone hits the merge button by mistake it won't cause any harm.

Edit: ~~If I recall from my testing, at one point I tried not writing to a file and that resulted in no crashing, though I would have to re-test.~~ It crashes just fine whether you write to a file or not.


plague006:
You can't process metatables. That's it.

mheguy:
You can't process metatables. That's it.

PR #574, opened 2018-12-22
Fix not being able to switch targets without first closing camera

mheguy:
Previously having a camera open and trying to open another one would cause an error and bug the camera out.

Also breaks `create_destroy_camera` into more single-purpose functions.

Would love one of the GUI gurus to swoop in and have this use the redmew GUI instead of factorio's `mod_gui`. :D

### DIFF
--- a/features/gui/camera.lua
+++ b/features/gui/camera.lua
@@ -1,12 +1,13 @@
 --[[
-Camera, used under MIT license.
-Copyright 2018 angelickite
+    Liberally rewritten from:
+    Camera, used under MIT license.
+    Copyright 2018 angelickite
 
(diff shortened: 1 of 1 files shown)


features/gui/camera.lua
```diff
@@ -167,22 +189,14 @@ local function on_tick()
...
+Command.add(
+    'watch',
+    {
+        description = 'Allows you to watch other players.',
+        arguments = {'target'},
+        default_values = {target = false}
+    },
+    camera_command
+)
 Event.on_nth_tick(120, on_tick)
```
Valansch, line 201: We should really keep a list of all our on nth tick handlers. We have too many 60, 120, 180 tick handlers. Those all collide. But since this is a problem in the entire project this instance here wont make a difference.

Valansch (approved):
Thank you. All great.

features/gui/camera.lua
mheguy: You're right, and you're right. Open an issue and I think we can figure something out to cover all "per_second" events.

Valansch: Event.every_n_ticks would be nice. It wouldn't fire it on the nth tick, but instead choose an empty slot.

PR #576, opened 2018-12-22
Overhaul commands

mheguy:
Put (nearly all) commands inside Lin's awesome command wrapper.


Valansch (approved):
Looking good to me. Very nice cleanup 👍

PR #577, opened 2018-12-22
Added a market event to do non-item related stuff

linaori:
Example usage:
```lua 
{
    price = 1000, 
    name = 'buff-biters', 
    description = 'Makes enemy biters stronger\n+10% attack\n+10% armor', 
    name_label = 'Buff enemy Biters', 
    type = 'buff', 
    sprite = 'entity/medium-biter'
}
Event.add(Retailer.events.on_market_purchase, function (event)
    if event.item.type ~= 'buff' then
        return
    end
    event.player.print(format('Purchased %dx %s', event.count, event.item.name_label))
end)
```
image

image

### DIFF
--- a/features/retailer.lua
+++ b/features/retailer.lua
@@ -130,8 +130,9 @@ local function redraw_market_items(data)
         local label = grid.add({type = 'label', caption = message})
         local label_style = label.style
         label_style.width = 93
-        label_style.height = 38
+        label_style.height = 32
         label_style.font = 'default-bold'
+        label_style.vertical_align = 'center'
 
         if is_missing_coins then
             label_style.font_color = color_red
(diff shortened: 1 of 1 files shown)


features/retailer.lua
```diff
@@ -18,10 +18,22 @@ local market_frame_close_button_name = Gui.uid_name()
 local item_button_name = Gui.uid_name()
 local count_slider_name = Gui.uid_name()
 local count_text_name = Gui.uid_name()
+local raise_event = script.raise_event
```
mheguy: Can I bug you to define `raise_event` with the other function definitions just a few lines up?

linaori:
All fixed. One thing I was curious about, perhaps granting the item should also go to an event listener, but not sure. It's now an if within the button pressed

linaori:
Made the text align vertically centered
image

Valansch:
Looks good :+1:

PR #579, opened 2018-12-22
Uses the retailer in crash site for markets

linaori:
- Brings back support for float prices
- Makes crash site use the Retailer with the minor GUI improvements  that it brings

I've left the functionality status quo in the crash site, but in theory things like outpost names could be added to the market by calling `Retailer.set_market_group_label(market_id, 'Add the name here')`


map_gen/presets/crash_site/outpost_builder.lua
```diff
@@ -1082,13 +1082,14 @@ Public.market_set_items_callback =
         end
 
         entity.destructible = false
+        local market_id = 'market-' .. Token.uid()
+        Retailer.add_market(market_id, entity)
```
grilledham, line 1086: What is the market_id used for and does it have to be unique?

My concern is that `Token.uid()` will return different numbers for players if used after `on_init` or `on_load` has run.
So either it will desync, or it wont desync because the id isn't used for anything, in which case what is it needed for. 
So my assumption would be that it might be used for something in the future, in which case it should be made desync safe now, or that is something we have to remember to do at a later date.

map_gen/presets/crash_site/outpost_builder.lua
```diff
@@ -1082,13 +1082,14 @@ Public.market_set_items_callback =
...
 
         local p = entity.position
         local x, y = p.x, p.y
         local d = math.sqrt(x * x + y * y)
 
-        local market_data = {}
-        for i, item in ipairs(data) do
+        for _, item in pairs(data) do
```
grilledham: I know I was mistaken about ipairs being faster, but I like it to mark that data is an array and not a dictionary. Recently however, I've been using a standard for loop for arrays.

grilledham (commented):
Overall I think the changes to the market are very good 👍

map_gen/presets/crash_site/outpost_builder.lua
linaori: I can rewrite it as such, it seems that all indices here are sequential 👍

map_gen/presets/crash_site/outpost_builder.lua
linaori: The idea is to make sure it has a unique value. Meaning that when a new market is created, it won't have the same name as an existing market group. In Crash site all markets have their own set of items, thus need a unique group each, I was thinking that with this we could ensure every market has a unique group and thus their own item set. 

In the retailer this group id is used to identify the items connected to it. I need a value that can't cause collisions to prevent items from being shared between markets.

Could also make a `Retailer.generate_group_id()`

grilledham: I was going to recommend a `Retailer.generate_group_id()` function

mheguy (approved):
Looks good to me

PR #581, opened 2018-12-22
Remove spaces in extant names and modify name function to remove spaces to prevent future mistakes

mheguy:


grilledham (approved):

mheguy:
https://github.com/Refactorio/RedMew/pull/583 included in this now. In addition this puts silly_players into a sort of deprecated/semi-retired state. I plan on overhauling it (see https://github.com/Refactorio/RedMew/issues/586) but for now it's in a bad state where it's not able to accomplish what it's setting out to do.

PR #583, opened 2018-12-22
Remove regulars backwards compatibility

mheguy:
https://github.com/Refactorio/RedMew/pull/506#issuecomment-443796936


mheguy:
Will be addressed in a different PR

PR #588, opened 2018-12-23
Restore ability to set map_extra_info

mheguy:
Previously I turned `set_map_extra_info` into `add_map_extra_info` because I didn't want everything competing to set info when they really wanted to be adding info. I think my head was in the right place, that most things should be adding. But it was dumb of me to outright get rid of/forbid setting the info. There are absolutely use-cases for wanting to explicitly set the full text (ex. if diggy were incoporating some part of terraforming, it wouldn't want to talk about toxic atmosphere)

So, this recreates/restores `Public.set_map_extra_info`. It also makes its use explicitly clear: it grants exclusive control over `map_extra_info` and it is an error to call it more than once. 

While I'm at it, should I add locks on all the set functions?

Am I missing use-cases where we might want multiple calls to the public `set` functions?

### DIFF
--- a/features/gui/info.lua
+++ b/features/gui/info.lua
@@ -36,6 +36,7 @@ local map_extra_info_key = 3
 local new_info_key = 4
 
 local rewarded_players = {}
+local map_extra_info_lock = {false}
 
 local editable_info = {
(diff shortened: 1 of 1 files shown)


Valansch:
Great :D

PR #589, opened 2018-12-23
Improve overall ore balancing for tendrils_impure.lua

BlueRock2:
Mining primarily iron and copper veins late game still provides an overabundance of stone and coal.   Would like to change purity from around 82% currently to 86% for iron and copper veins.   Standardized ore weights for each cluster to add up to 1000 so that percentage of each ore = its weight/10.   Coal and stone veins effectively the same percentage since players can tap them or not.   Also kept ore percentages scattered fragments the same since the stone and coal in them is useful early game, but doesn't affect late game much since the scattered fragments are low yield overall and not worth tapping late game.

### DIFF
--- a/map_gen/Diggy/Orepattern/tendrils_impure.lua
+++ b/map_gen/Diggy/Orepattern/tendrils_impure.lua
@@ -20,8 +20,8 @@
 --   zero: does nothing with this source
 --   one: adds the weight directly to the noise calculation
 --
--- weights:  recommend having resource weights for each cluster add up to 100
---           so that the weights = percentage of each ore type
+-- weights:  recommend having resource weights for each cluster add up to 1000
+--           so that it is apparent that every 10 weight = 1%.  eg. weight 860 (86%) + weight 80 (8%) + weight 60 (6%) = 100%
 
 return {
     { -- tendril medium large impure iron
@@ -40,9 +40,9 @@ return {
             }
         },
         weights = {
-            ['iron-ore']    = 80,
-            ['coal']        = 10,
-            ['stone']       = 8,
+            ['iron-ore']    = 860,
+            ['coal']        = 80,
+            ['stone']       = 60,
         },
         distances = {
             ['coal']        = 16,
@@ -66,9 +66,9 @@ return {
             }
         },
         weights = {
-            ['copper-ore']  = 80,
-            ['coal']        = 10,
-            ['stone']       = 8,
+            ['copper-ore']  = 860,
+            ['coal']        = 80,
+            ['stone']       = 60,
         },
         distances = {
             ['coal']        = 16,
@@ -92,9 +92,9 @@ return {
             },
         },
         weights = {
-            ['coal']        = 80,
-            ['iron-ore']    = 16,
-            ['stone']       = 5,
+            ['coal']        = 790,
+            ['iron-ore']    = 160,
+            ['stone']       = 50,
         },
         distances = {
             ['coal']        = 16,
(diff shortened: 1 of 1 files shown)


mheguy:
Figured I'd drop this here since it's info I have that could potentially be worth knowing. The ratios of ores in vanilla mapgen: https://docs.google.com/spreadsheets/d/1I0PHmQyMlNjZ7t4tdJ_Mdgl9oM1ao6yG3FAY10goRts/edit#gid=1176048894

BlueRock2:
On my first submission for tendrils_impure ore generation, I did some testing by clearing void on a 500,000 square tile area several times and adding up the ore counts.   But it isn't very accurate because I was manually hovering over the tendrils and adding up approximate amounts.    Doing this however, tends not to reveal the amounts added by impurities in the veins.   I think I may know enough about LUA now to have it accurately add up all the ore of each type within a square area for me, but would like to get my hands on existing code if it is already available somewhere.

mheguy:
In your script-output folder you'll get `resource_dump.lua` containing a table with all resources.
```lua
game.write_file('resource_dump.lua', serpent.block(game.player.surface.get_resource_counts()))
```

BlueRock2:
Thanks!, I will do some resource counts on several map generations tonight, as I believe you are right that the Diggy total tendril ore percentages need to be close to default map generation.  Except for maybe stone, which because of constructing all the supports would need to be slightly higher.   May still have to go by feel on that one.

mheguy:
>  May still have to go by feel

I think that's true overall. I'm not offering vanilla numbers to say they should be more or less like them. In vanilla the vast majority of coal and stone patches go unused. It's just for comparison.

If you want a ratio to potentially aim for, I would recommend checking consumed items consumed in a completed diggy save.

BlueRock2:
Taking into account past consumption sounds like a really good idea.   The Factorio default ore percentages appear to be based on what you would use if you kept cranking out equal amounts of all science packs over time.  But currently most players spend the most time in game before then.   I will gather the consumption numbers, but I think they may have changed noticeably each time depending on alien difficulty (more iron for bullets) or how hard it is to get uranium (delays switch from coal to nuclear power).   I think the optimal target resources is probably somewhere between current average consumption per Diggy game and that of resources needed for infinite science.

BlueRock2:
Just for reference while I do further testing to verify ore averages for this PR:   Diggy consumption averages for the currently playing game (close to end game) and last three end game saves normalized with iron =1 is copper = 0.59,  coal = 0.41 and stone = 0.20.    Interestingly I found that the default ore values you gave me are right on the nose for iron/copper ratio for infinite science (1 to 0.812).   But you are right, coal and stone in the default ratios appear much higher than needed and means a lot goes unused.   They must have did that to give people flexibility early/mid game, because infinite science production takes a lot lower ratio.   For example coal needed for infinite science is only 1/10 the amount of iron (if nuclear and electric furnaces are used)

BlueRock2:
After some testing, overall map ore averages for this PR come out to iron = 1,  copper = 0.75,  coal = 0.38 and stone = .30.    I think the iron/copper/coal ratio is just about right now, but taking stone down just a little more and re-testing.

BlueRock2:
Made slight tweeks, mostly concerning stone this last revision.   Ran 10 map generations on 707x707 area (approx half million tiles) and captured ore counts.   Map ore generation ratio averages came out to iron/copper/coal/stone as 1/0.80/0.40/0.25.    I believe these are good numbers.

mheguy:

PR #592, opened 2018-12-25
Add random_RGB to utils

mheguy:
Adds a utility that returns a table with integer RGB values.

### DIFF
--- a/utils/core.lua
+++ b/utils/core.lua
@@ -1,10 +1,18 @@
-local Module = {}
+-- Dependencies
 local Game = require 'utils.game'
+
+-- localized functions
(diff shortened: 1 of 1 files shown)


linaori (commented):
What if we have a Colour util that has a bunch of predefined colours, including a random function? I believe the factorio stdlib has something like this as well. Means you could do `Color.red` for example

mheguy:
That sounds like a resources file

mheguy:
Can't edit comment on phone. Wanted to add to previous: 
Sounds like resources file but I really like the idea. Would simplify color selection a bunch.

PR #598, opened 2018-12-27
add circleci linting support

mspielberg:
Uses CircleCI instead of Travis to run luacheck for faster response

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,0 +1,12 @@
+version: 2
+jobs:
+  build:
+    docker:
+      - image: mspielberg/circleci-busted:5.2
+    steps:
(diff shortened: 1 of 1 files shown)


mheguy:
Stickler has done well and has a user-friendly approach since it simply comments the errors. Travis now catching PRs that stickler can't process (thanks again for a solid yaml for travis) closes that hole. I can't think of a good reason to add a third CI to the mix that nobody on the team is familiar with. Will close this for now, but if anyone feels strongly we can definitely have a discussion.

mspielberg:
The idea would be to drop Travis and replace it with CircleCI, but I definitely understand the benefit of using familiar tools.

PR #602, opened 2018-12-28
Re-added the blueprint helper for diggy

linaori:
There's no reason this should be disabled


Valansch:
I think it was disabled because of desyncs?

linaori:
No, I disabled it to see if people were using it in Diggy, seems like some were. It's the task-list that desyncs

Valansch:
Okay then.

PR #603, opened 2018-12-28
Ability to disable market items without removal

linaori:
This allows you to add market items that are disabled, and enable or disable them later with a message. This comes in handy for several things:
 - Diggy shows all items now, but makes it impossible to buy them until the level is reached
 - Custom triggers can be added with a message on how to unlock them
 - Example: buff enemy team biter strength by 10% for 30 seconds could be disabled for 1 minute after buying it.

Disabled tooltip:
image

Overview when reaching level 1 in Diggy:
image

When reaching all techs:
image

When you have not reached the level nor have enough coins:
image

When trying to buy something that's disabled (fallback when the GUI wasn't updated)
image

### DIFF
--- a/features/retailer.lua
+++ b/features/retailer.lua
@@ -19,7 +19,8 @@ local market_frame_close_button_name = Gui.uid_name()
 local item_button_name = Gui.uid_name()
 local count_slider_name = Gui.uid_name()
 local count_text_name = Gui.uid_name()
-local color_red = {r = 1, b = 0, g = 0}
+local color_red = {r = 255, b = 0, g = 0}
+local color_dark_grey = {r = 169, g = 169, b = 169}
 
 local Retailer = {}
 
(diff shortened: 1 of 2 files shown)


features/retailer.lua
```diff
@@ -19,7 +19,8 @@ local market_frame_close_button_name = Gui.uid_name()
 local item_button_name = Gui.uid_name()
 local count_slider_name = Gui.uid_name()
 local count_text_name = Gui.uid_name()
-local color_red = {r = 1, b = 0, g = 0}
+local color_red = {r = 255, b = 0, g = 0}
```
mheguy: #596 :D

mheguy (approved):
Looks good

features/retailer.lua
linaori: Yes, that's what I stole the dark_grey from! 😄

Issue #582, opened 2018-12-22
Finish Tetris


Valansch:
Todo list:

- [x] Trees
- [x] Make the buttons behave as the market, using the sprite buttons
   - "number" on the button is the amount of votes so you can see what's going on
   - Use the "tooltip" to indicate the action
   - show voters
- [x] Show a timer on when the vote closes
- [x] Add biters
- [x] Disable buttons after vote
- [ ] Keep the tetrominoes visible at all times
- [x] Make chest unminable
- [x] Change drop to down
- [ ] (Fix wrong amounts in chest)
- [x] Fix "rotate clockwise" text
- [x] Chart new terrain more aggressively
- [ ] Rechart map if resources are destroyed
- [x] Fix resource duplication when rotating. How to reproduce: (X marks resources)
```
[ ][ ]
   [ ][X]
```

Fudster:
Can you make the chest minable when it's empty?

Valansch:
That was my thought too. Yes thats possible.

Valansch:
I cannot reproduce the wrong amount in chests anymore. /shrugs

mheguy:
Closed via #582

Valansch:
Nonono closed via #605

mheguy:
Lol, closed via itself. bravo me

PR #605, opened 2018-12-29
Tetris

Valansch:
Here we go! :D
Working on more luadoc...
#582


map_gen/combined/tetris/control.lua
```diff
@@ -0,0 +1,396 @@
...
+    end
+    for candidate, n_votes in ipairs(vote_sum) do
+        if max == n_votes then
+            table.insert(winners, candidate)
+        end
+        View.set_vote_number(options[candidate].button, n_votes)
+    end
+    local winner_option_index = winners[math.random(#winners)]
+    primitives.winner_option_index =
+    if _DEBUG then
```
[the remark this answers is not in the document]
Valansch: You know nothing stickler

map_gen/combined/tetris/tetrimino.lua
```diff
@@ -0,0 +1,383 @@
...
+        game.print('You are a smart motherfucker, that\'s right.')
+    end
+
+    local old_collision_box = self.collision_box
+    local surface = self.surface
+    local find_tiles_filtered = surface.find_tiles_filtered
+    local find_entities_filtered = surface.find_entities_filtered
+    local tetri_x = self.position.x
+    local tetri_y = self.position.y
+    local insert = insert --Intentionally shadowing upvalue
```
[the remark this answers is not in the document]
Valansch: Intentional.

map_gen/combined/tetris/control.lua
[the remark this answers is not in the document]
Valansch: You still know nothing.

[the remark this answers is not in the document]
mheguy: you sure about that? I see an empty definition
`primitives.winner_option_index =`

map_gen/combined/tetris/control.lua
```diff
@@ -0,0 +1,396 @@
...
+    Task.set_timeout_in_ticks(16, move_down)
+end
+
+machine.register_state_tick_action(states.voting, execute_vote_tick)
+
+machine.register_state_tick_action(states.down, execute_down_tick)
+
+machine.register_transition(states.voting, states.pause, function()
+    View.enable_vote_buttons(true)
+    game.print('Pausing...')
```
mheguy: In testing, I didn't realize the pause was due to inactivity. Making that clear would be helpful.

map_gen/combined/tetris/tetrimino.lua
```diff
@@ -0,0 +1,383 @@
...
+    self.collision_box = new_collision_box
+
+    return true
+end
+
+local function get_next_tetri_number()
+    local head = sequence._head
+    if head > 7 then
+        table.shuffle_table(sequence)
+        game.print(serpent.line(sequence))
```
mheguy: Debug code printing during live play

mheguy:
Could really benefit from map info.
The magnifying glass icon's use is unclear (it seemingly does nothing).
Removing vote retains the number in the icon.
Removing vote seems to retain the vote itself: voting move right then clearing vote will move the piece right.
An indicator of some kind that a new piece has been spawned would be nice. Maybe a line in the GUI below the points/score?
The first time you vote a piece to "move down" it seems to ignore that vote and the piece only falls 1 space.

Valansch:
A new piece has spawned once all buttons are enabled again and the timer is running again. I don't feel we need any more indication of that.

Valansch:
I cannot reproduce the move down issue, it's working as intended for me.
Magnifying glass changes player zoom, i'll add a tooltip.
Clear button issue is fixed :+1:

map_gen/combined/tetris/control.lua
```diff
@@ -0,0 +1,399 @@
...
+    end
+    Debug.print('calculating winner')
+    local vote_sum = {0,0,0,0,0}
+    for _, vote in pairs(player_votes) do
+        vote_sum[vote] = vote_sum[vote] + 1
+    end
+
+    local winners = {}
+    local max = math.max(vote_sum[1], vote_sum[2], vote_sum[3], vote_sum[4], vote_sum[5])
+    for candidate, n_votes in ipairs(vote_sum) do
```
linaori: pairs is faster and does the same in factorio

mheguy: I'm curious under what conditions you found pairs faster than ipairs. I've tested previously and found ipairs faster.
Here's a quick test I whipped together for this comment:
image

linaori: As frequently said by Klonan and according to https://springrts.com/wiki/Lua_Performance#TEST_9:_for-loops

mheguy: *shrug*
It's not what the test shows, but honestly I avoid ipairs as much as I can since we should be using pairs or manually iterating anyways.

Jayefuu (commented):
Sorry, I don't have time to review this week. From a technical standpoint I'm sure Plague and Lin have you covered. From a gameplay standpoint, nice work getting it running!

map_gen/combined/tetris/control.lua
```diff
@@ -0,0 +1,399 @@
...
+    local active_qchunks = Tetrimino.active_qchunks(tetri)
+    local storage = {}
+
+    local full_rows = {}
+    local rows = {}
+    for _, qchunk in ipairs(active_qchunks) do
+        local q_y = qchunk.y
+        if not rows[q_y] then
+            rows[q_y] = true
+            local y = tetri.position.y + 16 * q_y - 14
```
linaori: `local tetri_y = tetri.position.y` on top so it's only called once

map_gen/combined/tetris/control.lua
```diff
@@ -0,0 +1,399 @@
...
+            local row_full = true
+            for x = -178, 178, 16 do
+                local tile = tetri.surface.get_tile(x, y)
```
linaori: `local tetri_get_tile = tetri.surface.get_tile` on top for a performance boost

map_gen/combined/tetris/control.lua
```diff
@@ -0,0 +1,399 @@
...
+        player.zoom = zoom
+        player_zoom[player.index] = zoom
+    end
+)
+
+local function spawn_new_tetrimino()
+    table.insert(tetriminos, Tetrimino.new(game.surfaces.nauvis, {x = 0, y = primitives.tetri_spawn_y_position}))
+end
+
+local function collect_full_row_resources(tetri)
```
linaori, line 162: This function has a bunch of calls that will benefit from extracting multiple table access calls to a local

map_gen/combined/tetris/control.lua
```diff
@@ -0,0 +1,399 @@
...
+        for key, tetri in pairs(tetriminos) do
+            if not Tetrimino.move(tetri, 0, 1) then
+                tetrimino_finished(tetri) --If collided with ground fire finished event
+                tetriminos[key] = nil
+            end
+
+            local pos = tetri.position
+            Task.set_timeout_in_ticks(10, chart_area, {
+                force = game.forces.player,
+                surface = game.surfaces.nauvis,
```
linaori: those API calls can be extracted to a local just above the for loop

map_gen/combined/tetris/control.lua
```diff
@@ -0,0 +1,399 @@
...
+        primitives.stale_vote_turns = 0
+        View.set_last_move(winner.button)
+    else
+        View.set_last_move(nil)
+        local stale_vote_turns = primitives.stale_vote_turns
+        if stale_vote_turns >= pause_after_n_ticks then
+            Task.set_timeout_in_ticks(1, switch_state, {state = states.pause})
+        else
+            primitives.stale_vote_turns = stale_vote_turns + 1
+        machine.transition(states.moving)
```
linaori: minor indent issue

linaori (changes requested):

map_gen/combined/tetris/control.lua
Valansch: I would instead want a pause button, but that doesn't quite fit in the gui.
Btw if anyone has gui style suggestions i would appreciate them greatly

map_gen/combined/tetris/control.lua
Valansch: I'm guessing you tested with far too few iteration 80ms is far too short to be significant.
I tested with 100 million iterations and ipairs is consistently get 3% slower than pairs.

But even at 1 million iterations ipairs is still 3% slower for me.
@iltar i assume you tested on windows?

Valansch: Actually.. I just got the opposite result. 1-3% difference is too small to test reliably. But i think we can agree one way or another to use pairs over ipairs and i=1,... over pairs.

map_gen/combined/tetris/control.lua
Valansch: This is called 4 times. So 6 additional table accesses. About once per minute. There comes a point where readability is more important than unmeasurable performance differences. It remains questionable if this is the case here :thinking:

linaori: There shouldn't be a loss of readability when assigning to a local, if anything it would actually make it more readable because you can immediately track where this local is used

map_gen/combined/tetris/tetrimino.lua
```diff
@@ -0,0 +1,371 @@
...
+
+--- @type LuaTetriminoCollisionBox represents the collision box of a tetrimino. Contains a 4x4 2d array.
+-- @description If a cell is 1 the quad chunk represented by the array position is occupied. 0 if not
+-- @field 1 table of numbers first row
+-- @field 2 table of numbers second row
+-- @field 3 table of numbers thrid row
+-- @field 4 table of numbers forth  row
+
+local Token = require 'utils.token'
+local Task = require 'utils.schedule'
```
mheguy: Affected by #629
(`schedule` to `task`, `q` to `queue`

linaori (approved):

mheguy (approved):

mheguy:
A tour de force, truly the best among us. Code that will stand the test of time.

PR #607, opened 2018-12-29
Tweaked the Market GUI with auto-close and refresh

linaori:
- When a player runs too far from the market they had opened the GUI of, it will now close
 - When something changes in the market GUI for the currently open GUI for a player, it will refresh the contents


control.lua
```diff
@@ -3,6 +3,10 @@ require 'utils.math'
 
 -- Map layout and config dictate the map you play and the settings in it
 local config = require 'config'
+
+-- global for easy access
+Debug = require 'utils.debug'
```
mheguy, line 5: Since the comment above addresses config and map layout it might be best not to break them up.

And the comment for this specifically should probably be a little more descriptive (because by that logic everything should be global for easy access)

linaori: This needs to be loaded before the map_layout, otherwise it won't be present before the majority of the code is execute. I can move it to just above the config though, that would solve the problem

mheguy: Makes sense to me that we would load debug before the config.

mheguy:
Resolve conflicts and feel free to merge

PR #611, opened 2018-12-30
Document utils from factorio's util.lua

mheguy:
Of course a bit of refactoring for some flavor. :D

### DIFF
--- a/utils/table.lua
+++ b/utils/table.lua
@@ -9,7 +9,7 @@ local table_size = table_size
 --- Searches a table to remove a specific element without an index
 -- @param t table to search
 -- @param element to search for
-table.remove_element = function(t, element)
+function table.remove_element(t, element)
     for k, v in pairs(t) do
         if v == element then
             remove(t, k)
@@ -21,7 +21,7 @@ end
 --- Adds the contents of table t2 to table t1
 -- @param t1 table to insert into
 -- @param t2 table to insert from
-table.add_all = function(t1, t2)
+function table.add_all(t1, t2)
     for k, v in pairs(t2) do
         if tonumber(k) then
             insert(t1, v)
@@ -35,7 +35,7 @@ end
 -- @param t table to search
 -- @param e element to search for
 -- @returns the index of an element or -1
-table.index_of = function(t, e)
+function table.index_of(t, e)
     local i = 1
     for _, v in pairs(t) do
         if v == e then
@@ -52,7 +52,7 @@ local index_of = table.index_of
 -- @param t table to search
 -- @param e element to search for
 -- @returns true or false
-table.contains = function(t, e)
+function table.contains(t, e)
     return index_of(t, e) > -1
 end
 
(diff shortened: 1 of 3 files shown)


linaori (approved):

utils/table.lua
```diff
@@ -197,6 +197,40 @@ table.binary_search = function(t, target)
...
+-- return <string> the prettied table
+table.inspect = require 'inspect'
+
+--- Takes a table and returns the number of entries in the table. (Slower than #table, faster than iterating via pairs)
 table.size = table_size
+
+--- Creates a copy of a table, metatable, or LuaObject.
+-- @param object <table|LuaObject> the object to copy
+-- @return <table|LuaObject> the copied object
+table.deep_copy = table.deepcopy
```
grilledham, line 219: I looked at the implementation and I don't think this comment accurately describe what this function does.

- It actually takes any type, but given it makes no sense to copy anything but a table it's seems reasonable to say it takes table, instead of any.
- LuaObjects are tables, so this is either redundant, or the comment should mention special treatment for LuaObjects 
- metatables are tables, so same as above.
- The comment implies that this creates a deepcopy of metatables and LuaObjects which it doesn't.

A better describtion might be: Creates a deepcopy of a table. Metatables and LuaObjects inside the table are shallow copies.

utils/table.lua
```diff
@@ -197,6 +197,40 @@ table.binary_search = function(t, target)
...
+--- Compares the contents of two tables and returns if they are equal
+-- Notice: API calls are never equal to tables/stored LuaObjects.
+-- Ex. this_player = game.player
+-- table.compare(this_player, game.player) -- false
+-- table.compare(this_player, this_player) -- true
+-- table.compare(game.player, game.player) -- true
+-- @param tbl1 <table>
+-- @param tbl2 <table>
+-- @return <boolean>
+table.equals = table.compare
```
grilledham, line 232: I think the comment should say: This function should never be used to compare tables that are LuaObjects or tables that contain LuaObjects, because of the absurdity of the example you gave.
Note that the correct way to compare LuaObjects is with the `==` operator
```lua
this_player = game.player
this_player == game.player -- true

plague006: Changed in https://github.com/Refactorio/RedMew/pull/611/commits/bb44d51125807e35b64ffd94f02f74b2cd7c19c6

utils/table.lua
plague006: Changed in https://github.com/Refactorio/RedMew/pull/611/commits/bb44d51125807e35b64ffd94f02f74b2cd7c19c6

mheguy: Corrected in https://github.com/Refactorio/RedMew/pull/611/commits/0a88dfbf3dbacd909dd22dcf922b9922292f8baf

utils/table.lua
mheguy: I'm not sure what you actually want the comment to say. I copy pasted what you have currently but only read it once I pushed https://github.com/Refactorio/RedMew/commit/bb44d51125807e35b64ffd94f02f74b2cd7c19c6. I don't think you wanted that as the comment? Unless I'm mistaken.

mheguy: Discussed and corrected on discord

grilledham (approved):

PR #616, opened 2018-12-31
Added some basic docs about the Retailer

linaori:


Jayefuu (approved):
Comments describing the retailer functionality match the code Lin provided me to test in biter lane wars last week. Can confirm that the example code works as provided.

PR #617, opened 2018-12-31
Fish market with new GUI and stack limit

linaori:
This makes the fish market use the new market GUI and adds a stack limit to the items you can buy. This is particularly useful if you want to apply buffs with a max of 1 time and/or replace/update/disable a market item after X units is purchased.

image

image

image

image


mheguy:
> disable a market item after X units is purchased.

The idea of scarcity in the market is a cool idea, I like it.

linaori:
@plague006 currently it's only a stack limit what you can buy at the same time, but that's something that can be added in the future yes

features/fish_market.lua
```diff
@@ -27,28 +33,23 @@ local function spawn_market(_, player)
...
+    Retailer.add_market('fish_market', market)
+
+    for _, prototype in pairs(market_items) do
+        Retailer.set_item('fish_market', prototype)
     end
 
-    force.add_chart_tag(
-        surface,
-        {
-            icon = {type = 'item', name = market_item},
```
mheguy, line 41: I kinda like having the coin icon for the market since it's unique (can't be added by players). Any particular reason for removing it?

features/train_saviour.lua
```diff
@@ -1,5 +1,5 @@
 local Event = require 'utils.event'
-local Market_items = require 'resources.market_items'
+local market_items = require 'resources.market_items'
```
mheguy, line 2: The standard of localized requires starting with an upper case letter seems pretty consistent in the codebase. Any particular reason for the deviation?

mheguy (dismissed):
Looks good overall. It's worth noting that removing the wood market removes a key feature of the toxic_jungle map. It's worth noting this in map_layout.lua and/or toxic_jungle.lua if not making some adjustments to the map to compensate for the change.

linaori:
> It's worth noting this in map_layout.lua and/or toxic_jungle.lua if not making some adjustments to the map to compensate for the change.

I could add something like `error('this map is incomplete as there is no market compatible with the current wood system')`. The other option would be to revert the wood market but have no contents.

features/fish_market.lua
linaori: It's pretty much just clutter in a big map, but I can bring it back if needed

linaori:
Found an issue where the jungle map would remove flamethrowers (which was already broken due to plane and moving of items), so I've also fixed that. Currently you just have to use coins. There are 0 references to the wood market and it's compatible, so I'm going to leave it at that. If we wish to launch the toxic jungle and want to use wood as a currency, we have to make something to support something other than coins for the market. 

image

With infinite storage chest enabled as feature instead of misc map thing
image

infinite storage chest contents as loaded per feature
image

features/train_saviour.lua
linaori: It's how we name stuff that's not going to add behavior. Resources are big tables stored in an external file. This is how we would've named the table if it was defined in this file, hence I changed the style of the variable.

linaori:
Okay something is wrong with Travis (as usual), so we can ignore that most likely

PR #618, opened 2019-01-01
Update poke_messages.lua

SieveyEnvoys:
278 » 400 » ?
Yes I'm Vintage
I'll probably forget about this unless you ping me
__


linaori (approved):
I like them!

grilledham (approved):
👍 
Maybe some are repetitive, but that wouldn't be any different from the current list.

PR #621, opened 2019-01-03
 Have info pull changelog from server 

mheguy:
Pulls the `changelog` from the `misc` dataset and (if available) overwrites the existing (read from config) data in the "What's new" tab of info.

### DIFF
--- a/features/gui/info.lua
+++ b/features/gui/info.lua
@@ -699,7 +699,11 @@ local changelog_callback =
 
 Event.add(defines.events.on_player_created, player_created)
 
-Event.on_init(Server.try_get_data('misc', 'changelog', changelog_callback))
+Event.add(Server.events.on_server_started,
+    function()
+        Server.try_get_data('misc', 'changelog', changelog_callback)
+    end
+)
 
 Gui.on_click(main_button_name, toggle)
 
--- a/features/gui/info.lua
+++ b/features/gui/info.lua
@@ -5,6 +5,8 @@ local UserGroups = require 'features.user_groups'
 local Game = require 'utils.game'
 local PlayerRewards = require 'utils.player_rewards'
 local Color = require 'resources.color_presets'
+local Server = require 'features.server'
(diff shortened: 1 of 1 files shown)


features/gui/info.lua
```diff
@@ -684,8 +686,21 @@ local function create_map_extra_info(value, set)
...
+        local value = data.value -- will be nil if no data
+        if value then
+            editable_info[new_info_key] = value
+        end
+    end
+)
+
 Event.add(defines.events.on_player_created, player_created)
 
+Event.on_init(Server.try_get_data('misc', 'changelog', changelog_callback))
```
grilledham: I assume you were tired when you wrote this 😄 `Server.try_get_data` returns nil, you need to pass a function into `Event.on_init`.

Also `Event.on_init` isn't the right event to use. If you create a save in single player `on_init` is called then and only then. When you load the save on the server `on_load` is called. But `on_load` is called for every player that joins, so it is also the wrong event to use. 

For this type of problem I created the `Server.events.on_server_started` event, which does what you want. For an example of usage, see https://github.com/Refactorio/RedMew/blob/develop/features/user_groups.lua#L141

grilledham (dismissed):

grilledham:
On discord you talked about line breaks not working when stored in datasets. Is that still a problem, should I look into fixing that before this merges?

mheguy:
> Is that still a problem, should I look into fixing that before this merges?

Yes still a problem, not required for this to merge since I removed the functionality. Once the issue with line breaks is resolved I'll add a way to pass data from the game to the server.

features/gui/info.lua
mheguy: > I assume you were tired

Not tired, just dumb. Was especially misled since it worked just fine. :(

> on_server_started

Quite right.

Both implemented in ff01cff

Valansch (approved):
Good stuff

PR #623, opened 2019-01-03
Add source of message to Debug.print

mheguy:
Adds the source of the debug call to the debug message by default. Pass `true` to `no_source` to disable the inclusion of the source.

Also added a check for `game` to allow debug.print to run pre-init.

### DIFF
--- a/utils/debug.lua
+++ b/utils/debug.lua
@@ -18,8 +18,7 @@ end
 
 --- Takes the output from debug.getinfo and strips the function and full source names
 local function cleanup_debug(debug_table)
-    local clean_string = format('caller details: name: %s file: %s, line number: %s', debug_table.name, debug_table.short_src, debug_table.currentline)
-    return clean_string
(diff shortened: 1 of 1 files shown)


utils/debug.lua
```diff
@@ -16,9 +16,16 @@ local function increment()
     return next
 end
 
+--- Takes the output from debug.getinfo and strips the function and full source names
+local function cleanup_debug(debug_table)
+    local clean_string = format('caller details: name: %s file: %s, line number: %s', debug_table.name, debug_table.short_src, debug_table.currentline)
+    return clean_string
```
linaori: You can immediately return the result of format here

mheguy: A leftover from when I was removing a few bits rather than only adding a few bits. Thx for the spot. :)

linaori (approved):

Valansch:
Fixed your commit message.

PR #625, opened 2019-01-03
Add quality of life features

### LINKED ISSUE #551: Automatically limit passive provider chests to a configurable open slots
On the beach map someone suggested it would be nice if there was a mod to automatically limit the amount of open slots a passive provider chests has. Obviously a mod is overkill and we can actually do this in a scenario!

Should probably be done when an entity is placed by either a bot or player. Can use this API: https://lua-api.factorio.com/latest/LuaInventory.html#LuaInventory.hasbar

 - Might have to check whether it came from a blueprint and has pre-configured limited slots. 
 - Might cause issues with buffer stations that use passive provider chests. Fix for this is to adjust one and drag copy the rest

This can be implemented as RedMew  `features/limit_provider_chest.lua` and receive a configurable amount of slots (defaults to 1).

mheguy:
This adds 3 modules, each limited in scope, triggered by on_built (player and/or bot as applicable) and is toggle-able during runtime via public functions.

### restrict_chest
Closes #551
Disabled by default as I would like to add an individual opt-out in the future.
When a chest is placed, it restricts the number of slots available to just 1. This is intended to prevent assemblers outputting too many of a given product.

Specifics:
Only applies to passive provider chests and non-logistics chests.
The chest is only restricted if the placed chest would be 'uncapped'. So if the chest is being placed into a ghost which has a restriction, the restriction from the ghost takes precedence.
If a chest is upgraded to a chest of a different type (ex iron to steel), it will maintain the restrictions if one existed.
However:
If a chest is upgraded and the former chest was uncapped, the new chest will be restricted. At some point I might want to fix this.

### backer_name 
An extension of train_station_names (which this replaces).

Instead of only train stations, all entities which support backer names have a chance of being named after a player on the server, a regular, or keeping the backer name they were given.

### random_train_color 
If a placed entity is a locomotive, apply a random RGB color to it. Just adds a bit of fun to trains.

### ghosts_before_research
Not a module on its own per se, but on init set a 1 hour ghost time to live for the player force. When entities are destroyed they will now get ghosts even if robots are not researched. Especially helpful when trying to rebuild a complex blueprint.


linaori:
> If a chest is upgraded and the former chest was uncapped, the new chest will be restricted. At some point I might want to fix this.

Not having this will cause issues when upgrading unloading stations or buffers for example. I fear this might give some problems and unexpected behavior, leading into bottlenecks.

mheguy:
I would contend that buffers don't lead to bottlenecks, they simply put a timer on them. But I don't want to force a playstyle on others.

Instead I would like to stick to the current implementation for the time being (with it defaulting to disabled via the config) and come around to it when I have the time. As mentioned I want to add a per-player opt-out and I can fix that behaviour when I'm diving back into it.

Edit: It's also something that I think is reliant on 0.17 since bots will be able to fast replace entities, I feel I would end up having to do the same work twice so would rather do it all at once.

features/redmew_qol.lua
```diff
@@ -0,0 +1,214 @@
...
+
+-- Global registers
+local enabled = {}
+
+Global.register(
+    {
+        enabled = enabled
+    },
+    function(tbl)
+        enabled = tbl.enabled
```
Valansch, line 29: missing trailing comma /s

Valansch (approved):
Great stuff. Thank you.

PR #631, opened 2019-01-05
Redmew surface

mheguy:
## Redmew Surface

Broken up into commits to make review easier. There are 65 files but only 4 files are added and make up this module. `redmew_surface` (hereafter RS) itself does all the work, and there are 3 resource files for the individual aspects of world generation: `map_gen_settings`, `map_settings`, and `difficulty_settings`.

### RS for users

As long as a map hasn't explicitly selected to control an aspect of world generation, a user's choices are respected. Further, a user can also disable RS entirely or the individual aspects of worldgen via config settings.

### RS for devs

RS changes the surface that maps are generated on from `nauvis` to a new surface: `redmew`. It offers a way to acquire the surface name pre-init through `get_surface_name` and allows us to get the surface itself from `get_surface` after init.

```lua
```

These methods are intended for modules that act on a surface without being able to determine which surface to act on (ie. modules not being triggered by events that provide a surface).

##### Functions

The strength of RS is its ability to control world generation for maps. We no longer have to forcefully scan for and remove ores and trees or water via LUA. We can now simply disable their placement. However RS doesn't force a map to specify more than it wants/needs. If a map only wants to remove all copper ore generation while leaving all other settings to the user: it can do that.

RS splits worldgen into the 3 components that factorio itself splits them into: map_gen_settings, map_settings, and difficulty_settings.

For each of the 3 components, RS has an associated public function: `set_map_gen_settings`, `set_map_settings`, and `set_difficulty_settings`

RS works essentially by overwriting. The base is the user's settings and values provided through the `set` functions overwrite existing values. For example if a user selects rich resources but a map uses the `add` function to remove all copper: the map will spawn no copper but the others at rich settings.

##### Resources

To make things easier on map makers, there are many preset world gen options available to select from. These presets/components are available in 3 resource files named for each of the world gen components: `map_gen_settings.lua`, `map_settings.lua`, and `difficulty_settings.lua`. Each file contains in it tables of varying size/scope and settings. For example `waterworld` in `map_gen_settings.lua` contains a table that ensures nothing but water will spawn. This can be nice for a map that wants to fully create its terrain, ores, and entities.

#### Examples
To create a map with no ores, no enemies, no pollution, no enemy evolution, 3x tech costs, and sand set to high we would use the following:

```lua
-- We require redmew_surface to access the public functions and assign the table Public to the RS variable to access them easily.
-- We require the resources tables so that we don't have to write settings components by hand.
local MGSP = require 'resources.map_gen_settings' -- map gen settings presets
local DSP = require 'resources.difficulty_settings' -- difficulty settings presets
local MSP = require 'resources.map_settings' -- map settings presets

-- We create a custom table for the niche settings of wanting more sand
local extra_sand = {
    autoplace_controls = {
        sand = {frequency = 'high', size = 'high'}
    }
}

RS.set_map_gen_settings({MGSP.enemy_none, MGSP.ore_none, MGSP.oil_none, extra_sand})
RS.set_difficulty_settings({DSP.tech_x3})
RS.set_map_settings({MSP.enemy_evolution_off, MSP.pollution_off})
```
A water world:
```lua

local MGSP = require 'resources.map_gen_settings' -- map gen settings presets

RS.set_map_gen_settings({MGSP.waterworld})
```
A map that wants very high enemies and cliffs, but also wants very specific iron and copper generation. The ore generation is too specific to add to the resource file so we create a table for it:
```lua
local MGSP = require 'resources.map_gen_settings' -- map gen settings presets

local ore_gen = 
        {
            autoplace_controls = {
                ['iron-ore'] = {
                    frequency = 'high',
                    richness = 'very-low',
                    size = 'very-high'
                },
                ['copper-ore'] = {
                    frequency = 'low',
                    richness = 'very-high',
                    size = 'very-low'
                }
            }
        }

RS.set_map_gen_settings({MGSP.enemy_very_high, MGSP.cliff_very_high, ore_gen})
```
Updated for https://github.com/Refactorio/RedMew/pull/631/commits/19cbd47b5fc0dd8e16f1029a11a76d46f6c806f8


mheguy:
I'm waking travis up, he'll find errors on every commit because he scans the entire codebase.

map_gen/combined/dimensions.lua
```diff
@@ -15,149 +15,148 @@ global.current_magic_chest_index = 1
...
-        player.teleport(portal.target, portal.target_surface)
-        global.last_tp[player.name] = game.tick
-        player.print("Wooosh! You are now in the " .. portal.target_surface.name .. " dimention.")
-      end
+    for _, player_character in pairs(portal.source.find_entities_filtered {area = {{portal.position.x - global.portal_radius, portal.position.y - global.portal_radius}, {portal.position.x + global.portal_radius, portal.position.y + global.portal_radius}}, name = 'player', type = 'player'}) do
+        local player = player_character.player
+        if not global.last_tp[player.name] or global.last_tp[player.name] + global.teleport_cooldown * 60 < game.tick then
+            player.teleport(portal.target, portal.target_surface)
+            global.last_tp[player.name] = game.tick
+            player.print('Wooosh! You are now in the ' .. portal.target_surface.name .. ' dimention.')
```
linaori: dimention -> dimension? :D

map_gen/combined/dimensions.lua
```diff
@@ -15,149 +15,148 @@ global.current_magic_chest_index = 1
...
-  end
 end
 
-local function dim_on_tick(event)
-  if game.tick % 2 == 0 then
-    teleport_stuff()
-  else
-    teleport_players()
-  end
+local function dim_on_tick()
```
linaori, line 115: At some point we might have to improve performance here 😅 Not sure if we've used this map recently, but there are a bunch of things we can improve here

mheguy: Not for as long as I've been on redmew, so 6-7 months?
Regardless, this is far beyond the scope of this PR and I'm marking this resolved.

map_gen/shared/redmew_surface.lua
```diff
@@ -0,0 +1,324 @@
...
+            game.map_settings.max_failed_behavior_count = combined_map_settings.max_failed_behavior_count
+        end
+    end
+end
+
+--- Creates a new surface
+local function create_redmew_surface()
+    local surface
+
+    if global.config.redmew_surface.enabled and global.config.redmew_surface.map_gen_settings then
```
grilledham: when is `global.config.redmew_surface.enabled` set to `true`? And if it was `false` why are we still creating the redmew surface?

map_gen/shared/redmew_surface.lua
```diff
@@ -0,0 +1,324 @@
...
+
+    RS.add_difficulty_settings({difficulty_settings_presets.tech_x3, difficulty_settings_presets.tech_x4})
+
+    To create a map with no ores, no enemies, no pollution, no enemy evolution, 3x tech costs, but leaving everything else up to the user
+    we would use the following:
+
+    -- We require redmew_surface to access the public functions and assign the table Public to the RS variable to access them easily.
+    local RS = require 'map_gen.shared.redmew_surface'
+    -- We require the resources tables so that we don't have to write settings components by hand.
+    -- In general, don't create a table custom for a map, instead just add it to the resource files and call that so that others can make use of it.
```
grilledham: >> In general, don't create a table custom for a map ...

That is opinionated, I think people can make custom tables for niche presets if they prefer. Having a resource file full of niche presets could arguably make it harder to read.

map_gen/shared/redmew_surface.lua
```diff
@@ -0,0 +1,324 @@
...
+    set_difficulty_settings()
+    set_map_settings()
+
+    surface.request_to_generate_chunks({0, 0}, 4)
+    surface.force_generate_chunk_requests()
+    game.forces.player.set_spawn_position({0, 0}, surface)
+end
+
+--- On player create, teleport the player to the redmew surface
+--- When placed, locomotives will get a random color
```
grilledham: >> --- When placed, locomotives will get a random color

Copy and paste error?

grilledham (commented):
I like the difference between `add` and `set` but wonder if the names could be better. `add` to me implies that you can do it multiple times like adding to a list, I assume your thinking is it's like extend. 
But given you have to explain what `add` and `set` do, maybe we should just call them `with_user` and `with_default`

grilledham:
Overall this is a very cool module which will open up some great possibilities. 👍 

 However there is a lot to go through, so I still haven't finished my review.

map_gen/shared/redmew_surface.lua
mheguy: Set via config during control initialization.
I don't know why I had it in my head that the surface should be created regardless. Changed to just set `surface_name` to `'nauvis'` if RS is disabled.

map_gen/shared/redmew_surface.lua
mheguy: I sure hope so. Thanks for the catch.

map_gen/shape/infinite_mazes.lua
```diff
@@ -120,57 +142,60 @@ end
...
 
-    for a=1,iron_ore_count do
+    for _ = 1, iron_ore_count do
         maze_seed, value = get_random_maze_val(maze_seed)
         local pos = table.remove(coord_list, value % #coord_list + 1)
-        maze_data[pos[1]][pos[2]] = "iron-ore"
+        maze_data[pos[1]][pos[2]] = 'iron-ore'
     end
-    for a=1,copper_ore_count do
+    for _ = 1, copper_ore_count do
```
mheguy, line 182: I'm assuming this was to correct a linting error for an unused loop variable:
luacheck will also take `i` as a loop variable and not generate a warning if it's unused (just an FYI thing).

plague006:
>I like the difference between `add` and `set` but wonder if the names could be better. `add` to me implies that you can do it multiple times like adding to a list, I assume your thinking is it's like extend. 
But given you have to explain what `add` and `set` do, maybe we should just call them `with_user` and `with_default`

This makes me wonder if there's a benefit to an `add` and a `set`. You're right that the names are sub-optimal since adding with `add` may not have the intended effect since conflicting settings will overwrite one another. For `with_user_settings` and `with_default_settings` properly and accurately describing how they're distinct from one another is also giving me a good bit of difficulty.

My primary issue at this point, is that RS is meant to give map makers more tools while also being very easy to use. I'm wondering if, instead of learning two functions entirely, it isn't just easier to learn 1 function and how to use it in order to take full control. That is, instead of learning `add_map_gen_settings` vs `set_map_gen_settings`, simply learning that by giving `set_map_gen_settings` the `defaults` table from the presets you can override all of a player's options.

I know it's an extra step (ie. adding 1 extra argument to a function call) for map-makers but I think wanting to override *every* option from a user is a rare/niche case.

grilledham:
I think that makes sense. I would imagine that few maps will want to override the user settings with default. They would either want the user settings or have a particular setting they want to enforce, like no biters.

So starting with user settings and overriding makes sense. And if you really want to start from default settings for whatever reason you can with the default preset.

linaori:
Isn't it the goal to _overwrite_ specifics? This could be achieved with other solutions as well:

```lua
config.set('map_config.property.in.the.config', 100) -- has to convert the string to nested table indexes
config.set(function(map_config) 
    map_config.property.in.the.config = 100
end)
```

mheguy:
I'm not completely following your suggestion lin

mheguy:
https://github.com/Refactorio/RedMew/pull/631/commits/19cbd47b5fc0dd8e16f1029a11a76d46f6c806f8:
The `add` public functions have been removed. However the `set` functions no longer completely overwrite user options. Overwriting all user options remains an option as long as all options are set by the map.
The goal of this is to make the entire module easier to understand by reducing the number of functions while maintaining the same capabilities.

The documentation *should* be updated properly and should make no mention of "set" vs "add" nor should it mention anything about a "default".

linaori:
@plague006 it's the ability to set parts of the map config without having to expose the whole table

mheguy:
Made `create_redmew_surface` public so that if a map wants to delete and re-create the surface, it can do so more easily.

This change only makes the function public, it does *not* handle the fallout of deleting a surface.

map_gen/shared/redmew_surface.lua
```diff
@@ -0,0 +1,289 @@
...
+        end
+        Public.remove_player_created_event()
+    end
+)
+
+local function init()
+    if global.config.redmew_surface.enabled and global.config.redmew_surface.map_gen_settings then
+        Public.create_redmew_surface()
+    else
+        surface_name = 'nauvis'
```
grilledham: Wouldn't this cause desyncs. If you change a variable in `on_init` it either has to be in global or you have to make the same change in `on_load`.

map_gen/shared/redmew_surface.lua
```diff
@@ -0,0 +1,289 @@
...
+-- Localized functions
+local insert = table.insert
+local merge = util.merge
+local format = string.format
+
+-- Local vars
+local Public = {}
+
+-- Constants
+local surface_name = 'redmew'
```
grilledham: Why is this in the constants section when it is mutated on line 205?

map_gen/shared/redmew_surface.lua
```diff
@@ -0,0 +1,289 @@
...
+
+-- Global tokens
+local data = {
+    ['map_gen_settings_components'] = {},
+    ['map_settings_components'] = {},
+    ['difficulty_settings_components'] = {}
+}
+
+-- This just creates an empty primitives table, the definitions are to document what primitives might exist.
+local primitives = {
```
grilledham: Do we expect changes to `primitives ` and `data` to happen after `on_init`? If not do they need to be in global?

grilledham (changes requested):
Using `RS.set...` is really easy and everything just works as I'd expect, good job!
I only tried for `set_map_gen_settings` but I assume it works for the others too.

However, I have some questions / concerns about the implementation of redmew_surface.lua.

- Why is `create_redmew_surface()` public, what is the expected use case? Do we expect to have redmew_surface disabled, but still have a module want to create the surface anyway?

- Throwing errors when the `RS.set ...` functions are called twice doesn't seem like a good idea to me. Surely it's possible for two modules to make changes to map settings in a way that doesn't interfere. And if they do interfere is throwing an error better than just letting one module override the other.

- if `global.config.redmew_surface.map_gen_settings = false`  then `set_difficulty_settings()` and `set_map_settings()` don't get called.

- Just a style thing but, there are multiple times in the code where you have `global.config.redmew_surface. ...` Why not make a `local config = global.config.redmew_surface` near the top of the file. 

- In the resource files, I'd be tempted to move the default presets to the top. They are the first thing I looked for to understand the structure of the tables.

mheguy:
On mobile, so formatting will suck:
Create surface public: the config setting should take precedence. I dont have an actual use case in mind for it. I was tooling around and deleted the redmew surface and realized there was no way to recreate it.

Throwing an error: the rationale was that it could create unexpected behaviour unless the author was very careful about things like loading order. 

Can and will localize config, can and will properly handle possible config paths. 

Default at top of resources: I've debated back and forth with myself multiple times. As a doc tool the default at the top is great. As a "big clump of useless text" it slows down searching for coomponents. Would splitting the difference make sense? Throw a comment at the top pointing to the default table at the bottom?

map_gen/shared/redmew_surface.lua
mheguy: Will fix, either a leftover in code or in my mind from when RS would create a surface no matter what (and the surface name would not change)

plague006: My mind*

map_gen/shared/redmew_surface.lua
mheguy: I think if I put data and prim to local, I would move create_surface along with it.

map_gen/shared/redmew_surface.lua
mheguy: I appreciate you phrasing your statement in then form of a questions. 😛 

Surface_name was meant to be constant. When I made this change I didn't think through the consequences. Will fix.

map_gen/shared/redmew_surface.lua
mheguy: b6c1b7c changes create_surface back to local and so removes data and most of primitives from the global scope.

map_gen/shared/redmew_surface.lua
mheguy: Fixed in b6c1b7c

map_gen/shared/redmew_surface.lua
mheguy: Fixed in b6c1b7c

mheguy:
Need to fix spawn location/players not spawning on redmew surface.

grilledham (approved):

mheguy:
Thanks for polishing my turd grilled! :D
image

PR #634, opened 2019-01-06
Event fixes

grilledham:
Some quick fixes to the event module to fix issues identified by plague.

- Adding removable events in on_init should work now.
- Made it clear that adding removable events in on_load is not needed and will crash the game.


utils/event.lua
```diff
@@ -46,6 +50,7 @@
 -- ** Event.add_removable_function(event_name, func) **
 --
 -- Only use this function if you can't use Event.add_removable. i.e you are registering the handler at the console.
+-- The same restriction that apply to Event.add_removable also apply to Event.add_removable_function.
```
mheguy: "restrictions" or "applies"

utils/event_core.lua
```diff
@@ -6,7 +6,7 @@ local Public = {}
 local init_event_name = -1
 local load_event_name = -2
 
-Public.runtime = false
+Public.runtime = false -- Set to true after on_init or on_load has finished.
```
mheguy, line 9: Beyond the scope of this PR but:
Would it be worth also copying this to the `global` table? Something like `global.runtime` so that other modules that might want to check for that can do so without requiring `event_core`?

mheguy (approved):
Looks good to me. I can test this later in the week if needed, but I'm personally comfortable with merging and testing as we go.

utils/event_core.lua
grilledham: We could move it out of event_core. I did think about that but also considered beyond the scope of the event PRs.

We can't move this variable to the global table for two reasons. Firstly it's accessed before on_load which is before the globals are synced. Secondly it is modified in on_load and modifying global variables in on_load crashes the game.

Maybe the name is misleading, it's more to do with the stage that the client's code is in during loading. Therefore it would never need to be in the global table.

Valansch: Why have Public.runtime at all?
It seems to me that it does the same as `(game ~= nil)`

grilledham: I set `Public.runtime = true` at the end of `on_init`, `game` is accessible at the start of `on_init`

mheguy: Makes sense why it can't be in global. 👍

mheguy (approved):
With d74a9fc in place, no further issues.
Looking forward to merging once conflict is resolved.

Valansch:
Please double check my changes.

Valansch:
Wrong market event fixed in #642 see: https://github.com/Refactorio/RedMew/pull/642#pullrequestreview-191233703

mheguy:

PR #635, opened 2019-01-07
fixed \n escaping + small optimizations

grilledham:
This should fix the issue @plague006 had with new lines in datasets

### DIFF
--- a/features/server.lua
+++ b/features/server.lua
@@ -353,10 +353,11 @@ function Public.get_tracked_data_sets()
     for k, _ in pairs(data_set_handlers) do
         k = double_escape(k)
 
-        message[#message + 1] = '"'
-        message[#message + 1] = k
-        message[#message + 1] = '"'
-        message[#message + 1] = ','
+        local message_length = #message
+        message[message_length + 1] = '"'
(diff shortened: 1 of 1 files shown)


features/server.lua
```diff
@@ -344,22 +351,21 @@ function Public.get_tracked_data_sets()
...
+        k = double_escape(k)
 
-        table.insert(message, '"')
-        table.insert(message, k)
-        table.insert(message, '"')
-        table.insert(message, ',')
+        message[#message + 1] = '"'
+        message[#message + 1] = k
+        message[#message + 1] = '"'
+        message[#message + 1] = ','
```
linaori: This would be even better:
```lua
local message_length = #message
message[message_length + 1] = '"'
message[message_length + 2] = k
message[message_length + 3] = '"'
message[message_length + 4] = ','
```

linaori (approved):

features/server.lua
grilledham: yes

mheguy:
Works as expected now when pushing info:new_info to a dataset. Thanks for the fix. :)

PR #640, opened 2019-01-09
Change map loading method

mheguy:
### What?
In essence, this takes `map_layout` and breaks it up into parts.

The list of maps, entity, and terrain modules moves to resource files (maybe in the future a web-based list?)
Map selection itself is made into a distinct file which simply returns the name of the map.
The config options are now in config (threaded, regen decoratives, entity_modules, terrain_modules, etc) which will allow maps to determine the settings for themselves.
The generation of the map (that is, the calling of `generate`) is done by map_loader.

~~Val says he'll be able to handle the dynamic require soon(tm).~~

In addition, this PR will move around some of the contents of `map_gen`.
There will only be 5 folders in `map_gen` itself:
`maps` (`combined`+`presets`+`shape`), `features` (most of `misc`), `shared`, `data`, `entity` (`ores` + `terrain` + some of `misc`), `terrain` (tris_chunk_grid).

The distinction between `terrain` and `features` essentially being: terrain modules changes the map at creation without event triggers (player input). features are more "gameplay elements" in that they respond to player input in the game itself (events)

`maps` will contain subfolders for maps that have multiple files (with the exception map data derived from binary images which will remain in `data/presets`)

### Why?

It makes the file structure a little easier for newcomers to jump into. It makes the role of things like terrain modules more clear.

It makes selecting a map easier for end users and those less technically inclined. It gives maps access to config options they didn't previously have. (ex. decoratives regenning, tiles per tick, entity and terrain modules)

The biggest reason I'm personally excited though is that it opens up the very real possibility of launching maps from a front-end.

The way I imagine something like that going: present the user with the map list, they select beach and we change/replace map_selection.lua with 1 line: `return 'beach'` and the server is started with that map. It means any staff member can just click a button and launch a server with an existing map.

In some far flung future the list of maps might even contain the preview images and descriptions of maps.

### When?

I'm intentionally holding this as WIP and unfinished until the framework is agreed on.

After an agreement is reached on the framework I'll commit the movement of all files and push for a fast merge to avoid merge conflicts from emerging.


map_gen/shared/map_loader.lua
```diff
@@ -0,0 +1,27 @@
...
+
+if #config.terrain_modules > 0 then
+    shape = shape or b.full_shape
+
+    for _, m in ipairs(config.terrain_modules) do
+        shape = b.overlay_tile_land(shape, m)
+    end
+end
+
+if shape then
```
grilledham: Change this to `if shape ~= nil then`
if you require a file that return `nil`, require changes the return to `true`.
Currently diggy returns nil and breaks this.

config.lua
```diff
@@ -1,4 +1,4 @@
-_DEBUG = false
+_DEBUG = true
```
grilledham: If I had a penny for every time I've done this ...

config.lua
```diff
@@ -16,6 +16,21 @@ global.config = {
         -- New Scenario Features, appears in the "What's new" tab
         new_info_key = 'Nothing is new. The world is at peace'
     },
+    -- map generation settings for redmew's maps
+    map_generation = {
+        -- whether to regen decoratives
+        ['regen_decoratives'] = false,
+        -- the number of tiles to change per tick
```
grilledham: Not entirely accurate, it's the number of 'tiles' that are calculated per tick. All the tiles are placed in one tick, followed by hidden_tiles, entities, decorative and then a re chart on following ticks.

config.lua
```diff
@@ -16,6 +16,21 @@ global.config = {
         -- New Scenario Features, appears in the "What's new" tab
         new_info_key = 'Nothing is new. The world is at peace'
     },
+    -- map generation settings for redmew's maps
```
grilledham: Might be worth mentioning that these settings only matter for maps that use 'shapes'

grilledham (commented):
A few changes needed, but this mostly looks fine to me.

map_gen/shared/map_loader.lua
mheguy: Wait.. but that doesn't fix that.. because `true` isn't equal to nil lol

grilledham: I swear I updated my comment to `if type(shape) == 'function' then`
I guess it got lost in a page reload or something

grilledham (approved):

PR #644, opened 2019-01-10
Crashsite 0.4

grilledham:


features/player_stats.lua
```diff
@@ -149,14 +149,20 @@ end
 local function entity_damaged(event)
     local entity = event.entity
     if entity.valid and entity.type == 'player' then -- player taking damage
-        local index = entity.player.index
-        player_damage_taken[index] = player_damage_taken[index] + event.final_damage_amount
+        local player = entity.player
+        if player and player.valid then
```
mheguy, line 153: I'm curious if this is to address an actual case or if it's theoretical.
My understanding was that the API would never return an invalid player (or any invalid LuaObject).

Valansch: For example:
- player throws grenade
- on_capsule_used is dispatched
- handler bans player and calls game.remove_offline_players({player})
- grenade lands
- entity damage is dispatched
- event.entity.player is invalid, your code falls to pieces because you assumed the API would never return an invalid player :P

And this isn't even the case of another handler/mod messing with the entity before your own handler is called. Checking LuaObject.valid is never a bad idea :P

map_gen/presets/crash_site.lua
```diff
@@ -579,6 +723,14 @@ Global.register_init(
...
 
+--[[ local Event = require 'utils.event'
+Event.add(
+    defines.events.on_player_joined_game,
+    function(event)
+        game.players[event.player_index].character = nil
+    end
+) ]]
+
 return function(x, y, world)
```
Valansch, line 735: wont `return map` do the same?

map_gen/presets/crash_site/entity_died_events.lua
```diff
@@ -37,30 +41,71 @@ local spill_items =
 )
 
 local entity_spawn_map = {
-    ['medium-biter'] = 'small-worm-turret',
-    ['big-biter'] = 'medium-worm-turret',
-    ['behemoth-biter'] = 'big-worm-turret'
+    ['medium-biter'] = {name = 'small-biter', count = 2, chance = 1},
```
linaori, line 46: Should we aim to make this compatible with hydra mode?

features/player_stats.lua
grilledham: Yeah this fixes errors I ran into while testing. 
It was the case of a character not being attached to a player while dealing and taking damage. The character had personal laser turrets.

map_gen/presets/crash_site.lua
grilledham: No, because at the time the module returns map is nil. map is set to a function in on_init and on_load. Which is called before on_chunk_generated, which is when the function is needed. This gets around the problem by returning a dummy function that has a reference to the real function.

map_gen/presets/crash_site/entity_died_events.lua
grilledham: My personal thoughts on this is that crashsite needs a lot of control over what happens when entities die and so hydra might be the wrong level of abstraction for crashsite.

mheguy (approved):
Based on the lack of errors/reported issues from the game running live, and the fact that this touches nearly nothing outside the scope of crash site itself, I'm comfortable giving this a passing review.

mheguy:

PR #647, opened 2019-01-11
Remove math and table as globals.

mheguy:
Create exceptions for table.lua and math.lua to modify respective globals and add returns for respective globals.


Valansch (approved):
Good stuff. Thank you

grilledham:
LGTM 👍

PR #648, opened 2019-01-11
Fix bad copy paste in player_rewards

mheguy:
Must have been an errant copy+paste or other restructuring error, because it makes no sense.


mheguy:
Stickler got stuck, travis is because it scans the entire repo.
No lint warnings in vscode using luacheck.

mheguy:
Val would merge but can't from mobile. :D

PR #652, opened 2019-01-14
Loaders

grilledham:
The loaders in the base game that are normally disabled actually have crafting recipes. We could enabled them and have them craftable. 

This PR unlocks the loaders when logistics techs are researched.
```lua
local technology_map = {
    ['logistics'] = 'loader',
    ['logistics-2'] = 'fast-loader',
    ['logistics-3'] = 'express-loader'
}
```

Also it removed loaders from the market.

image
image
image


linaori:
I always hate them because they have no dedicated sprites :cry: 

But :+1: on adding them

mheguy:
Love the idea and it's a definite +1 from me. What would you think of moving it into `redmew_qol.lua`?

grilledham:
If you want to move it into `redmew_qol.lua` that is fine by me.

mheguy (approved):

PR #654, opened 2019-01-16
Terraforming danger ore

grilledham:
I don't know if I consider this finished yet, but it makes changes to generate which plague006 may wish to consider. 

I only made the minimum changes I needed to get things working, so I'm open to making more changes or merging this before making other changes.


map_layout.lua
```diff
@@ -76,10 +77,11 @@ local tiles_per_tick = 32
 --shape = require "map_gen.presets.world_map"
 --shape = require "map_gen.presets.lines_and_squares"
 --shape = require "map_gen.presets.spiral_of_spirals"
---shape = require "map_gen.presets.crash_site"
+--shape = require 'map_gen.presets.crash_site'
 --shape = require "map_gen.presets.dino_island"
 --shape = require "map_gen.presets.toxic_jungle"
 --shape = require "map_gen.presets.danger_ores"
+shape = require 'map_gen.presets.terraforming_danger_ores'
```
mheguy: Missing comment dashes

grilledham: fixed. If I had a penny for every time I've made that mistake 😄

mheguy:
Clean up linting (can address the global issue by adding the following to the tops of the affected files: `--luacheck: ignore global MAP_LAYOUT_REGISTER`), open issue for static/non-runtime-modifiable config and this is good to merge IMO

mheguy (approved):

PR #655, opened 2019-01-16
Added a Lazy Bastard feature with player limit for market items

linaori:
**_Untested in multi player!_**

This adds:
 - Lazy Bastard mode, where all items you place are removed by ghosts and you have no pick axes
 - Player limits on market items
 - Support for free market items
 - Built-in support for the `item_package` item type in markets

In the future we can add a function to the Retailer to get or change the amount of items a player can still buy.

Available, free item with player limit:
image

Player bought the item (limit is per player):
image

Stack limit = 3, player limit = 10
image

Stack limit = 10, player limit = 10
image

### DIFF
--- a/features/lazy_bastard.lua
+++ b/features/lazy_bastard.lua
@@ -149,6 +149,8 @@ Command.add('lazy-bastard-bootstrap', {
     local power_source = create_entity({name = 'hidden-electric-energy-interface', position = pos})
     power_source.electric_buffer_size = 30000
     power_source.power_production = 30000
+    power_source.destructible = false
+    power_source.minable = false
 
     -- in 7 minutes, remove the power source
     Task.set_timeout(420, clean_energy_interface, {entity = power_source})
--- a/features/lazy_bastard.lua
+++ b/features/lazy_bastard.lua
@@ -1,4 +1,5 @@
 local Event = require 'utils.event'
+local Game = require 'utils.game'
 local Retailer = require 'features.retailer'
 local insert = table.insert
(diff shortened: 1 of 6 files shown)


linaori:
There's only a bootstrapping problem. You can't place a roboport to start, any suggestions?

grilledham:
spawn one by command, or as part of the map.
Or give personal robo ports as the starting items

linaori:
Can now also use the following functions to manipulate item limits per player:
```lua
---Returns the configured market group item limit or -1 if there is none.
function Retailer.get_item_limit(market_group, item_name)

---Returns the remaining market group item limit or -1 if there is none for a given player.
function Retailer.get_player_item_limit(market_group, item_name, player_index)

---sets the configured market group item limit for a given player
function Retailer.set_player_item_limit(market_group, item_name, player_index, new_limit)
```

features/lazy_bastard.lua
```diff
@@ -133,4 +145,11 @@ Command.add('lazy-bastard-bootstrap', {
     for _ = 1, bot_count do
         create_entity({name = 'construction-robot', force = force, position = pos})
     end
+
+    local power_source = create_entity({name = 'hidden-electric-energy-interface', position = pos})
+    power_source.electric_buffer_size = 30000
+    power_source.power_production = 30000
```
grilledham, line 151: adding a `power_source.destructible = false` might be worth it just in case

linaori: I've also added minable = false, just in case 😄

PR #656, opened 2019-01-17
Add features to redmew_surface, change maps to use redmew_surface

mheguy:
c05c741 for changes to redmew_surface

Added the ability for maps to set spawn locations, the tiles that "spawn islands" are made from, and improved doc of what `first_player_position_check_override` does.

Shifted applicable maps to make use of redmew_surface's feature set to control world generation.

Cleaned up linting warnings along the way.


map_gen/combined/tetris/control.lua
```diff
@@ -68,7 +69,7 @@ local machine = StateMachine.new(states.voting)
 
 local player_zoom = {}
 local player_force = nil
-local nauvis = nil
+local play_surface = nil
```
grilledham, line 72: Tetris is broken. I think `player_force` and `play_surface` should be in global.
image
Steps to reproduce.
1. Load scenario in debug mode.
2. Save game.
3. Load game.
4. Wait for error message.

map_gen/presets/connected_dots.lua
```diff
@@ -11,13 +13,13 @@ local ore_block_size = 32
...
-    end
-
-    return tile
-end
+RS.set_map_gen_settings(
+    {
+        MGSP.ore_oil_none,
+        MGSP.cliff_none,
+        MGSP.enemy_none
+    }
```
grilledham, line 20: Connected_dots can't make use of the enemy_none preset as the no_enemies effect is only applied to part of the map, not the whole map.

map_layout.lua
```diff
@@ -40,8 +34,8 @@ local tiles_per_tick = 32
 --shape = require "map_gen.presets.maori"
 --shape = require "map_gen.presets.goat"
 --shape = require "map_gen.presets.GoT"
---shape = require "map_gen.presets.turkey" -- needs to be rebuilt from missing source image.
---shape = require "map_gen.presets.north_america" -- needs to be rebuilt from missing source image.
+--shape = require "map_gen.presets.turkey"
+--shape = require "map_gen.presets.north_america"
```
grilledham: I think the comments are still correct, the preset data for the maps need rebuilding.

grilledham:
I read through all the changes. Apart form the comments I made, nothing jumped out to me as wrong.

map_gen/combined/tetris/control.lua
mheguy: This *might* be more than just an issue with a map.

I tried the following:
placing `player_force` and `play_surface` into the global register

placing `player_force` and `play_surface` inside a table and placing that table into the global register

In both of those cases, loading the save led to nil values.

I tried changing instanced of `player_force` and `play_surface` to `global.` and in that case, there were no issues.

linaori: `play_surface` can be stored here as local cache. Meaning something like this:
```lua
if play_surface == nil then
    play_surface = RS.get_surface()
end
```
This means it won't have to be stored in globals and it will set the state the first time it hits this, reducing the calls to the `game` object and having a low memory footprint (I suppose).

grilledham (approved):
As Tetris working is beyond the scope of this PR, I approve.

map_gen/combined/tetris/control.lua
grilledham: The comment for `RS.get_surface()` says it shouldn't be called outside of events.
But putting `play_surface = RS.get_surface()` inside the callback in `Global.register` should work.

PR #660, opened 2019-01-19
Add seeds command

mheguy:
List the seeds of all surfaces -- if you guys think this is too niche a command (it may well be) I'll withdraw the PR and continue using in-game lua to check. :)

### DIFF
--- a/features/redmew_commands.lua
+++ b/features/redmew_commands.lua
@@ -7,6 +7,7 @@ local Command = require 'utils.command'
 
 local format = string.format
 local ceil = math.ceil
+local concat = table.concat
 
 --- Kill a player with fish as the cause of death.
 local function do_fish_kill(player, suicide)
@@ -191,6 +192,22 @@ local function search_command(arguments)
     p(format('-------- Page %d / %d --------', page, pages))
 end
 
+local function list_seeds()
+    local seeds = {}
+    local count_of_seeds = 0
+    for _, surface in pairs(game.surfaces) do
+        seeds[count_of_seeds + 1] = surface.name
+        seeds[count_of_seeds + 2] = ': '
+        seeds[count_of_seeds + 3] = tostring(surface.map_gen_settings.seed)
+        count_of_seeds = count_of_seeds + 4
+        seeds[count_of_seeds] = ', '
+    end
+
+    seeds[#seeds] = nil
+    seeds = concat(seeds)
+    Game.player_print(seeds)
+end
+
 -- Command registrations
 
 Command.add(
@@ -257,6 +274,15 @@ Command.add(
         allowed_by_server = true,
     },
     search_command
+)
+
+Command.add(
+    'seeds',
+    {
+        description = 'List the seeds of all surfaces',
+        allowed_by_server = true,
+    },
+    list_seeds
 )
 
 -- Commands with no functions, only calls to other modules


features/redmew_commands.lua
```diff
@@ -191,6 +191,14 @@ local function search_command(arguments)
     p(format('-------- Page %d / %d --------', page, pages))
 end
 
+local function list_seeds()
+    local seeds = ''
+    for _, surface in pairs(game.surfaces) do
+        seeds = format('%s: %s\n%s', surface.name, surface.map_gen_settings.seed, seeds)
```
grilledham: I'm pretty sure the `\n` gets ignored by player.print, or turned into ' ' so maybe just use a space character as it's clearer what this code will do.

features/redmew_commands.lua
```diff
@@ -191,6 +192,15 @@ local function search_command(arguments)
...
 end
 
+local function list_seeds()
+    local seeds
+    for _, surface in pairs(game.surfaces) do
+        local seed = format('%s: %s', surface.name, surface.map_gen_settings.seed)
+        seeds = concat({seed, seeds}, ', ')
+    end
+    Game.player_print(seeds)
+end
```
grilledham, line 209: Performance doesn't really matter for this command, but if you wanted to do the string concatenation more efficiently you would do something like:
```lua
local function list_seeds()
    local seeds = {}
    for _, surface in pairs(game.surfaces) do
        seeds[#seeds + 1] = surface.name
        seeds[#seeds + 1] = ': '
        seeds[#seeds + 1] = tostring(surface.map_gen_settings.seed)
        seeds[#seeds + 1] = ', '
    end

    seeds[#seeds] = nil
    seeds = concat(seeds)

    Game.player_print(seeds)
end
```

linaori: Would do `count_of_seeds +1, count_of_seeds + 2` etc, saves you from counting the seed size every time if you know how much it's going to be

plague006: You're both right that this could and should be better performance-wise: if it weren't a command that nobody is ever going to run. :D

mheguy: Let us maximize our efficiency in a command that will run less than 100 times per year. :D

grilledham: The key to micro optimisations is to start small.

features/redmew_commands.lua
```diff
@@ -193,11 +193,19 @@ local function search_command(arguments)
...
-    local seeds
+    local seeds = {}
+    local count_of_seeds = 1
     for _, surface in pairs(game.surfaces) do
-        local seed = format('%s: %s', surface.name, surface.map_gen_settings.seed)
-        seeds = concat({seed, seeds}, ', ')
+        seeds[count_of_seeds] = surface.name
+        seeds[count_of_seeds] = ': '
+        seeds[count_of_seeds] = tostring(surface.map_gen_settings.seed)
+        seeds[count_of_seeds] = ', '
```
grilledham, line 203: Bad plague! You need to increment the count after every insert.

features/redmew_commands.lua
plague006: Bad grilled! You need to increment the count after every insert.

features/redmew_commands.lua
mheguy: Ugh. This optimization joke went too far. :(

grilledham (approved):

features/redmew_commands.lua
linaori: If we're nitpicking on performance:
```lua
local function list_seeds()
    local seeds = {}
    local count_of_seeds = 0
    for _, surface in pairs(game.surfaces) do
        seeds[count_of_seeds + 1] = surface.name
        seeds[count_of_seeds + 2] = ': '
        seeds[count_of_seeds + 3] = tostring(surface.map_gen_settings.seed)
        count_of_seeds = count_of_seeds + 4
        seeds[count_of_seeds] = ', '
    end

    seeds[#seeds] = nil -- what is this line for?
    seeds = concat(seeds)

    Game.player_print(seeds)
end
```

mheguy: Removing the trailing comma

PR #663, opened 2019-01-20
Removed untracked stats

linaori:
These values are no longer updated, this removes the remaining references


grilledham:
I think I took care of this at the same time with 
https://github.com/Refactorio/RedMew/commit/1981dd6c0fd782f8c79e7e2ab05decbe6315598a

linaori:
👍

PR #664, opened 2019-01-20
Added grid_bot_islands map preset

Jayefuu:
Added map_gen/presets/grid_bot_islands map 
Moved entity_drop_amount from market.lua to config.lua
Added entity_drop_amount.chance to table to allow map makers to configure coin drop chances for biters, spitters, worms and nests from within map presets.


grilledham (approved):
Added an early return if `chance == 0`
Rest looks good to me 👍

mheguy:

PR #667, opened 2019-01-21
Corpse looting message

grilledham:
Print a message when players loot other players corpses.
Messages are printed to game, the logs and discord.

Also I changed the format of the color presets so I get auto complete in vs code.


linaori (approved):

mheguy:
Tested well on server, code looks fine. I like the change to color.
On an unrelated note I've noticed VSCode is auto-completing require paths now and that's *dope*.

PR #671, opened 2019-01-21
Fix walkabout

### LINKED ISSUE #666: Fix walkabout
The issue currently is that the players' details (their former force, character, etc) are only entered into the global table when the token is called (after the timeout) and so if someone leaves and returns within the time of their timeout, their details are lost since `clean_on_join` removes their walkabout status.

Needless to say: I want to change a lot about the way I "fixed" things then. Why did I make a second table in `global`? Only past me knows.

mheguy:
Not returning player from walkabout if they rejoin before timer expired, update refs to global.walking

Closes #666

### DIFF
--- a/features/redmew_commands.lua
+++ b/features/redmew_commands.lua
@@ -4,6 +4,8 @@ local Game = require 'utils.game'
 local Server = require 'features.server'
 local Timestamp = require 'utils.timestamp'
 local Command = require 'utils.command'
+local Walkabout = require 'features.walkabout'
+local redmew_version = require 'resources.version'
 
 local format = string.format
 local ceil = math.ceil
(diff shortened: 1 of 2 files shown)


grilledham:
Not tempted to move the global variables into Global?

mheguy:
Oddly enough: hadn't thought about it and just kinda took it as it was. But yeah that would make a lot more sense.

features/redmew_commands.lua
```diff
@@ -38,7 +38,7 @@ local function kill(args, player)
         end
     end
 
-    if global.walking and ((player and global.walking[player.index]) or (target and global.walking[target.index])) then
+    if global.walkabout and ((player and global.walkabout[player.index]) or (target and global.walkabout[target.index])) then
```
grilledham: Didn't these get moved into Global?

grilledham (dismissed):

linaori (approved):
Changes look alright, I've never used the function itself so I can't really comment on that

grilledham (approved):

PR #672, opened 2019-01-21
Score update

grilledham:
Some updates to the score gui.
- Added total coins spent
- Changed the styles
- Removed global to track rockets launched as you can get that from `force.get_item_launched`

### DIFF
--- a/features/gui/score.lua
+++ b/features/gui/score.lua
@@ -158,9 +158,9 @@ local function rocket_launched(event)
 
         game.print(message)
         Server.to_discord_bold(message)
-
-        refresh_score()
     end
+
+    refresh_score()
 end
 
 Gui.on_click(
--- a/features/gui/score.lua
+++ b/features/gui/score.lua
@@ -135,8 +135,9 @@ local function score_show(top)
 end
 
 local function rocket_launched(event)
-    local entity = event.entity
-    if not entity or not entity.valid then
+    local entity = event.rocket
+
+    if not entity or not entity.valid or not entity.force == 'player' then
         return
     end
 
@@ -151,12 +152,15 @@ local function rocket_launched(event)
     end
 
     count = game.forces.player.get_item_launched('satellite')
-    local message = 'A satellite has been launched! Total count: ' .. count
 
-    game.print(message)
-    Server.to_discord_bold(message)
+    if (count < 10) or ((count < 50) and ((count % 5) == 0)) or ((count % 25) == 0) then
+        local message = 'A satellite has been launched! Total count: ' .. count
 
-    refresh_score()
+        game.print(message)
+        Server.to_discord_bold(message)
+
+        refresh_score()
+    end
 end
 
 Gui.on_click(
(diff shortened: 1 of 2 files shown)


features/gui/score.lua
```diff
@@ -71,85 +58,100 @@ local function score_label_style(label, color)
...
+    end
+
+    local count = inventory.get_item_count('satellite')
+    if count == 0 then
+        return
+    end
 
-    local message = 'A rocket has been launched! Total count: ' .. count
+    count = game.forces.player.get_item_launched('satellite')
+    local message = 'A satellite has been launched! Total count: ' .. count
```
mheguy: I don't know if it's still bad since I don't really play anymore.. but I remember that on high rocket launch count maps the launch spam is kind of a nuisance.

How would you feel about something like:
```lua
 if count < 10 or 
((count % 5) == 0 and count < 50) or 
((count % 25) == 0) then
print
```
So plopping messages from 1-10,15,20,25,30,35,40,50 and every 25 after that.

features/player_stats.lua
```diff
@@ -204,11 +207,16 @@ function Public.get_coin_spent(player_index)
 end
 
 function Public.set_coin_spent(player_index, value)
```
plague006: Curious if you have any kind of hypothetical use case for this or for `set_coin_earned`

mheguy, line 209: Curious if you have any kind of hypothetical use case for this or for `set_coin_earned`

features/gui/score.lua
```diff
@@ -71,85 +58,100 @@ local function score_label_style(label, color)
...
+        caption = concat {'Coins spent: ', PlayerStats.get_total_coins_spent(), ' '}
+    }
+    score_label_style(label, Color.yellow)
 end
 
-local function rocket_launched()
-    local count = data.rockets_launched + 1
-    data.rockets_launched = count
+local function rocket_launched(event)
+    local entity = event.entity
```
plague006: `event.rocket`

features/gui/score.lua
```diff
@@ -71,85 +58,100 @@ local function score_label_style(label, color)
...
+        return
+    end
+
+    local count = inventory.get_item_count('satellite')
+    if count == 0 then
+        return
+    end
 
-    local message = 'A rocket has been launched! Total count: ' .. count
+    count = game.forces.player.get_item_launched('satellite')
```
mheguy, line 154: If we're only using `player` force launches, we should check that the `player` force launched it.

features/gui/score.lua
mheguy: `event.rocket`

mheguy (changes requested):

features/player_stats.lua
grilledham: I added it first when I added all the getter and setters. I later realised that change_coin_spent is a nice to have convenient function.

features/gui/score.lua
grilledham: god dammit.
I thought I tested this, but clearly not 😄

grilledham:
LGTM thanks for making the changes 👍

PR #674, opened 2019-01-23
Deal with negative numbers in a sane way in player rewards

mheguy:


utils/player_rewards.lua
```diff
@@ -2,16 +2,31 @@ local Global = require 'utils.global'
 local Game = require 'utils.game'
 local PlayerStats = require 'features.player_stats'
 local Command = require 'utils.command'
+
 local format = string.format
+local abs = math.abs
+local concat = table.concat
 
 local Public = {}
 local reward_token = {global.config.player_rewards.token} or {global.config.market.currency} or {'coin'}
```
grilledham, line 11: Why is `reward_token` in a table?
```lua
local reward_token = global.config.player_rewards.token or global.config.market.currency or 'coin'
```
And you can get rid of the `Global.register`

grilledham:
The comment for `Public.give_reward` says `-- @param amount <number> the amount of reward tokens to remove`

utils/player_rewards.lua
mheguy: In a table so that changing the reward via `Public.set_reward` during runtime doesn't cause desyncs.
Edit: in a table so that it can be entered into a token*

grilledham:
Rest looks good to me

linaori (approved):

grilledham (approved):

PR #675, opened 2019-01-23
Added sierpinski_carpet map preset

Jayefuu:


map_gen/presets/sierpinski_carpet.lua
```diff
@@ -0,0 +1,152 @@
...
+
+-- Create the Sierpinski carpet shape
+local function grid(x, y)
+    return not(x %3 <2  or y %3 <2)
+end
+
+grid = b.translate(grid,2,2)
+local sierpinski = grid
+
+for i = 1, 10, 1 do
```
mheguy, line 45: Just for the future, adding 1 is the default behaviour. So `for i = 1, 10 do` does exactly the same thing.

mheguy:

PR #676, opened 2019-01-23
Added map previews of maps by Jayefuu

Jayefuu:


mheguy:
Awesome, thank you. :D

PR #677, opened 2019-01-23
Action warnings

mheguy:
Just a little way to keep our messages consistent and make it a little quicker and easier to send messages.

### DIFF
--- a/utils/core.lua
+++ b/utils/core.lua
@@ -3,6 +3,7 @@
 -- Dependencies
 local Game = require 'utils.game'
 local Color = require 'resources.color_presets'
+local Server = require 'features.server'
 
 -- localized functions
 local random = math.random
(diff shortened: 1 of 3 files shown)


utils/core.lua
```diff
@@ -187,6 +201,27 @@ function Module.set_and_return(tbl, key, value)
     return value
 end
 
+--- Takes msg and prints it to all players. Also prints to the log and discord
+-- @param msg <string> The message to print
+-- @param warning_prefix <string> The name of the module/warning
+function Module.action_warning(msg, warning_prefix)
```
grilledham: In my mind it would make more sense to change the function signature to
```lua
function Module.action_warning(warning_prefix, msg)
```
As the prefix comes first in the message.
Also same change to function below.

grilledham (approved):

PR #678, opened 2019-01-23
Add revive ghosts

mheguy:
Adds a command so that admins can build/rebuild more easily after a griefer attacks something that can be restored via a blueprint.

### DIFF
--- a/features/admin_commands.lua
+++ b/features/admin_commands.lua
@@ -241,6 +241,19 @@ local function teleport_command(args, player)
     end
 end
 
+--- Revives ghosts around the player
+local function revive_ghosts(args, player)
+    local radius = args.radius
+    local pos = player.position
+    for _, e in pairs(player.surface.find_entities_filtered {area = {{pos.x - radius, pos.y - radius}, {pos.x + radius, pos.y + radius}}}) do
+        if e.type == 'entity-ghost' then
+            e.revive()
+        else
+            e.health = 10000
+        end
+    end
+end
+
 -- Event registrations
 
 Event.add(defines.events.on_built_entity, built_entity)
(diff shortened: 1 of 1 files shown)


features/admin_commands.lua
```diff
@@ -241,6 +241,19 @@ local function teleport_command(args, player)
     end
 end
 
+--- Revives ghosts around the player
+local function revive_ghosts(args, player)
+    local radius = args.radius
+    local pos = player.position
+    for _, e in pairs(player.surface.find_entities_filtered {area = {{pos.x - radius, pos.y - radius}, {pos.x + radius, pos.y + radius}}}) do
+        if e.type == 'entity-ghost' then
```
grilledham: Can you not pass in `type = 'entity-ghost'` for `find_entities_filtered`?

mheguy: I ended up changing the scope at last minute and didn't follow through on all aspects.

The basic intent is as a counter to grenade griefers.

features/admin_commands.lua
```diff
@@ -241,6 +241,19 @@ local function teleport_command(args, player)
...
 
+--- Revives ghosts and repairs entities around the player
+local function revive_and_repair(args, player)
+    local radius = args.radius
+    local pos = player.position
+    for _, e in pairs(player.surface.find_entities_filtered {area = {{pos.x - radius, pos.y - radius}, {pos.x + radius, pos.y + radius}}, force = player.force}) do
+        if e.type == 'entity-ghost' then
+            e.revive()
+        else
+            e.health = 10000
```
grilledham: I can't help shake the feeling that setting health on entities that don't support it can error.
From <https://lua-api.factorio.com/latest/LuaEntity.html>
>> Health of the entity. Setting health to less than 0 will set health to 0, entities with 0 health can not be attacked.  Setting health to higher than max health will set health to max health.
**Note: If used on an entity that doesn't support health, this field will be nil.**

I'm not sure if that applies here, but it might be worth playing it safe.

linaori (approved):

grilledham (approved):

PR #679, opened 2019-01-24
Initial Creation of Rails Only Scenario

R-Nukem:
adds the Rails Only map config to map_layout.lua and the corresponding file in map_gen/misc/

This map disables placement of all belts and roboports, forcing players to use rail systems for everything.


mheguy (changes requested):
If you'd like I can go line by line but I'll give the broad strokes first:
Maps go into `presets`, `misc` is for gameplay modules/features.
I believe you created a map because you set a map name and description.

You're copying nearly entire modules into your map.
Instead of copy pasting all of danger_ores into your map, just require danger ores and change `global.banned_entites` to suit your needs.

For the command, as a convention we use dashes for spaces (ex. `rail-market`)

For your event: you can only tie 1 function to each event definition. So if you use `script.on_event(defines.events.on_built_entity` no other module can hook onto that event. For that purpose, we have `utils.event` which lets you hook multiple events onto it. Ex: `Event.add(defines.events.on_built_entity`

If you need a hand or pointers, feel free to ping me on discord. :)

grilledham:
They didn't copy all of `danger_ore_banned_entities` they removed the part that checks if the entity is placed on ore. 
I think a misc module for build_banned_entities and then a preset for the rest would be the way to go.

map_gen/misc/rails_only_banned_entities.lua
```diff
@@ -0,0 +1,117 @@
...
+        end
+        if not global.banned_entites[name] then
+            return
+        end
+        -- Some entities have a bounding_box area of zero, eg robots.
+        local area = entity.bounding_box
+        local left_top, right_bottom = area.left_top, area.right_bottom
+        if left_top.x == right_bottom.x and left_top.y == right_bottom.y then
+            return
+        end
```
grilledham: Can probably remove the part that checks if the area is empty. The area was needed in the danger ore version to scan the surface for ores.

plague006:
On hold waiting for https://github.com/Refactorio/RedMew/pull/723

mheguy:
I don't think those had the effect you intended.

Either way, this is on hold waiting for https://github.com/Refactorio/RedMew/pull/723

PR #680, opened 2019-01-24
Builder efficiency!

mheguy:
Split into two commits for easier review:
68a6c9a - find and replace to localize functions
4ec8762 - replace insert() with tbl[#tbl + 1] = blah

I also think this works, but wasn't 100% sure and didn't want to pull the trigger on it.

```lua
function Builders.prepare_weighted_array(array)
    local total = 0
    local weights = {}
    for i = 1, #array do
        total = total + array[i].weight
        weights[i] = total
    end

    weights.total = total

    return weights
end
```

### DIFF
--- a/map_gen/shared/builders.lua
+++ b/map_gen/shared/builders.lua
@@ -11,7 +11,6 @@ local sin = math.sin
 local cos = math.cos
 local atan2 = math.atan2
 local tau = math.tau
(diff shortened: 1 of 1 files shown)


linaori (dismissed):

mheguy:
Rebase

PR #682, opened 2019-01-24
Add ability to modify changelog in-game

mheguy:
And update the changelog hourly(ish)

### DIFF
--- a/features/gui/info.lua
+++ b/features/gui/info.lua
@@ -717,8 +717,7 @@ Event.add(defines.events.on_player_created, player_created)
 
 Event.add(Server.events.on_server_started, download_changelog)
 
--- Hourly
-Event.on_nth_tick(215983, download_changelog)
+Server.on_data_set_changed('misc', download_changelog)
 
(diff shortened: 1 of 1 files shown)


grilledham:
Instead of updating the change log hourly you can register an event for when it changes.
https://github.com/Refactorio/RedMew/blob/develop/features/server.lua#L336

features/gui/info.lua
```diff
@@ -717,8 +717,7 @@ Event.add(defines.events.on_player_created, player_created)
 
 Event.add(Server.events.on_server_started, download_changelog)
 
--- Hourly
-Event.on_nth_tick(215983, download_changelog)
+Server.on_data_set_changed('misc', download_changelog)
```
grilledham: shouldn't that be
```lua
Server.on_data_set_changed('misc', changelog_callback)
```
instead?

mheguy: Yes and no? I split the function from the token registration, so now everything seems to play together nicely.

grilledham (approved):

PR #686, opened 2019-01-25
[Diggy] Restoring buffs every 61 ticks

### LINKED ISSUE #673: Diggy Level up buffs vanish.
**Describe the bug**
*Note, Single player*
Buffs from leveling up the market (Health, Bonus inventory slots and mining speed bonus) will vanish.

**Expected behavior**
Buffs will persist or be refreshed or re-added if lost.

**Steps to reproduce**
1. Start Diggy in single player
2. Any point after leveling up, save & exit.
3. Add or remove [a] mod[s] (As far as I can tell ANY mod change will trigger this)
4. When save is reloaded, buffs will be absent.

*Edit, removing mods doesn't all ways seem to make the buffs vanish. All test with adding mods did make the buffs vanish.*

**Screenshots**

**Map**
Which map you were playing/which server you were playing on when you found the bug.
Single player Diggy scenario.

**Additional context**

SimonFlapse:
Suggested solution for #673 

Updates all diggy buffs every 61 ticks. If they have been set to zero, restore the values to the active_modifier value.

This only fixes issues caused by mods changing the force modifiers to zero. 

Closes #673

### DIFF
--- a/map_gen/maps/diggy/feature/experience.lua
+++ b/map_gen/maps/diggy/feature/experience.lua
@@ -64,21 +64,21 @@ local gain_xp_color = Color.light_sky_blue
 local lose_xp_color = Color.red
 local unlocked_color = Color.black
 local locked_color = Color.gray
-local table_column_layout = {type = 'table', column_count = 2}
+local table_column_layout = { type = 'table', column_count = 2 }
 
-local level_up_formula = (function (level_reached)
+local level_up_formula = (function(level_reached)
     local difficulty_scale = floor(config.difficulty_scale)
     local level_fine_tune = floor(config.xp_fine_tune)
     local start_value = (floor(config.first_lvl_xp))
     local precision = (floor(config.cost_precision))
     local function formula(level)
         return (floor(
-            (1.15 ^ (level * 0.1))
-            + difficulty_scale * (level) ^ 3
-            + level_fine_tune * (level) ^ 2
-            + start_value * (level)
-            - difficulty_scale * (level)
-            - level_fine_tune * (level))
+                (1.15 ^ (level * 0.1))
+                        + difficulty_scale * (level) ^ 3
+                        + level_fine_tune * (level) ^ 2
+                        + start_value * (level)
+                        - difficulty_scale * (level)
+                        - level_fine_tune * (level))
         )
     end
     local value = formula(level_reached + 1)
@@ -100,9 +100,9 @@ local function calculate_level_xp(level)
     if level_table[level] == nil then
         local value
         if level == 1 then
-            value = level_up_formula(level-1)
+            value = level_up_formula(level - 1)
         else
-            value = level_up_formula(level-1)+calculate_level_xp(level-1)
+            value = level_up_formula(level - 1) + calculate_level_xp(level - 1)
         end
         insert(level_table, level, value)
     end
@@ -112,7 +112,7 @@ end
 ---@param level number a number specifying the current level
 ---@return number a percentage of the required experience to level up from one level to the other
 local function percentage_of_level_req(level, percentage)
-    return level_up_formula(level)*percentage
+    return level_up_formula(level) * percentage
 end
 
 ---Updates the market contents based on the current level.
@@ -139,7 +139,7 @@ function Experience.update_mining_speed(force, level_up)
         level_up = level_up ~= nil and level_up or 0
         if level_up > 0 and buff ~= nil then
             local level = get_force_data(force).current_level
-            local adjusted_value = floor(max(buff.value, 24*0.9^level))
+            local adjusted_value = floor(max(buff.value, 24 * 0.9 ^ level))
             local value = (buff.double_level ~= nil and level_up % buff.double_level == 0) and adjusted_value * 2 or adjusted_value
             mining_efficiency.level_modifier = mining_efficiency.level_modifier + (value * 0.01)
         end
@@ -185,7 +185,7 @@ function Experience.update_health_bonus(force, level_up)
     if buff.max == nil or force.character_health_bonus < buff.max then
         level_up = level_up ~= nil and level_up or 0
         if level_up > 0 and buff ~= nil then
-            local value = (buff.double_level ~= nil and level_up%buff.double_level == 0) and buff.value*2 or buff.value
+            local value = (buff.double_level ~= nil and level_up % buff.double_level == 0) and buff.value * 2 or buff.value
             health_bonus.level_modifier = health_bonus.level_modifier + value
         end
 
@@ -226,7 +226,7 @@ local function on_player_mined_entity(event)
         return
     end
 
-    print_player_floating_text_position(player_index, format('+%s XP', exp), gain_xp_color,0, -0.5)
+    print_player_floating_text_position(player_index, format('+%s XP', exp), gain_xp_color, 0, -0.5)
     add_experience(force, exp)
 end
 
@@ -254,7 +254,6 @@ local function on_research_finished(event)
         print_player_floating_text_position(player_index, text, gain_xp_color, -1, -0.5)
     end
     add_experience(force, exp)
-
 
     local current_modifier = mining_efficiency.research_modifier
     local new_modifier = force.mining_drill_productivity_bonus * config.mining_speed_productivity_multiplier * 0.5
@@ -363,8 +362,8 @@ local function redraw_heading(data, header)
     Gui.clear(frame)
 
     local heading_table = frame.add(table_column_layout)
-    apply_heading_style(heading_table.add({type = 'label', caption = 'Requirement'}).style, 100)
-    apply_heading_style(heading_table.add({type = 'label', caption = header_caption}).style, 220)
+    apply_heading_style(heading_table.add({ type = 'label', caption = 'Requirement' }).style, 100)
+    apply_heading_style(heading_table.add({ type = 'label', caption = header_caption }).style, 220)
 end
 
 local function redraw_progressbar(data)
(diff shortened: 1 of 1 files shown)


linaori:
Have you had the chance to test this with a mod?

SimonFlapse:
> Have you had the chance to test this with a mod?

Not really, I don't know what else we should do about it since it seems the mods are messing about with the modifiers. However with this players can set the buffs to 0 and it'll revert back to the correct value.

map_gen/Diggy/Feature/Experience.lua
```diff
@@ -507,6 +508,12 @@ local function update_gui()
             toggle(data)
         end
     end
+
+    --Resets buffs if they have been set to 0
+    local force = players[#players].force
```
linaori: What about `game.players.force`? Seems less error-prone

grilledham: do you mean `game.forces.player`?

linaori: No, `players.game.force`

mheguy:
Rebasing + force pushing to get up to date with the lint-free dev head.

SimonFlapse:
Is this still something we'd be interested in adding to diggy? Or are we looking for another solution/no solution.

map_gen/Diggy/Feature/Experience.lua
```diff
@@ -518,6 +519,12 @@ local function update_gui()
             toggle(data)
         end
     end
+
+    --Resets buffs if they have been set to 0
+    local force = game.forces.player
+    Experience.update_inventory_slots(force, 0)
+    Experience.update_mining_speed(force, 0)
+    Experience.update_health_bonus(force, 0)
```
mheguy: Assuming another mod is constantly trying to reset the values to 0, wouldn't that just create a weird situation where a player's health is constantly fluctuating? Or worse, their inventory.

map_gen/Diggy/Feature/Experience.lua
```diff
@@ -117,8 +117,7 @@ function Experience.update_mining_speed(force, level_up)
         mining_efficiency.level_modifier = mining_efficiency.level_modifier + (value * 0.01)
     end
     -- remove the current buff
-    local old_modifier = force.manual_mining_speed_modifier - mining_efficiency.active_modifier
-
+    local old_modifier = (force.manual_mining_speed_modifier == 0) and 0 or force.manual_mining_speed_modifier - mining_efficiency.active_modifier
```
mheguy: Rather than only accounting for a force.modifier of 0, this should put a floor of 0 on `force.manual_mining_speed_modifier - mining_efficiency.active_modifier` to prevent a situation where a mod sets a number `> 0` but `< active_modifier` (as this would result in a net buff less than the amount we aimed to give).

mheguy (changes requested):
I'm not sure how to handle the situation as a whole, but we can at least prevent diggy from inflicting a debuff on its players. Edit: my mistake that we wouldn't be creating a debuff, we would just be giving less than the total buff that we should be applying.

map_gen/Diggy/Feature/Experience.lua
SimonFlapse: Sounds fair. `force.manual_mining_speed_modifier - mining_efficiency.active_modifier` should never be < 0. I think that's everything we can do/should want to do about this issue.

map_gen/Diggy/Feature/Experience.lua
SimonFlapse: Sounds fair. `force.manual_mining_speed_modifier - mining_efficiency.active_modifier` should never be < 0. I think that's everything we can do/should want to do about this issue.

map_gen/Diggy/Feature/Experience.lua
SimonFlapse: Sounds fair. `force.manual_mining_speed_modifier - mining_efficiency.active_modifier` should never be < 0. I think that's everything we can do/should want to do about this issue.

map_gen/Diggy/Feature/Experience.lua
SimonFlapse: That seems like the wrong way to do it. Diggys buffs simply would not be compatible with the mod then. We can't, and should not put too much effort into making this compatible with mods.

mheguy:
On phone so evaluating is a bit difficult but:
It seems the incorrect style is being applied in regards to this line of the style guide:
No spaces after (, [, { or before ], ), }.

In addition, I think force.modifier == 0 and 0 is redundant now that you are instituting a floor of 0.

map_gen/maps/diggy/feature/experience.lua
```diff
@@ -139,13 +139,13 @@ function Experience.update_mining_speed(force, level_up)
...
             local level = get_force_data(force).current_level
-            local adjusted_value = floor(max(buff.value, 24*0.9^level))
+            local adjusted_value = floor(max(buff.value, 24 * 0.9 ^ level))
             local value = (buff.double_level ~= nil and level_up % buff.double_level == 0) and adjusted_value * 2 or adjusted_value
             mining_efficiency.level_modifier = mining_efficiency.level_modifier + (value * 0.01)
         end
         -- remove the current buff
-        local old_modifier = force.manual_mining_speed_modifier - mining_efficiency.active_modifier
-
+        local old_modifier = (force.manual_mining_speed_modifier == 0) and 0 or force.manual_mining_speed_modifier - mining_efficiency.active_modifier
```
mheguy: Pretty sure `(force.manual_mining_speed_modifier == 0) and 0 or ` is redundant since we're covering all cases where `force.manual_mining_speed_modifier < mining_efficiency.active_modifier`

map_gen/maps/diggy/feature/experience.lua
```diff
@@ -538,6 +538,12 @@ local function update_gui()
             toggle(data)
         end
     end
+
+    --Resets buffs if they have been set to 0
+    local force = game.forces.player
+    Experience.update_inventory_slots(force, 0)
```
mheguy, line 544: As discussed, "fighting" other mods for control every second might make for a worse player experience

SimonFlapse: That should only happen if the mod keeps setting the modifier to 0. And if it does we are not at fault. People must understand that adding mods can cause incompatible and inconsistency. I still don't think there is anything else we can do. So either we go with these changes or we just close the PR and not do anything.

mheguy: I don't think we should add behaviour to combat other mods directly as we don't know if we will make things better or worse, but I think clamping the value of `old_modifier` to not be negative is a very worthwhile change.

SimonFlapse: Agree, that's a good find :) But we'd not have a solution for #673 and it would keep being brought up

mheguy: We can't control what other mods do. All we can do is try to be compatible with others, we can't force them to be compatible with others.

map_gen/maps/diggy/feature/experience.lua
mheguy:
```suggestion
        local old_modifier = force.manual_mining_speed_modifier - mining_efficiency.active_modifier
```

mheguy: The suggestion thing is neat

map_gen/maps/diggy/feature/experience.lua
```diff
@@ -167,8 +167,8 @@ function Experience.update_inventory_slots(force, level_up)
         end
 
         -- remove the current buff
-        local old_modifier = force.character_inventory_slots_bonus - inventory_slots.active_modifier
-
+        local old_modifier = (force.character_inventory_slots_bonus == 0) and 0 or force.character_inventory_slots_bonus - inventory_slots.active_modifier
```
mheguy:
```suggestion
        local old_modifier = force.character_inventory_slots_bonus - inventory_slots.active_modifier
```

map_gen/maps/diggy/feature/experience.lua
```diff
@@ -185,13 +185,13 @@ function Experience.update_health_bonus(force, level_up)
...
         if level_up > 0 and buff ~= nil then
-            local value = (buff.double_level ~= nil and level_up%buff.double_level == 0) and buff.value*2 or buff.value
+            local value = (buff.double_level ~= nil and level_up % buff.double_level == 0) and buff.value * 2 or buff.value
             health_bonus.level_modifier = health_bonus.level_modifier + value
         end
 
         -- remove the current buff
-        local old_modifier = force.character_health_bonus - health_bonus.active_modifier
-
+        local old_modifier = (force.character_health_bonus == 0) and 0 or force.character_health_bonus - health_bonus.active_modifier
```
mheguy:
```suggestion
        local old_modifier = force.character_health_bonus - health_bonus.active_modifier
```

map_gen/maps/diggy/feature/experience.lua
```diff
@@ -269,6 +268,7 @@ local function on_research_finished(event)
 
     Experience.update_inventory_slots(force, 0)
     Experience.update_mining_speed(force, 0)
+    Experience.update_health_bonus(force, 0)
```
mheguy, line 271: Why is this new? Was there a reason health bonus wasn't handled here?

SimonFlapse: It wasn't needed, since no research added extra health. I've added it for consistency, and it ensures that all the buffs are force updated on a new research as well as every 61 ticks

map_gen/maps/diggy/feature/experience.lua
SimonFlapse: Yeah that'll work now since we are preventing the ´old_modifier´ from going below 0

mheguy (approved):

PR #687, opened 2019-01-25
Extra features for popup gui

### LINKED ISSUE #636: popup not only as "Attention!" popup
I would like to be able to set the popup window to not only show an attention with a warning icon.

It would be good for me to also show "Error!" or "Success!" with different icons.

SimonFlapse:
Added 13 new icons to the popup gui
Added the ability to change title

Updated popup-update commands wording, title and icon.

Closes #636


features/gui/popup.lua
```diff
@@ -5,7 +5,28 @@ local Command = require 'utils.command'
 
 local close_name = Gui.uid_name()
 
-local function show_popup(player, message)
+local icons = {
+    "ammo_icon",                            -- 1
```
linaori: Would be nice if we could refer to them externally:
```lua
show_popup(player, 'foo', 'bar', Popup.icons.danger_icon)
```

features/gui/popup.lua
```diff
@@ -5,7 +5,28 @@ local Command = require 'utils.command'
...
+    "fuel_icon",                            -- 8
+    "no_building_material_icon",            -- 9
+    "no_storage_space_icon",                -- 10
+    "not_enough_construction_robots_icon",  -- 11
+    "not_enough_repair_packs_icon",         -- 12
+    "recharge_icon",                        -- 13
+    "too_far_from_roboport_icon"            -- 14
+}
+
+local function show_popup(player, message, title, icon_id)
```
grilledham: What if we change `icon_id` to `sprite_path` then use linaori's sugestion above. This would allow use to use any sprite not just the ones provided by the module

mheguy:
If the number of sprites is large, maybe plop them into a resource file?

features/gui/popup.lua
SimonFlapse: I've added the ability to provide a sprite path, while keeping the "shortcuts" for the 14 similar icons

mheguy:
Is https://github.com/Refactorio/RedMew/pull/687#pullrequestreview-196601458 resolved?

And when you get a chance can you clean up the lint warnings?

SimonFlapse:
There you go

mheguy:

PR #688, opened 2019-01-25
Token table of contents

mheguy:
As usual my documentation of functions I only have a loose grasp of leaves something to be desired: feel free to commit against it or make suggestions and I'll incorporate them.

In addition I would like feedback on ideas for a command, admin-only and not restricted to debug:
`/tokens`
to print the contents table:
```lua
for k, v in pairs(global.tokens.contents) do game.player.print(k .. ' - ' .. v) end
```
and `/tokens <number|string>`
```lua
local index = args.entry
if type(index) == 'string' then
    index = global.tokens.contents[index]
end
log(serpent.block(global.tokens[index]))
```

### DIFF
--- a/features/corpse_util.lua
+++ b/features/corpse_util.lua
@@ -12,7 +12,8 @@ Global.register(
     player_corpses,
     function(tbl)
         player_corpses = tbl
-    end
+    end,
+    'corpse_util'
 )
 
 local function player_died(event)
--- a/features/create_particles.lua
+++ b/features/create_particles.lua
@@ -16,11 +16,16 @@ local settings = {
     max_particles_per_second = 4000,
 }
 
-Global.register({
-    settings = settings,
-}, function (tbl)
-    settings = tbl.settings
-end)
+Global.register(
+    {
+        settings = settings
+    },
+    function(tbl)
+        settings = tbl.settings
+    end,
+    'create_particles'
+)
+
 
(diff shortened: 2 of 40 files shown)


mheguy:
FP to rebase and fix a merge conflict

grilledham:
I'm working on an alternative that will make this PR redundant. 
The changes to blueprint helper to use Global instead of Token could be a separate PR.
I'll add the comments to utils/global and utils/token when my PR is ready.

mheguy:
The alternative you're talking about is debug-only. This is meant for being safe on a live server. I see them as complimentary, not alternatives to one another.

grilledham:
In practice you would never do the debugging on a live public server, you would take the save into single player first and if you are doing that you can just switch _DEBUG to true

mheguy:
Yes... that's definitely how I always do it... yep...

mheguy:
I want this for things like env_dumps and whatnot and I really do see them as serving 2 distinct purposes. A GUI is only practical if you're in-game and in _DEBUG, it won't help with finding the source of, say, a desync attached to a table entry set to delete itself on a timer.

grilledham:
Actually I think _DEBUG will help with the situation you described, if you have to open the save to change _DUMP_ENV changing _DEBUG isn't an extra burden.

PR #689, opened 2019-01-25
Added a toast feature

linaori:
Will add toast messages to a given player.
image

Known limitations:
 - Does not support adding clickable elements (can be done with some fiddling though)
 - It's rendered in sequence of opened GUI elements, meaning it can be opened between elements
 - One color for the bar, though it's easy to add this feature if needed

Features:
 - customizable up-time
 - whole toast is clickable and will close if that happens
 - support for default setup with message
 - calling the function will return a container you can safely add elements to
 - closes when the timeout is reached
 - closes when the player leaves the game

```lua
local Toast = require 'features.gui.toast' 
local toasty = Toast.toast_player(player, duration_in_seconds) -- empty toast
local toasty = Toast.toast_player(player, duration_in_seconds, message) -- adds a label to the toast
```

In the first example it will create an empty toast, but you can add gui elements to the toast. In the second example you can also add elements but they will be placed underneath a message. The first example is useful if the default message isn't enough.

### DIFF
--- a/features/gui/toast.lua
+++ b/features/gui/toast.lua
@@ -1,0 +1,160 @@
+local Event = require 'utils.event'
+local Game = require 'utils.game'
+local Global = require 'utils.global'
+local Gui = require 'utils.gui'
+local Color = require 'resources.color_presets'
+
+local type = type
+local tonumber = tonumber
+local pairs = pairs
+local size = table.size
+
+local Public = {}
+
+local memory = {
+    id = 0,
+    active_toasts = {},
+}
+
+Global.register(memory, function (tbl) memory = tbl end)
+
+---Creates a unique ID for a toast message
+local function autoincrement()
+    local id = memory.id + 1
+    memory.id = id
+    return id
+end
+
+---Toast to a specific player
+---@param p number|string|LuaPlayer player index or object
+---@param duration number in seconds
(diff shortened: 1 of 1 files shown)


features/gui/toast.lua
```diff
@@ -0,0 +1,160 @@
...
+local size = table.size
+
+local Public = {}
+
+local memory = {
+    id = 0,
+    active_toasts = {},
+}
+
+Global.register(memory, function (tbl) memory = tbl end)
```
mheguy: `Global.register(memory, function (tbl) memory = tbl end, 'toast')`
To take advantage of #688 while not causing any problems with the current state of things 😄

linaori: It's a named index for lookups?

linaori: @plague006 question, can you use the stacktrace to automatically create this key? That would make things a lot easier tbh

mheguy: As `traceback` and `get_info` are in lua's debug library, I wouldn't feel comfortable using them for a production feature.

mheguy: So yes, you give a string and it creates a named index in global.tokens.contents and the value is the numeric index of that module inside the global.tokens array

mheguy (approved):
Testing went well in MP. No remaining bugs as far as we could find.

PR #691, opened 2019-01-26
Optimize island_resort: localize functions, get rid of inserts

mheguy:


linaori (approved):

mheguy:
FP to resolve conflict

PR #693, opened 2019-01-27
Have travis issue nightly releases

mheguy:
Travis will push nightly releases. Each release gets its own git tag, but the tags each start with `nightly/` so it shouldn't cause any real clutter. In the future I'll see what I can do about having travis clean up old tags.

When Travis' cron job is triggered, it will not check for linting errors for now, since we haven't cleaned up yet. In the future I would like to lint before pushing builds.
Currently Travis can write the `version` and `map_selection` files for maps/scenarios, and can assemble zip files for each scenario so that they're ready to play. When the deflater is able to handle conditional requires, I should be able to integrate it into the deploy process as well.

As part of a trial period, Travis will not publish actual releases yet but will instead create drafts of releases.

This is what releases will look like when we get the deflator+map loading where we want them:
https://github.com/plague006/RedMew/releases/tag/nightly%2F2019-01-27-9f05e06

And a build log:
https://travis-ci.org/plague006/RedMew/builds/484944639

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,16 +1,46 @@
 language: python
 sudo: false
-
 env:
-  - LUA="lua=5.2"
-
+  matrix:
+    - LUA="lua=5.2"
+branches:
+  only:
+    - master
+    - develop
 before_install:
   - pip install hererocks
-  - hererocks lua_install -r^ --$LUA
-  - export PATH=$PATH:$PWD/lua_install/bin
-
+  - hererocks ~/build/lua_install -r^ --$LUA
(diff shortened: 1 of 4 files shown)


grilledham:
I assume you meant dynamic requires rather than condition.

But why do we need that, where in the code base do we use dynamic requires?

mheguy:
#640

mheguy:
And yes, I don't even want to re-read my PR comment because I wrote it at 2am after fighting 17 hours to get it working.

grilledham:
I'm pretty sure #640 could be refactored to not need dynamic requires

PR #694, opened 2019-01-27
Alternative toast implmentation

grilledham:
I didn't like the `on_gui_click` event implementation as it meant every click on any scenario gui would require a check to see if the element or any of it's parents are a toast. So I had a go at changing that. 

This does mean a slight change to the public api, you now have to opt in to template elements closing the toast.

Here is an example of a custom toast.
```lua
local Toast = require('features.gui.toast')
local Gui = require 'utils.gui'

local button_name = Gui.uid_name()
local close_button_name = Gui.uid_name()

Gui.on_click(
    button_name,
    function(event)
        event.player.print("don't close clicked")
    end
)

Gui.on_click(
    close_button_name,
    function(event)
        Toast.close_toast(event.element)
    end
)

function test()
    Toast.toast_all_players_template(
        15,
        function(container, player)
            container.add {type = 'button', name = button_name, caption = "dont' close"}
            container.add {type = 'button', name = close_button_name, caption = 'close'}
        end
    )
end
```

I also changed the `on_tick` event to be removed when there are no active toasts.

I also changed the public functions that took a player|number to just be player.

### DIFF
--- a/features/gui/toast.lua
+++ b/features/gui/toast.lua
@@ -75,11 +75,6 @@ end
 ---@param duration number in seconds
 local function toast_to(player, duration)
     local frame_holder = player.gui.left.add({type = 'flow'})
-    local frame_holder_style = frame_holder.style
-    frame_holder_style.left_padding = 0
-    frame_holder_style.top_padding = 0
-    frame_holder_style.right_padding = 0
-    frame_holder_style.bottom_padding = 0
 
     local frame =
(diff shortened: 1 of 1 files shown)


grilledham:
Test save on s9 'toast_test.zip'
/c test()
/c local t = require 'features.gui.toast' t.toast_player(game.player, 10, 'hello\ngoodbye')

linaori (approved):

PR #696, opened 2019-01-27
Improve versioning

mheguy:
Have server record approx map start time as the version (if version is empty)
Have the base version.lua be empty and instead rely on the command to provide info about nil version.
Rename command to redmew-color since hitting tab after `color` doesn't autocomplete or present `color-redmew` as an option
Make build script easier to read by using double quotes properly.


grilledham (approved):
I can't review `.travis/release_builder.sh` because I'm outside my comfort zone, but the rest looks good 👍

PR #700, opened 2019-01-28
Add _LIFECYCLE as an indicator of where we are in the data lifecycle

mheguy:
See: https://github.com/Refactorio/RedMew/wiki/The-data-lifecycle

Mark the start of the control stage at the beginning of control.lua
On_load and On_init inside event_core's hook on `script`.
Similarly, mark the beginning of runtime at the end of event_core's `script` hooks on on_init and on_load.

### DIFF
--- a/control.lua
+++ b/control.lua
@@ -1,3 +1,8 @@
+-- If you're looking to configure anything, you want config.lua. Nearly everything in this file is dictated by the config.
+
+-- Info on the data lifecycle and how we use it: https://github.com/Refactorio/RedMew/wiki/The-data-lifecycle
+_LIFECYCLE = 4 -- Control stage
+
 -- Omitting the math library is a very bad idea
 require 'utils.math'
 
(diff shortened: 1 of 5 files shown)


control.lua
```diff
@@ -1,3 +1,8 @@
+-- If you're looking to configure anything, you want config.lua. Nearly everything in this file is dictated by the config.
+
+-- Info on the data lifecycle and how we use it: https://github.com/Refactorio/RedMew/wiki/The-data-lifecycle
+_LIFECYCLE = 4 -- Control stage
```
linaori, line 4: What about something like this?
```lua
function Public.set_stage(stage) end
function Public.in_settings() end
function Public.in_control() end
function Public.passed_stage(stage) end
function Public.before_stage(stage) end
-- etc
```

mheguy: You mean creating a module around tracking the lifecycle?

linaori: I suppose I am :thinking:

mheguy: I'm open to the idea in principle, but before I write another unused module, can you think of applications outside of very niche cases? In general if a function is called outside of the time it's meant to be called, it just fails gracefully. It's important for token, event and event_core to have that information to prevent desyncs.

linaori (approved):

grilledham (approved):

PR #702, opened 2019-01-28
Added anti grief for trains

### LINKED ISSUE #701: [Anti grief] Register train destruction
We recently had a player flip and go run around a manual driven train, destroying multiple trains.

I would propose adding destruction of trains and wagons to the anti grief module.

It should not be triggered on damage, since it's common for people to hit, but not destroy other trains, unintentionally.

It's practical impossible to blame someone for destroying other trains without catching it in progress.

SimonFlapse:
Features include:
* Popups for players when they destroy part of a train
* Kicking players when they have destroyed more than 2 train parts
* Banning guest when they have destroyed more than 3 train parts
* Jailing regulars when they have destroyed more than 3 train parts
* Sets train to automatic mode when a train part has been destroyed
* Does not trigger if the train is in automatic mode
* Toggle able just like on_capsule_used
* Works with on_capsule_used, triggering the ban for guests if they already have been kicked once, regardless of offence.

Tested with heavy mode and on the test server.

closes #701

### DIFF
--- a/features/nuke_control.lua
+++ b/features/nuke_control.lua
@@ -241,7 +241,7 @@ local function on_entity_died(event)
 
     -- If there was a passenger who was unpunished along with a punished passenger, let the admins know
     if player_punished and player_unpunished then
-        local name_string = table.concat(name_list, ' - ')
+        local name_string = table.concat(name_list, ', ')
         Utils.print_admins({'nuke_control.multiple_passengers', num_passengers, name_string})
     end
 end
--- a/features/nuke_control.lua
+++ b/features/nuke_control.lua
@@ -12,13 +12,13 @@ local Ranks = require 'resources.ranks'
 local format = string.format
 local match = string.match
 
-local function allowed_to_nuke(player)
+local function is_trusted(player)
     return Rank.equal_or_greater_than(player.name, Ranks.auto_trusted)
 end
 
 local function ammo_changed(event)
     local player = Game.get_player_by_index(event.player_index)
-    if allowed_to_nuke(player) then
+    if is_trusted(player) then
         return
     end
     local nukes = player.remove_item({name = 'atomic-bomb', count = 1000})
@@ -39,7 +39,7 @@ end
 
 local function on_player_deconstructed_area(event)
     local player = Game.get_player_by_index(event.player_index)
-    if allowed_to_nuke(player) then
+    if is_trusted(player) then
         return
     end
     player.remove_item({name = 'deconstruction-planner', count = 1000})
@@ -141,7 +141,7 @@ local function on_capsule_used(event)
         return
     end
 
-    if not allowed_to_nuke(player) then
+    if not is_trusted(player) then
         local area = {{event.position.x - 5, event.position.y - 5}, {event.position.x + 5, event.position.y + 5}}
         local count = 0
         local entities = player.surface.find_entities_filtered {force = player.force, area = area}
@@ -213,7 +213,7 @@ local function on_entity_died(event)
     for i = 1, #passengers do
         local player = passengers[i]
         if player.valid then
-            if allowed_to_nuke(player) then
+            if is_trusted(player) then
                 player_unpunished = true
                 name_list[#name_list + 1] = player.name
             else
--- a/features/nuke_control.lua
+++ b/features/nuke_control.lua
@@ -4,6 +4,7 @@ local Utils = require 'utils.core'
 local Game = require 'utils.game'
 local Task = require 'utils.task'
 local Token = require 'utils.token'
+local Global = require 'utils.global'
 local Server = require 'features.server'
 local Report = require 'features.report'
 local Popup = require 'features.gui.popup'
@@ -11,6 +12,44 @@ local Ranks = require 'resources.ranks'
 
 local format = string.format
 local match = string.match
+
+local players_warned = {}
+local entities_allowed_to_bomb = {
+    ['stone-wall'] = true,
+    ['transport-belt'] = true,
+    ['fast-transport-belt'] = true,
+    ['express-transport-belt'] = true,
+    ['construction-robot'] = true,
+    ['player'] = true,
+    ['gun-turret'] = true,
+    ['laser-turret'] = true,
+    ['flamethrower-turret'] = true,
+    ['rail'] = true,
+    ['rail-chain-signal'] = true,
+    ['rail-signal'] = true,
+    ['tile-ghost'] = true,
+    ['entity-ghost'] = true,
+    ['gate'] = true,
+    ['electric-pole'] = true,
+    ['small-electric-pole'] = true,
+    ['medium-electric-pole'] = true,
+    ['big-electric-pole'] = true,
+    ['logistic-robot'] = true,
+    ['defender'] = true,
+    ['destroyer'] = true,
+    ['distractor'] = true
+}
+
+Global.register(
+    {
+        players_warned = players_warned,
+        entities_allowed_to_bomb = entities_allowed_to_bomb
+    },
+    function(tbl)
+        players_warned = tbl.players_warned
+        entities_allowed_to_bomb = tbl.entities_allowed_to_bomb
+    end
+)
 
 local function is_trusted(player)
     return Rank.equal_or_greater_than(player.name, Ranks.auto_trusted)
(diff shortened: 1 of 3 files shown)


mheguy:
Since you're working on nuke_control, want to push items from `global` into `Global`?

SimonFlapse:
> Since you're working on nuke_control, want to push items from `global` into `Global`?

You'd like me to change it to use utils.global?

mheguy:
Yes please

mheguy:
(And using `config` for the config options instead of `global`)

features/nuke_control.lua
```diff
@@ -157,6 +171,47 @@ local function on_capsule_used(event)
...
+    local force = event.force
+    if force ~= nil and force == game.forces.player then
+        local train = event.cause.train
+        if train ~= nil and train.manual_mode then
+            train.manual_mode = false
+            for _, player in pairs(train.passengers) do
+                if not player.admin then
+                    local count = players_warned[player.index]
+                    if not count or count < 2 then
+                        Utils.print_admins(""..player.name.." used a train to destroy an entity")
```
mheguy: Not sure why you have an empty string included?

features/nuke_control.lua
```diff
@@ -157,6 +171,47 @@ local function on_capsule_used(event)
...
+                        Popup.player(player, 'You have destroyed part of a train!\nThis is a formal warning and it has been logged\n\nContinual will lead to sanctions')
+                        players_warned[player.index] = (count ~= nil) and count + 1  or 1
+                    else
+                        if allowed_to_nuke(player) and count > 2 then
+                            Report.jail(player)
+                            Popup.player(player, 'You have destroyed part of a train!\nYou have now been placed in jail\n\nContact moderators for assistance\nGo to redmew.com/discord for further enquiries')
+                        elseif count == 2 then
```
mheguy: Kicking someone before you jail them seems antithetical to escalated punishments.

features/nuke_control.lua
```diff
@@ -157,6 +171,47 @@ local function on_capsule_used(event)
...
+                            players_warned[player.index] = count + 1
+                            if nuke_control.enable_autokick then
+                                Popup.player(player, 'You have destroyed part of a train!\nYou have been kicked for this offence\n\nContinual will lead to further sanctions')
```
mheguy: A popup informing someone that they have been kicked is redundant since getting kicked gives you a message.

features/nuke_control.lua
```diff
@@ -157,6 +171,47 @@ local function on_capsule_used(event)
...
+                if not player.admin then
+                    local count = players_warned[player.index]
+                    if not count or count < 2 then
+                        Utils.print_admins(""..player.name.." used a train to destroy an entity")
+                        Popup.player(player, 'You have destroyed part of a train!\nThis is a formal warning and it has been logged\n\nContinual will lead to sanctions')
+                        players_warned[player.index] = (count ~= nil) and count + 1  or 1
+                    else
+                        if allowed_to_nuke(player) and count > 2 then
+                            Report.jail(player)
+                            Popup.player(player, 'You have destroyed part of a train!\nYou have now been placed in jail\n\nContact moderators for assistance\nGo to redmew.com/discord for further enquiries')
```
mheguy: I really like the popup informing someone they're jailed since they can't close it. It would be nice to have it built into jailing.
The message says you're destroying part of a train but that's not necessarily true: cars, tanks, characters, etc. can all be on the tracks.

mheguy (dismissed):

features/nuke_control.lua
SimonFlapse: Newcott and I tested it and it's not obvious why you got kicked

features/nuke_control.lua
SimonFlapse: The thought here was to kick regulars and donators as a warning, that requires no human action, while the jail is instead of banning a regular or donator. Jailing and banning both requires human interaction to revert. Kicking does not. The player gets two popup warnings before they get kicked.

features/nuke_control.lua
SimonFlapse: Simply a brain fart, I think I intended for it to have parts of a string leading up to the player name.

features/nuke_control.lua
SimonFlapse: I'm a bit torn with this, because the popup might limit the visibility of the players chat window.

features/nuke_control.lua
SimonFlapse: Example: 
image

SimonFlapse: The popup stays active when the user logs in again

features/nuke_control.lua
mheguy: You're letting guests kill more trains than regulars before you actually stop them. Doesn't that seem weird to you?

features/nuke_control.lua
mheguy: It does restrict the view of chat slightly, but if you're less verbose, less of the chat window is restricted.
image

features/nuke_control.lua
SimonFlapse: They should have the same number. At the third they get kicked and fourth is ban or jail. But I'll make it 2 and 3 for guests since I made the train stop when it destroys something.

features/nuke_control.lua
SimonFlapse: With #687 we could move the "You have been jailed" to the title and place "Read chat and respond to enquries from staff" I'm the message part. This would make it as small as possible.

grilledham:
I'm against automatic kicking, banning or jailing of regulars. I'm ok with printing messages to game chat or admin chat or whatever when regulars / admins crash trains though.

features/nuke_control.lua
mheguy: > They should have the same number. 

I don't know about the current system, but I'm telling you that during tests I got 3 kills as a guest before being jailed and 2 kills as a regular before being jailed. So in total: more kills as a guest.

mheguy:
I think at this point I'm against automatic kicking/banning of anyone. We can auto-jail them if they're griefing. The only option they have at that point is to spam chat, so I'll write a patch for that case today.

SimonFlapse:
Feel free to commit to this pr, if needed be.

mheguy:
With #715, jailed players can only spam if they realize they're not in the player force and spam with `/s`. If that ever actually becomes a problem, I'll patch that too.

mheguy:
Rebasing + force pushing to get up to date with the lint-free dev head.

mheguy:
Tested well in SP, but cases with multiple passengers are untested.

SimonFlapse:
When I tested with Newcott, multiple passengers where handled individual and everyone got a "strike"

features/nuke_control.lua
```diff
@@ -168,8 +172,82 @@ local function on_player_joined(event)
...
+    -- (we don't care about anything else that happened to be on the tracks)
+    local entity = event.entity
+    if not entity or not entity.valid or not entity.train or entity.force ~= game.forces.player then
+        return
+    end
+    -- Check that the player force did the killing
+    local force = event.force
+    if not force or not force.valid or force ~= game.forces.player then
+        return
+    end
```
grilledham, line 197: The `entity.force ~= game.forces.player` check above makes these checks redundant.

features/nuke_control.lua
```diff
@@ -168,8 +172,82 @@ local function on_player_joined(event)
     end
 end
 
+local train_to_manual =
+    Token.register(
+    function(train)
+        if train and train.valid then
+            train.manual_mode = true
+        end
```
grilledham, line 188: We know `train` wont be `nil` at this point, so only `train.valid` needs checking.

features/nuke_control.lua
```diff
@@ -168,8 +172,82 @@ local function on_player_joined(event)
...
+
+    -- Go through the passengers and punish any guests involved
+    local player_punished
+    local player_unpunished
+    local name_list = {}
+    for i = 1, #passengers do
+        local player = passengers[i]
+        if not player.valid or allowed_to_nuke(player) then
+            player_unpunished = true
+            name_list[#name_list + 1] = player.name
```
grilledham: Can't access `player.name` if `player.valid` is `false`

grilledham (commented):
Now might be the time to change the `allowed_to_nuke` function name to something that better suits its current use.

mheguy:
RB for conflict

grilledham (approved):

PR #703, opened 2019-01-28
Added chat message of deducted experience on death

SimonFlapse:


linaori (approved):

mheguy:
Does this change incentivize shame and/or shaming the player that died for costing the team xp?

linaori:
yes


not more than before though

PR #704, opened 2019-01-28
Added a GUI element + toasts to notify of biter evolution

linaori:
Adds:
 - A button to indicate the current evolution with the latest biter to spawn
 - A toast when a new type of biter spawns

image

image
image
image
image
image
image


grilledham (approved):
LGTM 👍 
But I wonder if we really need to check evolution every 3.5 seconds.

linaori:
I'm not too sure either, it can be tweaked if needed

PR #707, opened 2019-01-29
Add /whois command

mheguy:
Rename stat get functions to match others
`/whois <player>` -- and if you explicitly don't want to see their inventory you would use `/whois <player> no`. The specific use case I had in mind was when you want to get some info from the server console about a player.

image

### DIFF
--- a/features/player_create.lua
+++ b/features/player_create.lua
@@ -62,6 +62,7 @@ local function player_created(event)
 
     if _DEBUG and player.admin then
         UserGroups.add_regular(player.name)
+        game.print("HEY GRILLED DON'T LAUNCH THIS MAP! DEBUG MODE IS ENABLED!!!")
     end
 
     if _CHEATS then
--- a/features/redmew_commands.lua
+++ b/features/redmew_commands.lua
@@ -239,7 +239,7 @@ local function print_player_info(args, player)
         'Rank: ' .. UserGroups.get_rank(target),
         UserGroups.is_donator(target.name) and 'Donator: yes' or 'Donator: no',
         'Time played: ' .. Utils.format_time(target.online_time),
-        'AFK time: ' .. target.afk_time or 0,
+        'AFK time: ' .. Utils.format_time(target.afk_time or 0),
         'Force: ' .. target.force.name,
         'Surface: ' .. target.surface.name,
         'Tag: ' .. target.tag,
--- a/features/redmew_commands.lua
+++ b/features/redmew_commands.lua
@@ -223,22 +223,6 @@ local function print_version()
     Game.player_print(version_str)
 end
 
---- Returns a string indicating the player's rank
-local function get_rank(player)
-    if player.admin then
-        return 'Admin'
-    end
-
-    local name = player.name
-    if UserGroups.is_donator(name) then
-        return 'Donator'
-    elseif UserGroups.is_regular(name) then
-        return 'Regular'
-    end
-
-    return 'Guest'
-end
-
 --- Prints information about the target player
 local function print_player_info(args, player)
     local name = args.player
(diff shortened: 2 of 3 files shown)


features/redmew_commands.lua
```diff
@@ -213,6 +216,45 @@ local function print_version()
...
 
+--- Prints information about the target player
+local function print_player_info(args, player)
+    local name = args.player
+    local target = game.players[name]
+    if not target then
+        Game.player_print('Target not found')
+    end
+    local index = player.index
+    local m_inventory = player.get_inventory(defines.inventory.player_main)
```
grilledham: You seem to be getting the inventory of the player who ran the command not the target?

features/redmew_commands.lua
```diff
@@ -213,6 +216,45 @@ local function print_version()
...
+    m_inventory = m_inventory.get_contents()
+    local q_inventory = player.get_inventory(defines.inventory.player_quickbar)
+    q_inventory = q_inventory.get_contents()
```
grilledham: `m_inventory ` and `q_inventory` only seem to be used below `if player and player.admin and not args.inventory` so why not move them there?

features/redmew_commands.lua
```diff
@@ -213,6 +216,45 @@ local function print_version()
...
+        'Distance walked: ' .. PlayerStats.get_walk_distance(index),
+        'Coin earned: ' .. PlayerStats.get_coin_earned(index),
+        'Coin spent: ' .. PlayerStats.get_coin_spent(index),
+        'Deaths: ' .. PlayerStats.get_death_count(index),
+        'Crafted items: ' .. PlayerStats.get_crafted_item(index),
+        'Chat messages: ' .. PlayerStats.get_console_chat(index)
+    }
+    Game.player_print(concat(info_t, ' - '))
+
+    if player and player.admin and not args.inventory then
```
grilledham: If the server runs this command wont `player` be nil?

features/redmew_commands.lua
```diff
@@ -294,19 +336,30 @@ Command.add(
...
-        allowed_by_server = true,
+        allowed_by_server = true
     },
     print_version
 )
 
+Command.add(
+    'whois',
+    {
+        description = 'provides information about a given player, admins get additional information',
```
grilledham: Adding that you can hide showing the inventory if you add a second argument might be nice.

features/redmew_commands.lua
```diff
@@ -213,6 +216,45 @@ local function print_version()
...
+        'Surface: ' .. player.surface.name,
+        'Tag: ' .. player.tag,
+        'Distance walked: ' .. PlayerStats.get_walk_distance(index),
+        'Coin earned: ' .. PlayerStats.get_coin_earned(index),
+        'Coin spent: ' .. PlayerStats.get_coin_spent(index),
+        'Deaths: ' .. PlayerStats.get_death_count(index),
+        'Crafted items: ' .. PlayerStats.get_crafted_item(index),
+        'Chat messages: ' .. PlayerStats.get_console_chat(index)
+    }
+    Game.player_print(concat(info_t, ' - '))
```
grilledham: Just a thought but adding `\n` to the separator might make this easier to read from the server console, and will be treated as `' '` while in game

features/redmew_commands.lua
mheguy: Not just the inventory... *facepalm*

linaori (approved):

features/player_create.lua
```diff
@@ -62,6 +62,7 @@ local function player_created(event)
 
     if _DEBUG and player.admin then
         UserGroups.add_regular(player.name)
+        game.print("HEY GRILLED DON'T LAUNCH THIS MAP! DEBUG MODE IS ENABLED!!!")
```
grilledham, line 65: Because I'm so paranoid about this, I've never actually done it.

grilledham (approved):

features/player_create.lua
mheguy: Now that the warning is in place, you're sure to do it. :D

PR #708, opened 2019-01-29
Add get_player_from_any

mheguy:
There are a number of places in the repo that want to take player as multiple types. This should help consolidate some of that code.

### DIFF
--- a/utils/game.lua
+++ b/utils/game.lua
@@ -45,6 +45,25 @@ function Game.get_player_by_index(index)
     end
 end
 
+--- Returns a valid LuaPlayer if given a number, string, or LuaPlayer. Returns nil otherwise.
+-- obj <number|string|LuaPlayer>
+function Game.get_player_from_any(obj)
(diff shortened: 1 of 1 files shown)


utils/game.lua
```diff
@@ -45,6 +45,25 @@ function Game.get_player_by_index(index)
...
+--- Returns a valid LuaPlayer if given a number, string, or LuaPlayer. Returns nil otherwise.
+-- obj <number|string|LuaPlayer>
+function Game.get_player_from_any(obj)
+    local o_type = type(obj)
+    local p
+    if type == 'number' then
+        p = Game.get_player_by_index(obj)
+    elseif o_type == 'string' then
+        p = game.players[obj]
+    elseif o_type == 'table' and obj.is_player() then
```
grilledham: `elseif o_type == 'table' and obj.valid and obj.is_player() then`
else this line could throw an error.

grilledham (approved):

PR #712, opened 2019-01-29
Refine/add functions in Table

mheguy:
`index_of` was improperly documented by me, it did not return the index of a dictionary table entry, only if it were an array.

`contains` - functionality unchanged
`array_contains` - a faster version of `contains` meant for arrays only

`index_of` - now functions the way described in the documentation
`index_of_in_array` - a faster version of `index_of` meant for arrays only


linaori (approved):

mheguy:
Rebase + FP

PR #713, opened 2019-01-30
Overhaul rank system, command, donator, add probation rank

### LINKED ISSUE #610: Rewrite Command to factor in regulars and donators properly
https://github.com/Refactorio/RedMew/pull/566 adds regulars but it's kinda clunky/tacked on.

Once #576 and #566 are merged it would be nice to redo that system and also accommodate any potential donator-only features (ex. allowing them to change their sign-in message).

mheguy:
Overhauls the rank system to allow for expansion/changing quickly and easily.

This enforces a hierarchy of ranks that exists in resources.ranks and everything simply references that table.
features.rank_system hosts the operators to evaluate a player's rank, allows the setting and resetting of ranks, handles the server side of things, and auto-promotes to a new (name-pending) rank between `guest` and `regular`.

In commands, changes the `rank_only` paradigm to instead take the required rank and treats donator as its own distinct flag outside of the hierarchy.

The donator feature is split off from the rankings, and now includes donator_message inside it.

Only mild testing was done in SP, and nothing in MP.

Closes #610
Closes #747


map_gen/presets/rotten_apples.lua
```diff
@@ -194,7 +195,7 @@ local function research_finished(event)
     local research = event.research
     local force = research.force
 
-    if table.contains(military_techs, research.name) then
+    if array_contains(military_techs, research.name) then
```
grilledham: Beyond the scope for this PR, but using an array rather than a dictionary for military_techs makes me shudder.

utils/table.lua
```diff
@@ -48,28 +48,47 @@ function table.add_all(t1, t2)
 end
 
 --- Checks if a table contains an element
--- @param t <table> to search
--- @param e <any> table element to search for
--- @returns <number|string> the index of the element or -1 for failure
+-- @param t <table>
+-- @param e <any> table element
+-- @returns <number|string|nil> the index of the element or nil
 function table.index_of(t, e)
```
grilledham: Doesn't this return \<any\>? As you can have any type for the key in a table.

features/rank_system.lua
```diff
@@ -0,0 +1,239 @@
...
+        player_ranks = tbl.player_ranks
+    end
+)
+
+-- Local functions
+
+--- Check each online player and if their playtime
+local function check_playtime()
+    local auto_trusted = Ranks.auto_trusted
+    local time_for_trust = Config.time_for_trust
```
grilledham, line 81: I don't see this value in the config file.

features/rank_system.lua
```diff
@@ -0,0 +1,239 @@
...
+end
+
+-- Exposed functions
+
+--- Returns the player's rank as a name.
+-- @param player_name <string>
+-- @return <string>
+function Public.get_player_rank_name(player_name)
+    return index_of(Ranks, (player_ranks[player_name] or 0))
+end
```
grilledham, line 172: Given how often this method could be called, I'd be tempted to make a dictionary that maps ranks_number -> rank_name.

features/rank_system.lua
```diff
@@ -0,0 +1,239 @@
...
+    local actor = Utils.get_actor()
+
+    if Public.equal(player_name, rank) then
+        Game.player_print(format('%s is %s rank already.', player_name, Public.get_rank_name(rank)))
+    else
+        player_ranks[player_name] = rank
+        Server.set_data(ranking_data_set, player_name, rank)
+        game.print(format("%s set %s's rank to %s.", actor, player_name, Public.get_rank_name(rank)))
+    end
+end
```
grilledham, line 339: Might be worth mentions in the comments to use `reset_rank` for setting to guest

features/rank_system.lua
```diff
@@ -0,0 +1,239 @@
...
+
+--- Resets a player's rank
+-- @param player_name <string>
+function Public.reset_rank(player_name)
```
grilledham: I'd add to the comments that this resets to guest rank and respects auto_trusted time, or words to that effect.

features/rank_system.lua
```diff
@@ -0,0 +1,239 @@
...
+--- Check each online player and if their playtime
+local function check_playtime()
+    local auto_trusted = Ranks.auto_trusted
+    local time_for_trust = Config.time_for_trust
+    local less_than = Public.less_than
+    local set_data = Server.set_data
+
+    for _, p in pairs(game.connected_players) do
+        local player_name = p.name
+        if (p.online_time > time_for_trust) and less_than(player_name, auto_trusted) then
```
grilledham: Shouldn't this not promote players with the probation rank?

features/rank_system.lua
```diff
@@ -0,0 +1,239 @@
...
+)
+
+Event.add(
+    Server.events.on_server_started,
+    function()
+        Public.sync_rankings()
+    end
+)
+
+Event.on_nth_tick(nth_tick, check_playtime)
```
grilledham: With such a long timeout it would be possible for a player to have:
time_for_trust < player.time < time_for_trust + nth_tick
and not be promoted. For small nth_tick this would be fine, but probably not if it's an hour. 

Another strategy would be to keep track of online guests with player_join and player_leave and only iterate through those players on nth_tick.
Bonus points if you keep it as an ordered list, then you only have to check players until the first fail.

resources/ranks.lua
```diff
@@ -0,0 +1,7 @@
+return {
+    probation = -1,
+    guest = 0,
+    auto_trusted = 1,
+    regular = 2,
+    admin = 3,
+}
```
grilledham, line 8: So these are going to become the values that are stored in Scenario Data?
As changing those values would be annoying, maybe we should use -10, 0, 10, 20, 30 so we have some wiggle to add ranks.

grilledham:
I like this. We will need to test it in MP before I can approve it though.

grilledham:
I don't think `set_rank` and `reset_rank` should print messages, that's up to the calling code to decide.
Those messages make sense if the functions are called from a command. But maybe for example the antigreif module wants to call those function, then it would want to suppress the messages, or print different ones.

I suggest the functions return a Boolean instead.

grilledham:
Currently we can set and reset the ranks of players who have never been in the game, I'm ok with that (I'm also ok with not allowing it), I just want to know if that was intentional

map_gen/presets/rotten_apples.lua
mheguy: Yeah, when I looked at it while making this change I noticed it as well. Time and experience teaches. :D

features/rank_system.lua
mheguy: I know I wrote it, so I must have discarded the changes when I discarded debug and cheat mode at some point. :D

mheguy:
> Currently we can set and reset the ranks of players who have never been in the game, I'm ok with that (I'm also ok with not allowing it), I just want to know if that was intentional

Intentional as there have been multiple times I wanted to regular or un-regular players who hadn't been on a particular map. But on reflection I should include a warning in those cases since typos are a thing.

features/rank_system.lua
mheguy: Nice catch, thank you.

resources/ranks.lua
mheguy: Right, to not have to migrate the data as often: that makes total sense.

mheguy: I forgot a couple functions that are important/useful: `increase_player_rank` and decrease. Without being able to add or subtract by 1, I'm actually unsure how else to implement them besides having the ranks be +- 1 of each other.

mheguy:
> 
> 
> > Currently we can set and reset the ranks of players who have never been in the game, I'm ok with that (I'm also ok with not allowing it), I just want to know if that was intentional
> 
> Intentional as there have been multiple times I wanted to regular or un-regular players who hadn't been on a particular map. But on reflection I should include a warning in those cases since typos are a thing.

So that was my view on the matter when I was just a moderator.. as a dev looking at my system and trying to account for typos: I want to enforce the existence of a player when running `/commands` but not have the rank_system module itself check for a LuaPlayer because I think there are legitimate reasons to want to change ranks without a player having been on a particular map.

features/rank_system.lua
mheguy: Went with a "nothing fancy" approach.

On join and rank == guest enter the `guests` table.

On nth tick
if auto_trusted or higher rank: off the list
elseif guest rank and enough time: off the list and rank to auto_trust
elseif not online: off the list

So you only stay on the list if you are: lower rank than auto_trust, don't have enough time, are online.

If the implementation works for you, feel free to hit resolve convo.

mheguy: This implementation means I can keep the timer at 1h and still catch anyone who meets the criteria since it doesn't rely on a player being online.

mheguy: Because this specifically will be a pain to test in MP, just wanted to show it working in SP:

image

resources/ranks.lua
grilledham: Something like this
```lua
local ranks = {
    probation = -10,
    guest = 0,
    auto_trusted = 10,
    regular = 20,
    admin = 30
}

local sorted_ranks = {}
local rank_to_index = {}

for k, v in pairs(ranks) do
    sorted_ranks[#sorted_ranks + 1] = v
end

table.sort(sorted_ranks)

for k, v in pairs(sorted_ranks) do
    rank_to_index[v] = k
end

function change_rank(current_rank, change)
    local index = rank_to_index[current_rank]

    local new_index = math.clamp(index + change, 1, #sorted_ranks)

    return sorted_ranks[new_index]
end
```
Example of use
```lua
change_rank(ranks.guest,3) -- returns 30
change_rank(ranks.regular,-2) -- returns 0
```

mheguy: Ty for doing my work for me. <3

mheguy:
redmew_qol.pick_name() will need a fix when `global.regulars` is no longer available
outpost_builder.do_capture_outpost() will need a fix for `global.donators` no longer being available

mheguy:
Fix conflicts -> rebase and FP

mheguy:
#747 Added in

mheguy:
https://github.com/Refactorio/RedMew/pull/713/commits/d21ad8d8b542da0d18eac2c074ec966499d858fe should keep things more consistent so that loaded saves and server scenario data do not conflict.

mheguy:
More changes! (yay~~)

Added `increase_player_rank_to` and `decrease_player_rank_to`. Unlike `set_rank` this ensures that when we mean to increase/decrease a rank, we don't accidentally do the opposite.

Remove all prints from the module and instead return strings that calls can choose to use or ignore.

Have `reset_player_rank` call `set_player_rank` to consolidate where we set server data.

---

Add color to `Game.player_print`

---

Overhaul `regular` and `probation` to take format `/regular <add|remove> <name>`.
As part of the overhaul, have them use `increase/decrease_player_rank_to` to restrain their actions.

---

Limited testing done in SP.

mheguy:
~~`/whois` 1459.479 Script @/factorio/4/temp/currently-playing/utils/command.lua:253: Error while running 'whois' with arguments {inventory = false, player = "plague006"}: /factorio/4/temp/currently-playing/features/rank_system.lua:80: bad argument #2 to '__index' (string expected, got table)~~

mheguy:
RB+FP for merge conflict.
Moved what I could in rank_system into locale, and only what I had to in dependent modules.
~~Listing of `rank` in `/whois` is disabled until I can figure out a sane way to incorporate all of that together.~~

plague006:
Best way to tackle whois might be concatenating everything into print, example:
`game.print({"", {"item-name.iron-plate"}, ": ", 60})`

grilledham:
I updated the player list gui
image
image
image
Not tested in multiplayer

grilledham (approved):
On the basis that this was tested in multiplayer, this is probably safe enough to merge.

PR #715, opened 2019-01-30
Change how we treat players in jail

mheguy:
Move from global to Global

Change jailed players to a jailed force so that they can't spam general chat.

Route jailed players' chat into admin chat, and route an admin speaking in general chat to the jailed force.

Stop any vehicle a griefer was in when they're jailed.

When a player is jailed, clear their chat console so they see only what we want them to see. No distractions from old general chat.

Make players indestructible. No more players killing a griefer when we're trying to collect evidence.

Change their color and chat color to white in order to single them out more and to ensure their chat is legible.

Register/unregister the on_chat hook so that we don't have it listening if we don't need it.

Known shippable:
They can still use `/s` to spam. If anyone ever figures that out then I'll also track their command usage and if they use `/s` X times while jailed, mute them.

### DIFF
--- a/features/report.lua
+++ b/features/report.lua
@@ -2,8 +2,6 @@ local Gui = require 'utils.gui'
 local Utils = require 'utils.core'
 local Game = require 'utils.game'
 local Global = require 'utils.global'
-local Event = require 'utils.event'
-local Token = require 'utils.token'
 local Popup = require 'features.gui.popup'
 local Color = require 'resources.color_presets'
 
@@ -15,7 +13,6 @@ local report_tab_button_name = Gui.uid_name()
 local jail_offender_button_name = Gui.uid_name()
 local report_body_name = Gui.uid_name()
 local jail_name = 'Jail'
-local jail_force = 'jailed'
 local default_group = 'Default'
 local prefix = '------------------NOTICE-------------------'
 local prefix_e = '--------------------------------------------'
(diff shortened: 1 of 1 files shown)


features/report.lua
```diff
@@ -11,16 +15,51 @@ local report_tab_button_name = Gui.uid_name()
...
+    Token.register(
+    function(event)
+        local index = event.player_index
+        local player = Game.get_player_by_index(index)
+        local message
+
+        if jail_data[index] then
+            message = 'From jail: ' .. event.message
+            Utils.print_admins(message, player.name)
+        elseif player.admin then
```
linaori: This can be a bit weird to non-admins to see

linaori (approved):

features/report.lua
mheguy: Non-admins don't see messages from jail. it's to prevent jailbirds from spamming our players.

linaori: For admins to talk back, they have to use the chat. This means that non-admins will see half the discussion, admins talking to jailed players without replies

mheguy: Ah right. I'll have to think on that one.

mheguy: Oh right. My thought was that we actually whisper them.

mheguy: But I've noticed some admins don't whisper and instead use chat.

PR #716, opened 2019-01-30
Debugger

grilledham:
I don't consider this finished, but I'm at the stage where I want to take a step back and think about this some more / get feedback.

As it's not meant for production code, it should be safe to merge and only use for debugging purposes.

## Disclaimer

The debugger should only be used as a tool to inspect data. Because of the way the debugger displays data it may cause desyncs in multiplayer. Differences in data displayed between clients in itself does not mean the data isn't desync safe.

## How to use

set `_DEBUG = true` in config.lua
`/debug` to open

## Global - Everything in Global.register / Global.register_init - content of global.tokens
image

Can also inspect data further by typing in the path. Effectively anything you type in the textfield is treated as if you had run it using /c 

image

## global - Everything else in global

image

## package - Everything in _G.package.loaded - all loaded modules
Excludes the base modules, so only contains the scenario ones.
The idea was to be able to inspect local variables in a file by looking at the upvaules for the Public functions. 
I think it could still use some work, but I'm not sure the direction I want to take it.

image

## _G Everything in _G excluding all the base stuff, so only what the scenario puts in there.

Also defines because it's useful to see the content. Really defines shouldn't be there. 
image

### DIFF
--- a/control.lua
+++ b/control.lua
@@ -106,3 +106,8 @@ end
 if _DUMP_ENV then
     require 'utils.dump_env'
 end
+
+-- Needs to be at bottom so tokens are registered last.
+if _DEBUG then
+    require 'features.gui.debug.command'
+end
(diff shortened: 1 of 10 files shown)


mheguy:
After waiting a week on https://github.com/Refactorio/RedMew/pull/688 I'm merging this just so I have something in the codebase to help me while coding.

PR #718, opened 2019-01-31
Toast for capture / upgrade outposts

grilledham:
Tested by myself in multiplayer

### DIFF
--- a/map_gen/presets/crash_site/outpost_builder.lua
+++ b/map_gen/presets/crash_site/outpost_builder.lua
@@ -8,6 +8,7 @@ local PlayerStats = require 'features.player_stats'
 local RS = require 'map_gen.shared.redmew_surface'
 local Server = require 'features.server'
 local Color = require 'resources.color_presets'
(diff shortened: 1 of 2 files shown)


map_gen/presets/crash_site/outpost_builder.lua
```diff
@@ -1521,7 +1535,7 @@ function Public.do_random_loot(entity, weights, loot)
     end
 
     entity.operable = false
-    entity.destructible = false
+    --entity.destructible = false
```
linaori, line 1538: on purpose?

linaori (approved):
Looks good!

map_gen/presets/crash_site/outpost_builder.lua
grilledham: yeah, really I should make a config file like diggy has, but that's a problem for future me

grilledham:
I'm going to merge this so I can test plague's script to update the scenario on the server

mheguy:
Failing travis' tests is an actual fail now. :p

PR #721, opened 2019-02-01
Add team module

mheguy:
In-file doc should cover most things:

It's a team balancing module. Currently it takes a set number of teams pre-init and can keep teams balanced based on the selected event hooks.


mheguy:
I want to put more work into this. I don't like it.

PR #722, opened 2019-02-01
Adaption of danger_ore to ban entities

SimonFlapse:
Needed for the prehistoric map

Needs #687

### DIFF
--- a/map_gen/shared/banned_entities.lua
+++ b/map_gen/shared/banned_entities.lua
@@ -56,12 +56,6 @@ Event.add(
             return
         end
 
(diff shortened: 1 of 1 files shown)


mheguy:
This module is fine for inclusion in a specific map, but it's not generic enough to be in `shared`.
You have things like a hardcoded list of entities and hardcoded feedback when an entity is destroyed.

These factors indicate this is a bespoke feature of a single map. I'll happily merge this into whatever map you would like, and I'll write a generic feature.

SimonFlapse:
Great, thanks :) Didn't have the time to do anything like it.

PR #723, opened 2019-02-01
Add generic entity restriction module

mheguy:
**Hey!**
A lot of this module got overhauled and fixed up. This comment is completely rewritten as well.

Shameless copy-paste of the in-code doc:

This module restricts the placement of entities and ghosts based on an allowed and banned list,
as well as by the (optionally) provided function.

The table of allowed_entities are *always* allowed to be placed.
The table of banned_entities are *never* allowed to be placed, and are destroyed.

For anything not in either of those lists, you can use the set_keep_alive_callback function to set a keep_alive_callback function.
This means you can use any custom logic you want to determine whether an entity should be destroyed or not.
The callback function is supplied a valid LuaEntity as an argument.
A return of true indicates the entity should be kept alive, while false or nil indicate it should be destroyed.

Refunds for items that were placed can be toggled on or off via the enable and disable_refund functions

Lastly, this module raises 2 events: on_pre_restricted_entity_destroyed and on_restricted_entity_destroyed events.
They are fully defined below.
```
Examples (only the first example will include the require):
-- A map which allows no roboports:
local RestrictEntities = require 'map_gen.shared.entity_placement_restriction'
RestrictEntities.add_banned({'roboport'})

-- A map which allows only belts (for a foot race map, of course)
-- The function provided does nothing but return nil
-- every entity will be destroyed except those on the allowed list
RestrictEntities.add_allowed({'transport-belt'})
RestrictEntities.set_keep_alive_callback(function() end)

-- Danger ores (a lot of important code omitted for the sake of a brief example)
RestrictEntities.add_allowed({belts, power_poles, mining_drills, 'pumpjack'})
RestrictEntities.set_keep_alive_callback(
    function(entity)
        if entity.surface.count_entities_filtered {area = entity.bounding_box, type = 'resource', limit = 1} == 0 then
            return true
        end
    end
)
```

-----

In-module debug is included in commit: https://github.com/Refactorio/RedMew/pull/723/commits/d6e37cdba72a98eefaa37a00cb825982c3da6831

I'm not sure where to put `danger_ore_banned_entities.lua` now.. Do I just whip up a wiki page for this module and drop the contents of banned ores into it? Do we add the code to the modules that traditionally use danger ore?

I can't figure out why nothing actually requires `danger_ore_banned_entities`. -_-

Closes https://github.com/Refactorio/RedMew/pull/722

### DIFF
--- a/map_gen/shared/entity_placement_restriction.lua
+++ b/map_gen/shared/entity_placement_restriction.lua
@@ -121,8 +121,7 @@ local on_built_token =
 
         event.ghost = ghost
         event.player = p
-        raise_event(Public.events.on_pre_restricted_entity_destroyed, deep_copy(event)) -- use deepcopy so that any potential writes to `event` aren't passed backwards
-
+        raise_event(Public.events.on_pre_restricted_entity_destroyed, event)
         -- Need to revalidate the entity since we sent it via the event
         if entity.valid then
             entity.destroy()
--- a/map_gen/shared/entity_placement_restriction.lua
+++ b/map_gen/shared/entity_placement_restriction.lua
@@ -1,11 +1,12 @@
 --[[
     Regulates the placement of entities and their ghosts
     Can be called through public function to provide lists of allowed or banned entities.
-    Can also set a keep_alive_callback function to check what is beneath the entity, for example for checking for ores or certain tiles.
+    Can also set a keep_alive_callback function to process information about the entity about to be destroyed
+    for example, checking for resources underneath the entity or checking the pollution in the entity's chunk.
 
-    Example of keep_alive_callback function to check if there are ores beneath:
-    function keep_alive_callback (surface, area)
-        local count = surface.count_entities_filtered {area = area, type = 'resource', limit = 1}
+    Example of keep_alive_callback function to check if there are resources beneath:
+    function keep_alive_callback (entity)
+        local count = entity.surface.count_entities_filtered {area = area, type = 'resource', limit = 1}
         if count == 0 then
             return true
         end
--- a/map_gen/shared/entity_placement_restriction.lua
+++ b/map_gen/shared/entity_placement_restriction.lua
@@ -1,20 +1,20 @@
 --[[
     Regulates the placement of entities and their ghosts
     Can be called through public function to provide lists of allowed or banned entities.
-    Can also set a logic function to check what is beneath the entity, for example for checking for ores or certain tiles.
+    Can also set a keep_alive_callback function to check what is beneath the entity, for example for checking for ores or certain tiles.
 
-    Example of logic function to check if there are ores beneath:
-    function logic(surface, area)
+    Example of keep_alive_callback function to check if there are ores beneath:
+    function keep_alive_callback (surface, area)
         local count = surface.count_entities_filtered {area = area, type = 'resource', limit = 1}
         if count == 0 then
             return true
         end
     end
 
-    Allowed entities: are exempt from the logic check and are never destroyed.
+    Allowed entities: are exempt from the keep_alive_callback check and are never destroyed.
     Banned entities: are always destroyed.
 
-    Maps can hook the on_restricted_entity_destroyed event to generate whatever feedback they deem appropriate
+    Maps can hook the on_pre_restricted_entity_destroyed and on_restricted_entity_destroyed events.
 ]]
 local Event = require 'utils.event'
 local Game = require 'utils.game'
@@ -24,12 +24,13 @@ local table = require 'utils.table'
 
 -- Localized functions
 local raise_event = script.raise_event
+local deep_copy = table.deep_copy
 
 local Public = {
     events = {
         --[[
-        on_restricted_entity_destroyed
-        Triggered when an entity is destroyed by this script
+        on_pre_restricted_entity_destroyed
+        Called before an entity is destroyed by this script
         Contains
             name :: defines.events: Identifier of the event
             tick :: uint: Tick the event was generated.
@@ -39,30 +40,42 @@ local Public = {
             stack :: LuaItemStack
             ghost :: boolean indicating if the entity was a ghost
         ]]
+        on_pre_restricted_entity_destroyed = script.generate_event_name(),
+        --[[
+        on_restricted_entity_destroyed
+        Called when an entity is destroyed by this script
+        Contains
+            name :: defines.events: Identifier of the event
+            tick :: uint: Tick the event was generated.
+            player_index :: uint
+            player :: LuaPlayer
+            ghost :: boolean indicating if the entity was a ghost
+            item_returned :: boolean indicating if the item was returned by this module
+        ]]
         on_restricted_entity_destroyed = script.generate_event_name()
     }
 }
 
 -- Global-registered locals
 
-local allowed_entites = {}
-local banned_entites = {}
+local allowed_entities = {}
+local banned_entities = {}
 local primitives = {
     event = nil,
     allowed_ents = nil,
     banned_ents = nil,
-    logic = nil
+    keep_alive_callback = nil
 }
 
 Global.register(
     {
-        allowed_entites = allowed_entites,
-        banned_entites = banned_entites,
+        allowed_entities = allowed_entities,
+        banned_entities = banned_entities,
         primitives = primitives
     },
     function(tbl)
-        allowed_entites = tbl.allowed_entites
-        banned_entites = tbl.banned_entites
+        allowed_entities = tbl.allowed_entities
+        banned_entities = tbl.banned_entities
         primitives = tbl.primitives
     end
 )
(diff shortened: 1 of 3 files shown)


plague006:
#722 becomes:
```lua
local RestrictEntities = require 'map_gen.shared.entity_placement_restriction'

local banned_entities = {
    'inserter',
    'long-handed-inserter',
    'fast-inserter',
    'filter-inserter',
    'stack-inserter',
    'stack-filter-inserter',
    'electric-mining-drill'
}

RestrictEntities.add_banned(banned_entities)

Event.add(
    RestrictEntities.events.on_restricted_entity_destroyed,
    function(event)
        if not event.ghost then
            require 'features.gui.popup'.player(p, [[
You don't know how to operate this item!

Advice: Only burner inserters and burner mining drills works in prehistoric land
]], nil, nil, 'map_gen.shared.banned_entities')
        end
    end
)
```

plague006:
`danger_ore_banned_entities.lua`
becomes:
```lua
local RestrictEntities = require 'map_gen.shared.entity_placement_restriction'

local allowed_entities = {
    'transport-belt',
    'fast-transport-belt',
    'express-transport-belt',
    'underground-belt',
    'fast-underground-belt',
    'express-underground-belt',
    'small-electric-pole',
    'medium-electric-pole',
    'big-electric-pole',
    'substation',
    'electric-mining-drill',
    'burner-mining-drill',
    'pumpjack'
}

local function logic(surface, area)
    local count = surface.count_entities_filtered {area = area, type = 'resource', limit = 1}
    if count == 0 then
        return true
    end
end

RestrictEntities.add_allowed(allowed_entities)
RestrictEntities.set_logic(logic)
```

mheguy:
`terraforming.lua`
gets a big chunk of its code ripped out

map_gen/shared/entity_placement_restriction.lua
```diff
@@ -0,0 +1,224 @@
...
+-- Localized functions
+local raise_event = script.raise_event
+
+local Public = {
+    events = {
+        --[[
+        on_restricted_entity_destroyed
+        Triggered when an entity is destroyed by this script
+        Contains
+            name :: defines.events: Identifier of the event
```
grilledham, line 54: Untested, but I have a suspicion that because of the way the event is raised the name will be the old event.
It might be necessary to do `event.name = Public.event.on_restricted_entity_destroyed` down below.

map_gen/shared/entity_placement_restriction.lua
```diff
@@ -0,0 +1,224 @@
...
+            tick :: uint: Tick the event was generated.
+            created_entity :: LuaEntity
```
grilledham: Wouldn't the entity be destroyed by this point, and therefore invalid?

map_gen/shared/entity_placement_restriction.lua
```diff
@@ -0,0 +1,224 @@
...
+}
+
+-- Global-registered locals
+
+local allowed_entites = {}
+local banned_entites = {}
+local primitives = {
+    event = nil,
+    allowed_ents = nil,
+    banned_ents = nil,
```
grilledham: the `allowed_ents` and `banned_ents` variables seem redundant, isn't that information contained in whether or not `allowed_entites` or `banned_entites` are empty.

map_gen/shared/entity_placement_restriction.lua
```diff
@@ -0,0 +1,224 @@
...
+        -- Some entities have a bounding_box area of zero, eg robots.
+        local area = entity.bounding_box
+        local left_top, right_bottom = area.left_top, area.right_bottom
+        if left_top.x == right_bottom.x and left_top.y == right_bottom.y then
+            return
+        end
+
+        -- Takes the logic function (if provided) and runs it with the surface and area as arguments
+        -- this allows the logic function to scan underneath the entity. if true is returned, we exit.
+        -- If false, we destroy the entity and return it
```
grilledham: >>and return it ... to the player?

or return the item to the player.

map_gen/shared/entity_placement_restriction.lua
```diff
@@ -0,0 +1,224 @@
...
+        local logic = primitives.logic
+        if logic and logic(entity.surface, area) then
```
grilledham: The abstraction here feels wrong, I think we should just pass the entity into the logic function.
That way we can remove doing the work to get the surface and area if the logic function doesn't want that.
Also with the above code for checking if the area is empty it wouldn't be possible for the logic function to take action on robots.

map_gen/shared/entity_placement_restriction.lua
```diff
@@ -0,0 +1,224 @@
+--[[
+    Regulates the placement of entities and their ghosts
+    Can be called through public function to provide lists of allowed or banned entities.
+    Can also set a logic function to check what is beneath the entity, for example for checking for ores or certain tiles.
```
grilledham: Maybe `logic` should be renamed to `keep_alive_callback` or something to make it easier to remember if you have to return true or false

map_gen/shared/entity_placement_restriction.lua
```diff
@@ -0,0 +1,224 @@
...
+
+        -- Takes the logic function (if provided) and runs it with the surface and area as arguments
+        -- this allows the logic function to scan underneath the entity. if true is returned, we exit.
+        -- If false, we destroy the entity and return it
+        local logic = primitives.logic
+        if logic and logic(entity.surface, area) then
+            return
+        end
+
+        if primitives.banned_ents and not banned_entites[name] and not primitives.allowed_ents then
```
grilledham: This line confuses me. 
If `banned_entites` are always destroyed as stated on line 15, then why is this check happening after `logic()`.
Also I don't get why it's checking for `allowed_ents`

map_gen/shared/entity_placement_restriction.lua
```diff
@@ -0,0 +1,224 @@
...
+            return
+        end
+
+        entity.destroy()
+        event.item_returned = false
+        if not ghost then
+            p.insert(event.stack)
+            event.item_returned = true
+        end
+        event.ghost = ghost
```
grilledham: I'd probably refactor
```lua
entity.destroy()
event.item_returned = false
if not ghost then
    p.insert(event.stack)
    event.item_returned = true
end
event.ghost = ghost
```
into 
```lua
entity.destroy()

if ghost then
    event.ghost = true
else
    p.insert(event.stack)
    event.item_returned = true
end
```

map_gen/shared/entity_placement_restriction.lua
```diff
@@ -0,0 +1,224 @@
...
+    if not primitives.event then
+        add_event()
+    end
+end
+
+--- Removes from the list of allowed entities
+-- @param ents <table> array of entity strings
+function Public.remove_allowed(ents)
+    for _, v in pairs(ents) do
+        allowed_entites[v] = true
```
grilledham: both `remove_allowed` and `add_allowed` appear to add the entities to `allowed_entites`

map_gen/shared/entity_placement_restriction.lua
```diff
@@ -0,0 +1,224 @@
...
+    if not primitives.event then
+        add_event()
+    end
+end
+
+--- Removes from the list of banned entities
+-- @param ents <table> array of entity strings
+function Public.remove_banned(ents)
+    for _, v in pairs(ents) do
+        banned_entites[v] = true
```
grilledham: both `add_banned` and `remove_banned` appear to add the entities to `banned_entites`

map_gen/shared/entity_placement_restriction.lua
```diff
@@ -0,0 +1,224 @@
...
+            ghost :: boolean indicating if the entity was a ghost
+        ]]
+        on_restricted_entity_destroyed = script.generate_event_name()
+    }
+}
+
+-- Global-registered locals
+
+local allowed_entites = {}
+local banned_entites = {}
```
grilledham: I don't often get to say this but 'entities' is spelt wrong 😛

grilledham (dismissed):

map_gen/shared/entity_placement_restriction.lua
mheguy: It's intended as a speed optimization. Because those tables are dictionaries, so it's `size()` vs a lookup. If I'm wrong about it being faster I would be happy to switch it. I haven't A B tested it.

map_gen/shared/entity_placement_restriction.lua
mheguy: The name is changed correctly. `raise_event()` writes to `event.` `mod_name`, `name`, and `tick` when called.

map_gen/shared/entity_placement_restriction.lua
mheguy: Fixed

map_gen/shared/entity_placement_restriction.lua
mheguy: 20 times, no less. :D

map_gen/shared/entity_placement_restriction.lua
mheguy: Because of a different refactor, this wouldn't work.

map_gen/shared/entity_placement_restriction.lua
mheguy: That was just bad logic on my part. Fixed.

map_gen/shared/entity_placement_restriction.lua
```diff
@@ -0,0 +1,238 @@
+--[[
+    Regulates the placement of entities and their ghosts
+    Can be called through public function to provide lists of allowed or banned entities.
+    Can also set a keep_alive_callback function to check what is beneath the entity, for example for checking for ores or certain tiles.
+
+    Example of keep_alive_callback function to check if there are ores beneath:
+    function keep_alive_callback (surface, area)
```
grilledham: No one updates the docs 😛 
This now takes an entity.

map_gen/shared/entity_placement_restriction.lua
```diff
@@ -0,0 +1,238 @@
...
+        end
+
+        local p = Game.get_player_by_index(event.player_index)
+        if not p or not p.valid then
+            return
+        end
+
+        event.ghost = ghost
+        event.player = p
+        raise_event(Public.events.on_pre_restricted_entity_destroyed, deep_copy(event)) -- use deepcopy so that any potential writes to `event` aren't passed backwards
```
grilledham: >> -- use deepcopy so that any potential writes to `event` aren't passed backwards

By that logic shouldn't this function deep copy the table before it writes to it and use the copied table
As this function does the very thing it's protecting itself from.

map_gen/shared/entity_placement_restriction.lua
mheguy: You only ever see when the doc isn't updated :p

map_gen/shared/entity_placement_restriction.lua
mheguy: You're totally right. I never thought about the fact that all instances of an event share the same `event` table.. huh.

mheguy:
I never really noticed it in the API before, but should I be raising `script_raised_destroy` ([API](https://lua-api.factorio.com/latest/events.html#script_raised_destroy)), using my custom event only, or both?

grilledham (approved):
Test `map_gen.shared.danger_ore_banned_entities.lua` in single player and it seemed OK.

PR #724, opened 2019-02-02
Added groundhogs, updated dino_island

SimonFlapse:
Needs #723 
Plague edit: closes https://github.com/Refactorio/RedMew/pull/722


plague006:
Actually, this will close #723. :D

mheguy:
FP to catch up to dev

grilledham:
Maybe dino_islands with groundhogs should be a separate map, rather than changing dino_islands.

SimonFlapse:
Since dino_island needed to be updated anyways, I just added a boolean value to enable groundhogs. We can split them up if you'd still want that, when #723 has been merged

mheguy:
FP to get #723 in. 
The integration of `RestrictEntities` looks good to me.

SimonFlapse:
I've added a "toggle" for groundhogs. Minor testing, LGTM

mheguy:
I think the only changes remaining that aren't covered by the toggle are hail hydra, disabling loaders, and the tweaks to the ore numbers. I'll leave that for @grilledham to decide how to proceed.

grilledham:
I don't think we really have a good solution for toggles in presets available currently, so I'm willing to accept this as is.

PR #725, opened 2019-02-02
Refactor day_night

mheguy:
Makes day_night a little more accessible, cleans up the code a bit.

### DIFF
--- a/config.lua
+++ b/config.lua
@@ -267,6 +267,23 @@ global.config = {
     -- adds a useless button with the biter percentage
     evolution_progress = {
         enabled = true
+    },
+    -- sets the day/night cycle or a fixed light level. use_day_night_cycle and use_fixed_brightness are mutually exclusive
+    day_night ={
+        -- enables/disables the module
+        enabled = false,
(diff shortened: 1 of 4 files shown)


mheguy:
Rebase

config.lua
```diff
@@ -267,6 +267,23 @@ global.config = {
...
         enabled = true
+    },
+    -- sets the day/night cycle or a fixed light level. use_day_night_cycle and use_fixed_brightness are mutually exclusive
+    day_night ={
+        -- enables/disables the module
+        enabled = false,
+        -- for info on day/night cycles see https://github.com/Refactorio/RedMew/wiki/Day-Night-cycle
+        use_day_night_cycle = false,
+        day_night_cycle = {
+            ['ticks_per_day'] = 25000,
```
linaori: The `['']` are not needed for those keys and removing them will probably improve auto completion if I remember correctly. This syntax is only needed if otherwise invalid characters are used in the keys

linaori (approved):

config.lua
mheguy: You're right, ty for the catch

PR #729, opened 2019-02-02
Market tracking

mheguy:
Adds a command to wipe all markets created via the `/market` command.
^^ I finally got around to it @Jayefuu. It's not *exactly* what you wanted, but it's not terrible either.
Adds a `/destroy` command which is essentially just an alias to `game.player.selected.destroy()`
Pushes market buffs from global into Global 
Tidies up market.lua

It's amazing how much space you save reading a file when:

`global.player_speed_boost_records[player.index]`  
becomes  
`speed_records[p_name]`

Requires  #713

### DIFF
--- a/features/admin_commands.lua
+++ b/features/admin_commands.lua
@@ -331,6 +331,16 @@ local function revive_ghosts(args, player)
     end
 end
 
+local function destroy(_, player)
+    local ent = player.selected
+    if ent then
+        Game.player_print(ent.name .. ' destroyed')
+        ent.destroy()
+    else
+        Game.player_print('Nothing found to destroy.')
+    end
+end
+
 -- Event registrations
 
 Event.add(defines.events.on_built_entity, built_entity)
@@ -498,4 +508,13 @@ Command.add(
         required_rank = Ranks.admin,
     },
     Apocalypse.begin_apocalypse
+)
+
+Command.add(
+    'destroy',
+    {
+        description = 'Destroys the entity under your cursor',
+        admin_only = true
+    },
+    destroy
 )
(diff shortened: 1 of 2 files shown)


grilledham:
Why did you switch from using player indexes to player names?

linaori:
I think this doesn't close the GUI for players that have it open, please test this 😅

mheguy:
Index -> name
Faster lookup in the on_nth_tick function, which is the only time/place where we're going through the full list of records. Instead of sending each index to `Game.get_player_by_index` we can just grab the player directly and run our checks. That was the primary reason. But it also has a few nice knock-on effects:

Better speed in the boost functions since they were calling `player.name` a couple times already.
In the event of debugging, player names are easier to read than numbers.

mheguy:
> I think this doesn't close the GUI for players that have it open, please test this 😅

Discussed on discord, tested cases of buying and walking out of range of the (former) market. While not necessarily the intended behaviour, it doesn't seem to cause any issues.

plague006:
FP to get the market fix for boost death rebased in.

grilledham:
re Index -> name
But it does mean this breaks if a player's name changes while they have a boost.

One thing you could look at doing is unregistering the events when the boost records are empty.

mheguy:
You feel strongly about using index, I don't care enough either way so I'll use index. :)

> One thing you could look at doing is unregistering the events when the boost records are empty.

Danged good idea son.

mheguy:
Changed to require #713 so that I don't forget to change this later

grilledham (approved):

PR #731, opened 2019-02-02
Fix debugger

grilledham:
Fixes that the debbuger would crash if you tried to inspect a LuaObject that doesn't have a help function (remote, script) or a LuaObject that doesn't have a valid property (game).

Also applied the same fixes to Debug.object_type.

Change map_loader so that it loads the map selection first, so configs in map selection are applied before redmew_surface is loaded.


mheguy:

PR #732, opened 2019-02-02
Pling for toast

linaori:
Adds the following features:
 - Game persistent setting storage
 - A sound notification for toast messages

To change the volume of the toast messages, a new command can be used:
```
/redmew-setting-set toast-volume 0.5
```

To show the current volume:
```
/redmew-setting-get toast-volume
```

To show all available settings:
```
/redmew-setting-all
```

To make use of the new Settings system
```lua
local Settings = require 'utils.redmew_settings'
local toast_volume_name = 'toast-volume'
Settings.register(toast_volume_name, 'fraction', 1.0)

-- somewhere when the game is running
player.play_sound({
    path = 'some-sound', 
    volume_modifier = Settings.get(player.index, toast_volume_name)
})
```

In the future this can be enhanced by syncing the settings from and to the server, and making it possible top change them via a gui for example.

image
image
image

### DIFF
--- a/features/admin_commands.lua
+++ b/features/admin_commands.lua
@@ -8,6 +8,7 @@ local Event = require 'utils.event'
 local Command = require 'utils.command'
 
 local format = string.format
+local loadstring = loadstring
 
 --- A table of players with tpmode turned on
 global.tp_players = {}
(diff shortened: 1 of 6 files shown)


linaori:
Rebased against develop and tested locally, should be good to merge

grilledham (approved):

PR #737, opened 2019-02-03
Chunklist: Create new table instead of writing to and passing the old event table

mheguy:
I never put 2 and 2 together: the event module passes refs to each handler, but those refs are all to the same `event` table. So *maybe* I shouldn't be writing into a table that other modules are relying on. Especially since passing it through `raise_event()` overwrites some of the existing data.

### DIFF
--- a/map_gen/shared/chunklist.lua
+++ b/map_gen/shared/chunklist.lua
@@ -7,11 +7,9 @@
 local Global = require 'utils.global'
 local RS = require 'map_gen.shared.redmew_surface'
 local Event = require 'utils.event'
-local table = require 'utils.table'
 
 -- Localized functions
 local raise_event = script.raise_event
-local deep_copy = table.deep_copy
 
 -- Local vars
 local surface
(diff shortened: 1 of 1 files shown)


linaori:
I believe we add keys to existing events all the time after firing them again manually. I don't think _adding_ stuff will cause harm

mheguy:
I agree that adding keys isn't really an issue. However I think overwriting existing keys is really sloppy, and it can be avoided.
I had a pretty thorough look through the codebase and the chunklist module was the only instance of the `event` table being passed through `raise_event` (thereby overwriting `name`, `tick`, and `mod_name`).

grilledham:
It's probably best anytime you want to use raise_event to just make a new table with the data that should be in the event. 
```lua
local event = {key1 = data1, key2 = data2}
raise_event(event)
```
And let factorio fill in the name and tick for you.

I'd only worry about making shallow copies of tables as any handler that modifies data in the event table is doing it wrong.
With the exception of changes to userdata, for example `event.entity.destroy()`

grilledham (approved):

PR #738, opened 2019-02-03
Clean up global

mheguy:
Pushes `tp_players` from global to Global in admin_commands.lua
Pushes `paint_brushes_by_player ` from global to Global in paint.lua
Pushes multiple entries from global to Global in reactor_meltdown.lua

nuke_control.lua - consolidate multiple global entries into a single namespace

### DIFF
--- a/features/nuke_control.lua
+++ b/features/nuke_control.lua
@@ -75,37 +75,40 @@ local function item_not_sanctioned(item)
     return (name:find('capsule') or name == 'cliff-explosives' or name == 'raw-fish' or name == 'discharge-defense-remote')
 end
 
-global.entities_allowed_to_bomb = {
-    ['stone-wall'] = true,
-    ['transport-belt'] = true,
-    ['fast-transport-belt'] = true,
-    ['express-transport-belt'] = true,
-    ['construction-robot'] = true,
-    ['player'] = true,
-    ['gun-turret'] = true,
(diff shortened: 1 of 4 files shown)


features/nuke_control.lua
```diff
@@ -75,38 +75,41 @@ local function item_not_sanctioned(item)
...
-    ['electric-pole'] = true,
-    ['small-electric-pole'] = true,
-    ['medium-electric-pole'] = true,
-    ['big-electric-pole'] = true,
-    ['logistic-robot'] = true,
-    ['defender'] = true,
-    ['destroyer'] = true,
-    ['distractor'] = true
+global.nuke_control = {
+    entities_allowed_to_bomb = {
```
linaori, line 79: Is this really a global? Could possibly either use `Global` or be a local list

linaori (approved):

features/nuke_control.lua
mheguy: It doesn't need to be in `global`, but it does need to be accessible. Every now and again there's a map that does things a little different and we end up having to modify the list during runtime.

When 0.17 hits with its undo features, we will definitely want to look at refactoring antigrief and nuke_control together. This is just a stopgap measure to make `global` a little more tidy.

linaori: Let's keep it as is then

PR #742, opened 2019-02-04
Apply lifecycle

mheguy:


features/force_control.lua
```diff
@@ -76,7 +76,7 @@ end
 ---@param callback function function(number level_reached, LuaForce force)
 ---@param lua_force_name string|nil only register for this force (optional)
 function ForceControlBuilder.register(level_matches, callback, lua_force_name)
-    if game then
+    if _LIFECYCLE <= 4 then
```
grilledham: Isn't this check backwards, shouldn't it be `if _LIFECYCLE > 4 then`?

features/gui/evolution_progress.lua
```diff
@@ -25,7 +25,7 @@ local button_sprites = {
 }
 
 local function get_evolution_percentage()
-    if not game then
+    if _LIFECYCLE <= 4 then
```
grilledham: I think this would error if called in `on_load` 
I'm thinking about how to do this case with `_LIFECYCLE` and I'm thinking `not game` is just simpler.

features/server.lua
```diff
@@ -335,6 +335,12 @@ end
 --     end
 -- )
 function Public.on_data_set_changed(data_set, handler)
+    if _LIFECYCLE == 8 then
+        error('cannot call during runtime', 2)
+    end
+    if _LIFECYCLE > 4 then
+        log('Should not call on_data_set_changed after the control stage.')
+    end
```
grilledham: I'm not really a fan of logging warning for things that might be fine. If people know what they are doing we should just let them.

map_gen/maps/crash_site.lua
```diff
@@ -128,7 +128,7 @@ local spawn_callback =
 )
 
 local function init()
-    local on_init = game ~= nil
+    local on_init = _LIFECYCLE > 4
```
grilledham: I think this is wrong, `on_init` needs to be true in stage 5 and false in stage 6.
So it should be  `on_init = _LIFECYCLE  == 5`. Taking advantage of the fact `init()` is only called in stage 5 and 6

map_gen/shared/day_night.lua
```diff
@@ -55,12 +55,15 @@ Event.on_init(init)
...
--- Can only be used during or after init.
+-- Can only be called during or after init.
 -- @param day_night_cycle <table> containing specific, required keys: ticks_per_day, dusk, evening, morning, dawn
 -- @param surface <LuaSurface> to set the day/night cycle of
 -- @returns <boolean> true if set properly, nil if not
 -- @see Venus::world_settings
 function Public.set_cycle(day_night_cycle, surface)
+    if _LIFECYCLE < 5 then
+        error('cannot call before on_init', 2)
+    end
```
grilledham: I'm wondering if this check isn't covered by the `if not surface or not surface.valid then` below.
In the sense of how would you of gotten this far to call this function with an invalid surface.
Also this then might apply to the two cases below as well.

utils/global.lua
```diff
@@ -4,6 +4,9 @@ local Token = require 'utils.token'
 local Global = {}
 
 function Global.register(tbl, callback)
+    if _LIFECYCLE == 4 then
+        error('can only be called during the control stage', 2)
+    end
```
grilledham, line 9: Shouldn't this be `if _LIFECYCLE ~= 4 then`?

utils/global.lua
```diff
@@ -14,6 +17,9 @@ function Global.register(tbl, callback)
 end
 
 function Global.register_init(tbl, init_handler, callback)
+    if _LIFECYCLE == 4 then
+        error('can only be called during the control stage', 2)
+    end
```
grilledham, line 22: Shouldn't this be `if _LIFECYCLE ~= 4 then`?

utils/gui.lua
```diff
@@ -183,6 +183,9 @@ Gui.on_pre_player_hide_top = custom_handler_factory(on_pre_hidden_handlers)
 -- This function must be called in the control stage, i.e not inside an event.
 -- @param element_name<string> This name must be globally unique.
 function Gui.allow_player_to_toggle_top_element_visibility(element_name)
+    if _LIFECYCLE > 4 then
+        error('cannot call after the control stage', 2)
+    end
```
grilledham, line 188: For consistently with above maybe we should change this to:
```lua
if _LIFECYCLE ~= 4 then
    error('can only be called during the control stage', 2)
end
```

utils/task.lua
```diff
@@ -77,6 +77,9 @@ end
 -- @param func_token <number> a token for a function store via the token system
 -- @param params <any> the argument to send to the tokened function
 function Task.set_timeout_in_ticks(ticks, func_token, params)
+    if _LIFECYCLE < 5 then
+        error('cannot call before on_init', 2)
+    end
```
grilledham, line 82: I think calling `set_timeout_in_ticks` before `on_init` should be fine. It would be a problem to call it `on_load` though.

utils/task.lua
```diff
@@ -88,6 +91,9 @@ end
 -- @param func_token <number> a token for a function store via the token system
 -- @param params <any> the argument to send to the tokened function
 function Task.set_timeout(sec, func_token, params)
+    if _LIFECYCLE < 5 then
+        error('cannot call before on_init', 2)
+    end
```
grilledham, line 96: Same as above 
Calling `set_timeout` before `on_init` should be fine. It would be a problem to call it `on_load` though.

grilledham (dismissed):

utils/task.lua
mheguy: It's not fine to call before on_init
```lua
function Task.set_timeout_in_ticks(ticks, func_token, params)
    local time = game.tick + ticks
```
No `game` before init.

utils/task.lua
mheguy: It's not fine to call before on_init
```lua
function Task.set_timeout_in_ticks(ticks, func_token, params)
    local time = game.tick + ticks
```
No `game` before init.

features/gui/evolution_progress.lua
plague006: `<= 4 and not 6`, but I'll just stick with not game

features/server.lua
mheguy: I'll remove it, but we are letting them. My thought was just documenting the risky behaviour (that it might not be registered for both init and load)

mheguy:
Thanks grilled, doing all these started really playing with my mind.

utils/task.lua
grilledham: Agreed. I was mistaken.

utils/task.lua
grilledham: Agreed. I was mistaken.

grilledham (approved):
This LGTM 👍

PR #743, opened 2019-02-08
Create market on launch

mheguy:
Creates a market on init.

Checked for maps that called retailer and for maps that used the term 'market' and where appropriate, toggled the config option.

No more will people have to whine about me forgetting to plop a market on maps. :D

### DIFF
--- a/config.lua
+++ b/config.lua
@@ -88,6 +88,8 @@ global.config = {
     -- adds a market
     market = {
         enabled = true,
+        -- will create a standard market on game startup at the {0,-2} coordinates
+        create_standard_market = true,
         currency = currency,
 
         -- defines the chance that killing an entity will drop coins and the min and max it can drop upon death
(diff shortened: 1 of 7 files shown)


config.lua
```diff
@@ -83,6 +83,8 @@ global.config = {
     -- adds a market
     market = {
         enabled = true,
+        -- will create a standard market on game startup
+        create_standard_market = true,
```
linaori, line 92: What if we allow `false` or coordinates here instead? That means a scenario can decide where to put the market

mheguy: A boolean or a table in a config option seems a touch confusing for a layperson tbh. How about I just add a field for coordinates?

grilledham (dismissed):

mheguy:
market.lua was badly conflicted and I could use another look at it again. I'll be able to test tomorrow.

linaori (approved):
Code wise the merge seems to be alright

grilledham (approved):

PR #744, opened 2019-02-08
Add 1000x

mheguy:
Also adds infinite storage chest's cost to config

### DIFF
--- a/map_gen/maps/1000x.lua
+++ b/map_gen/maps/1000x.lua
@@ -97,7 +97,7 @@ RS.set_map_gen_settings(
         MGSP.sand_only
     }
 )
-RS.set_difficulty_settings({technology_price_multiplier = tech_cost})
+RS.set_difficulty_settings({{technology_price_multiplier = tech_cost}})
 RS.set_map_settings({map_settings})
 
 -- Setup market inventory
--- a/map_gen/maps/1000x.lua
+++ b/map_gen/maps/1000x.lua
@@ -187,7 +187,7 @@ Event.on_init(
         force.technologies['turrets'].researched = true
         force.technologies['military'].researched = true
         force.technologies['logistics'].researched = true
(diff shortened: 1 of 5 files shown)


map_gen/maps/1000x.lua
```diff
@@ -0,0 +1,380 @@
...
+    function()
+        local force = game.forces.player
+        local surface = RS.get_surface()
+        local pos = {0, -15}
+
+        -- Techs
+        force.technologies['automation'].researched = true
+        force.technologies['logistics'].researched = true
+        force.technologies['turrets'].researched = true
+        force.technologies['military'].researched = true
```
grilledham, line 174: Because events are not raised during `on_init` this failed to enabled loaders. We have 3 options:
1. Raise the event manually.
2. `if loader.enabled then enable them`
3. In redmew_qol.lua have some code that run on tick 1 and checks if the techs have been enabled.

map_gen/maps/1000x.lua
```diff
@@ -0,0 +1,383 @@
...
+        local force = game.forces.player
+        local surface = RS.get_surface()
+        local pos = {0, -15}
+
+        -- Techs
+        force.technologies['automation'].researched = true
+        force.technologies['turrets'].researched = true
+        force.technologies['military'].researched = true
+        force.technologies['logistics'].researched = true
+        if config.redmew_qol.enabled and config.redmew_qol.loaders.enabled then
```
grilledham: `config.redmew_qol.loaders.enabled` should be `config.redmew_qol.loaders`

mheguy:
Going to merge this. I'm hoping that in the future @Jayefuu or some other kind soul will redo the oregen. :D

map_gen/maps/1000x.lua
mheguy: Raising the event manually during on_init will only fix things until 0.17 which will remove that bug.

I'll PR the change for the qol module

PR #745, opened 2019-02-08
Add apocalypse

mheguy:
Refactored hail_hydra slightly to allow it to be turned on and off during runtime. The diff on github looks weird, but none of the code inside any of the functions was altered.

This command is meant to be run at the end of a map (preferably after the final save has been made).

It turns evolution up to 1.0, it enables hydra mode and changes the hydra spawn rate so that each killed behemoth biter respawns *and* has a 1% chance to duplicate.

After a 60 sec timeout, it spawns 4 behemoth biters at player spawn.

Depends #713


Jayefuu:
HA. That's awesome. 

Unrelated.... it would be fun to repeat that crash site apocalypse you, me and Newcott did that time.

Jayefuu:
One thing to consider with this feature is that we often move the completed save to public_saves/final after we finish with a new challenge map.

For this reason might it be useful to add code to only allow this command on certain maps? I can see it being very popular on crash site and deathworld maps but would be wary that it would ruin the final map save if someone or the script didn't make a copy before or during running /apocalypse.

features/apocalypse.lua
```diff
@@ -0,0 +1,98 @@
...
+    Token.register(
+    function(data)
+        local surface = data.surface
+        local group = data.group
+        local p_spawn = data.p_spawn
+
+        local create_entity = surface.create_entity
+
+        for i = 1, 2 do
+            local spawn_pos = surface.find_non_colliding_position('behemoth-biter', p_spawn, 300, 1)
```
linaori:
```lua
local aliens = {
    'behemoth-biter',
    'behemoth-biter',
    'behemoth-spitter',
    'behemoth-spitter',
}
```
Then you can iterate over it.

features/apocalypse.lua
```diff
@@ -0,0 +1,98 @@
...
+        player_force = player.force
+    else
+        surface = RS.get_surface()
+        player_force = game.forces.player
+    end
+
+    player_force.recipes['atomic-bomb'].enabled = false
+    for _, p in pairs(game.connected_players) do
+        p.remove_item({name = 'atomic-bomb', count = 1000})
+    end
```
grilledham: What about atomic-bombs stored in chests, the output slot of assembly machines or on the ground?

Thinking about it again. It might just be easier to remove nukes when they are equipped, like nuke_control does

linaori: Mini-game, hide and seek :joy:

features/apocalypse.lua
mheguy: Thanks. A previous iteration had 20 biters and 10 spitters so changing how I did wasn't something I put any thought into.

features/apocalypse.lua
mheguy: I considered scanning around for them, but I've seen that fail in another mod due to bots and belts not being checked.

Ultimately though, nukes (and the removal thereof) were more of an issue in a previous iteration of the feature so I'm removing this bit of code entirely.

mheguy:
Adding a dependency on #713 which will cause linting to fail, but will save me from having to change it later.

grilledham (approved):

PR #747, opened 2019-02-09
Add everyone is regular option

muppet9010:
Adds a config option so that all players are deemed to be regulars. This way on private servers all players can do all actions. The default value is off as per current scenario behaviour.

### DIFF
--- a/config.lua
+++ b/config.lua
@@ -284,7 +284,11 @@ global.config = {
         -- brightness is a number between 0.15 and 1
         use_fixed_brightness = false,
         fixed_brightness = 0.5
-    }
+    },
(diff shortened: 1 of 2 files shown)


grilledham:
I think this is a good idea. For people who want to play private games of the scenario the regulars don't make a lot of sense. 
And making everyone a regular is simpler than adding extra checks else where in the code.

This might need reworking after #713 is merged, but I will keep this in mind.

mheguy:
Added into https://github.com/Refactorio/RedMew/pull/713 (with authorship attribution, hope you're cool with that @muppet9010)

muppet9010:
Yea is cool.

PR #750, opened 2019-02-12
Fix print_except

mheguy:
Fixes https://github.com/plague006/RedMew/blob/596f329337c0af54df9eb79892bcfb9936cc6782/features/player_colors.lua#L61 by putting the optional component as the third parameter

### DIFF
--- a/utils/core.lua
+++ b/utils/core.lua
@@ -34,7 +34,7 @@ end
 --- Takes msg and prints it to all players except provided player
 -- @param msg <string> The message to print
 -- @param player <LuaPlayer> the player not to send the message to
--- @param color <table> the color to use for the message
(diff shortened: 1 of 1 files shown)


utils/core.lua
```diff
@@ -33,9 +33,9 @@ end
 
 --- Takes msg and prints it to all players except provided player
 -- @param msg <string> The message to print
--- @param color <table> the color to use for the message
 -- @param player <LuaPlayer> the player not to send the message to
-function Module.print_except(msg, color, player)
+-- @param color <table> the color to use for the message
```
grilledham: We could mention this defaults to Color.white if omitted.

grilledham (approved):

PR #756, opened 2019-02-15
Fix/apocalypse hail hydra desync

mheguy:
apoc: set data in token rather than passing data, move to Global, change command from some silly string to running it twice. This also gives us a chance to create a save that can be reverted to. Move strings to localised strings

hydra: move data into Global to allow changes during runtime (that don't cause desyncs)

Mostly tested in SP with heavy mode. Would benefit from re-testing as I only tested https://github.com/Refactorio/RedMew/commit/41c88ec682d3f59c5f8dae6f814379b377fb3f12 without heavy mode.

### DIFF
--- a/features/admin_commands.lua
+++ b/features/admin_commands.lua
@@ -503,9 +503,7 @@ Command.add(
 Command.add(
     'apocalypse',
     {
-        description = "Calls for the endtimes. This really ends the map, so you must use '/apocalypse end this map'",
-        arguments = {'confirmation'},
-        capture_excess_arguments = true,
+        description = "This really ends the map. When you first run it, the game will save. When run a second time, the apocalypse will begin.",
         required_rank = Ranks.admin,
     },
     Apocalypse.begin_apocalypse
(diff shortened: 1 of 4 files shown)


grilledham:
> hydra: move data into Global to allow changes during runtime (that don't cause desyncs)

This confuses me, wasn't hydra already in global?

linaori (approved):

mheguy:
I'm punching above my weight class trying to describe this, but here goes my attempt:

Hydras were in `global` but the local refs were breaking I think?

Ignoring the apocalypse module entirely, if you simply run `/c global.config.hail_hydra.hydras['behemoth-biter'] = {['behemoth-biter'] = 0.01}` you can get a desync.
When I changed this line <https://github.com/Refactorio/RedMew/pull/756/files#diff-49feeb962373b0ff53ce316e5fbf3877L54> to `local hydra = global.config.hail_hydra.hydras[name]` the desync went away both with the console LUA I pasted as well as the apocalypse module.

While that fixed it, it didn't feel like a strong long-term solution, so I plopped things into `Global` and made some setters.

grilledham (approved):

PR #727, opened 2019-02-02
Donator commands

mheguy:
~~Requires https://github.com/Refactorio/RedMew/pull/713~~
~~Requires https://github.com/Refactorio/RedMew/pull/757~~
The following maps will need to shut down before I can transition to tables:
- [x] Redmew - Valentine's Day
- [x] Redmew - Crash Site
- [x] Redmew - Diggy - Redux
- [x] RedMew - Line and Trees
- [x] RedMew - Tetris

Closes #733

PR #757, opened 2019-02-15
Have donator handle string and table.

mheguy:
This opens the way to allowing for welcome_messages to hold a table of possible join messages as in https://github.com/Refactorio/RedMew/pull/727

This PR in particular is to allow us a graceful transition where both data types are handled properly.

### DIFF
--- a/features/donator.lua
+++ b/features/donator.lua
@@ -57,9 +57,10 @@ local function player_joined(event)
     end
 
     local message
-    if type(messages) == 'string' then
+    local messages_type = type(messages)
+    if messages_type == 'string' then
         message = messages
-    elseif type(messages) == 'table' then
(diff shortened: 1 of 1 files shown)


features/donator.lua
```diff
@@ -48,11 +51,18 @@ local function player_joined(event)
...
 
-    local message = d.welcome_messages
-    if not message then
+    local messages = d.welcome_messages
+    if not messages then
         return
     end
 
+    local message
+    if type(messages) == 'string' then
```
grilledham: You could make a `local messages_type = type(messages)` to avoid getting the type twice.

grilledham (approved):

features/donator.lua
mheguy: Derp. ty

PR #759, opened 2019-02-16
left align rank cells

grilledham:
I would like to test this with multiple people to confirm that this change does look better.


mheguy:
You're a monster for not including screenshots. :p

mheguy:
Before
image

After:
image

mheguy:
I believe "Regular (D)" would be the longest possible string in that column. So we could trim the width if that seems appropriate.

Can we also have a popup for it like deaths has that reveals that it is a donator tag?

grilledham:
The longest possible string is `Auto Trusted (D)` I set the column width for that case.

mheguy:
I will always forget that bastard rank. lol

PR #761, opened 2019-02-16
Fix unjail from report not working

mheguy:
Hacky fix, but it should hold together until we either redo all griefing modules and/or until we move to localizedstrings.


grilledham (approved):
Tested in mp, didn't find any issues

PR #762, opened 2019-02-17
Add toast command

mheguy:


plague006:
Causes desyncs in MP. Will comment again when fixed.

mheguy:
Works fine now.

grilledham:
We could consider renaming 'toast-all' to just 'toast' to make it consistent with 'popup'. Or we change 'popup' to 'popup-all'

mheguy:
Tested in SP and MP.

features/gui/toast.lua
```diff
@@ -247,4 +252,40 @@ function Public.toast_all_players(duration, message)
...
+        allowed_by_server = true
+    },
+    function(args)
+        local player_name = args.player
+        local player = game.players[player_name]
+        if player then
+            Public.toast_player(player, 15, args.msg)
+            Utils.print_admins(format('%s sent a toast to %s', Utils.get_actor(), player_name))
+        else
+            Game.player_print({'common.fail_no_target'}, Color.yellow)
```
grilledham: Currently if you accidentally put a space in the player name like so 
`/toast-player grilledha m some message`
You get 'Target not found'

I feel this message could be better. Maybe something like 'player <player_name> not found'

grilledham (approved):

features/gui/toast.lua
mheguy: You're right the message was weak sauce. I took it from another command without thought to improving it.

mheguy:
After refining error: Ran command once properly and once with error, both acted as expected.

PR #764, opened 2019-02-17
Split commands into their respective modules

mheguy:
No actual new code, just splitting things out.


mheguy:
Tested in SP with success.

grilledham (approved):

PR #765, opened 2019-02-17
Fix bug preventing jailing of a player in a train

mheguy:
Came up during the testing of #702: Trains cannot have their speed set via `LuaEntity.speed`


mheguy:
Had Val throw an eye on this since it's 4:30am. :D

PR #766, opened 2019-02-17
Changed error when no map is selected

Valansch:
I intentially didnt say how to select a map because that would have made the error message bigger and the bigger the error message the lower the chance that it will be read.

@plague006 I am sorry Father for I have sinned.

### DIFF
--- a/map_gen/shared/map_loader.lua
+++ b/map_gen/shared/map_loader.lua
@@ -1,6 +1,6 @@
-local success, shape = pcall(function() require 'map_selection' end)
+local success, shape = pcall(function() return require 'map_selection' end)
 if not success then
-	error('\n NO MAP SELECTED \n The RedMew scenario contains many different maps. If you are unsure how to select a map checkout our guide at: \n https://redmew.com/guide \n -The RedMew Team ', 0)
+	error('\n\n--map_selection.lua not found--\nReleases on github include this file.\nFor info: \nhttps://redmew.com/guide\n - The RedMew Team ', 0)
 end
 
 local b = require 'map_gen.shared.builders'
--- a/map_gen/shared/map_loader.lua
+++ b/map_gen/shared/map_loader.lua
@@ -1,8 +1,7 @@
-local shape = require 'map_selection'
-local b = require 'map_gen.shared.builders'
-local RS = require 'map_gen.shared.redmew_surface'
-local config = global.config.map_generation
-
+local success, shape = pcall(function() require 'map_selection' end)
+if not success then
+	error('\n NO MAP SELECTED \n The RedMew scenario contains many different maps. If you are unsure how to select a map checkout our guide at: \n https://redmew.com/guide \n -The RedMew Team ', 0)
+end
 if #config.entity_modules > 0 then
     shape = shape or b.full_shape
     shape = b.apply_entities(shape, config.entity_modules)
--- a/map_gen/shared/map_loader.lua
+++ b/map_gen/shared/map_loader.lua
@@ -2,6 +2,11 @@ local success, shape = pcall(function() require 'map_selection' end)
 if not success then
 	error('\n NO MAP SELECTED \n The RedMew scenario contains many different maps. If you are unsure how to select a map checkout our guide at: \n https://redmew.com/guide \n -The RedMew Team ', 0)
 end
+
+local b = require 'map_gen.shared.builders'
+local RS = require 'map_gen.shared.redmew_surface'
+local config = global.config.map_generation
+
 if #config.entity_modules > 0 then
     shape = shape or b.full_shape
     shape = b.apply_entities(shape, config.entity_modules)


map_gen/shared/map_loader.lua
```diff
@@ -1,4 +1,8 @@
-local shape = require 'map_selection'
+local success, shape = pcall(function() require 'map_selection' end)
+if not success then
+	error('\n NO MAP SELECTED \n The RedMew scenario contains many different maps. If you are unsure how to select a map checkout our guide at: \n https://redmew.com/guide \n -The RedMew Team ', 0)
```
mheguy: Please tell them that the error is that they downloaded the repository instead of a release. If you recall, one of the major reasons for not including `map_selection.lua` is so that they download a release which has a version number.

mheguy (changes requested):

map_gen/shared/map_loader.lua
Valansch: But what if they want to tinker with it? How does a release help? We have so many maps this would limit them from only using the ones in our releases.

map_gen/shared/map_loader.lua
```diff
@@ -1,4 +1,8 @@
-local shape = require 'map_selection'
+local success, shape = pcall(function() require 'map_selection' end)
```
plague006: `return require 'map_selection'`

mheguy: return require 'map_selection'

mheguy (dismissed):

map_gen/shared/map_loader.lua
mheguy: Per discussion on discord:
```
\n\n--map_selection.lua not found--\nReleases on github include this file.\nFor info: \nhttps://redmew.com/guide
```

map_gen/shared/map_loader.lua
Valansch: 🤦‍♂️

Valansch:
@plague006 resolved

grilledham:
As a developer the `pcall` is very annoying as it turns the useful error message into a message about not loading a map.
Can we change map_loader.lua to this:
```lua
local shape = require 'map_selection'
if shape == true then
	error('\n\n--map_selection.lua not found--\nReleases on github include this file.\nFor info: \nhttps://redmew.com/guide\n - The RedMew Team ', 0)
end

local config = global.config.map_generation

if #config.entity_modules > 0 then
    shape = shape or b.full_shape
    shape = b.apply_entities(shape, config.entity_modules)
end

if #config.terrain_modules > 0 then
    shape = shape or b.full_shape

    for _, m in ipairs(config.terrain_modules) do
        shape = b.overlay_tile_land(shape, m)
    end
end

if type(shape) == 'function' then
    local surfaces = {
        [RS.get_surface_name()] = shape
    }

    local gen = require('map_gen.shared.generate')
    gen.init({surfaces = surfaces, regen_decoratives = config.regen_decoratives, tiles_per_tick = config.tiles_per_tick})
    gen.register()
end
```
And map_gen.maps.default.lua to:
```lua
-- A blank file to allow map_selection to have a default
return false
```
Or am I missing something?

mheguy:
Val's initial comment/proposal was off the mark in my opinion. This PR/issue is not about selecting a map. We have the default map because we explicitly don't require them to select a map. 

The issue this PR seeks to resolve is for the following use-case: an end user unfamiliar with LUA and coding in general downloads the repository instead of a release. They have no `map_selection.lua` and they run the scenario. They are greeted with a big unfriendly error message which means nothing to them.

This PR basically just changes the error from this:
image

to this:
image

On a related note, we could also expand the error message we give for no map required. That case is handled in the penultimate line of code in the file: `error('You forgot to require a map')`

grilledham:
To be clear my issue with this current implementation is the suppressing of errors to devs.

Suppose I make this simple syntax error, or any other error for that matter.
image

Before we would get an error message telling us where the mistake was.
image

Now we get a useless and misleading message. 
image
This would be particularly confusing for a dev who wasn't in the loop about this change.

I think my proposal above gives, us what we want. But really I'd be happy with any proposal that doesn't suppress the error message when a map is loaded correctly, but there is an error in it.

linaori:
We should check what error it is with pcall and give the proper message (like how exceptions work)

mheguy:
You're right about errors in the map functions. I completely failed to consider that.

```if shape == true then error```
`require` returns `true` if a map has a nil return, so we would have to make sure all maps that do not return a function explicitly return `false`

grilledham:
> `if shape == true then error`
`require` returns `true` if a map has a nil return, so we would have to make sure all maps that do not return a function explicitly return `false`

I agree that is annoying, we need a better solution.

mheguy:
If the file `map_selection.lua` exists then require will return the map's value or true if nil.
So we could force all maps to not return nil and we check for that.
pcall can give a friendly warning to end users downloading the repository but will suppress map errors.

I think in this case the greatest benefit with the lowest harm is to do nothing but include a note at the top of the readme so that users ending up in our repo have the best chance of seeing it. If I'm being realistic though, end users *will not* download the repository, nor will they download source code from releases.

Eventually we will figure out deflating dynamic requires and that will give us more options.

mheguy:
Closing after speaking with Val on discord: no good way to fix this atm.

PR #768, opened 2019-02-17
Quadrants

### LINKED ISSUE #714: Four Quadrants - Each quadrant produces different things and feeds each-other
Relatively simple premise. There are four quadrants to the map, with a large void divider (horizontal and vertical) that intersects the map center.

Either the divider would be sufficiently thick to prevent roboport coverage reaching the other side, with belts as the transfer medium, or it would be 4 tiles thick allowing players to direct transfer across the void and create undergrounds where needed instead of in preset locations.

Each quadrant would have a different set of recipe restrictions, so that all four quadrants are required to get anything done.

All quadrants would have access to "vital" recipes, such as inserters, mining drills, furnaces. This is to ensure a quadrant can get set up without relying on another quadrant.

Everything else is distributed to quadrants, possibly randomly, to force exchanging products. Green circuit recipes might only be available to the left two quadrants, whilst red circuits are only available to the bottom-right quadrant.

SimonFlapse:
Based on the map suggestion in #714

Four isolated quadrants must come together to launch a rocket

- [x] Resource transfer between quadrants
- [x] Prevent logistical and electrical network connection between quadrants
- [x] Teleportation between quadrants from spawn area (Only if empty inventory)
- [x] Identical spawn 
- [x] Oil patch for quadrant 3
- [x] Recipe restrictions on each quadrant (Currently broken in MP)
- [x] Research restricted to quadrant 1
- [x] 4 friendly forces
- [x] Increased ore count for quadrant 3
- [x] Restricted building capabilities on other quadrants
- [x] Automatic inventory empty to chest
- [x] Adapt recipes for 0.17 changes
- [x] Added train crossings
- [x] Added force_sync for chat and chart tags
- [x] Added informational compilatrons
- [x] Converted strings to localizedString
- [x] Added danish translation

Needs #795 

Closes #714


SimonFlapse:
I now see that I actually deleted the test.lua map during the initial creation of this map. I can't see the usage of test.lua anyways, so unless we'd like it, I'd just leave it as deleted.

SimonFlapse:
I hope to run a play test before 0.17 is out.

SimonFlapse:
**Quadrant selection GUI:**
image
**Map view:**
image

map_gen/maps/quadrants/scenario.lua
```diff
@@ -0,0 +1,274 @@
...
+
+local start_iron = b.resource(rectangle, 'iron-ore', constant(750))
+local start_copper = b.resource(rectangle, 'copper-ore', constant(600))
+local start_stone = b.resource(rectangle, 'stone', constant(300))
+local start_coal = b.resource(rectangle, 'coal', constant(450))
+local start_tree_1 = b.entity(tree_rectangle_1, 'tree-01')
+local start_oil = b.resource(oil_rectangle, 'crude-oil', b.exponential_value(100000, 0, 1))
+local start_tree_2 = b.entity(tree_rectangle_2, 'tree-01')
+
+start_iron = b.combine({ b.translate(start_iron, base_x, base_y), b.translate(start_iron, -base_x, -base_y), b.translate(start_iron, base_x, -base_y), b.translate(start_iron, -base_x, base_y) })
```
grilledham: I think you want `b.any` rather than `b.combine`

SimonFlapse: I'm actually not sure of the difference (For the guide) but yeah I'd any too now after having looked closer at the builder

SimonFlapse:
I've added the ability to empty your inventory into created chests.
Considerations:
- Locking it behind donator rank and having to buy it in the market
- Can spawn chests on tutorial tiles
- Can be used like a regular chest, giving the player an early steel chest (Cannot be mined)
- Might spawn in another quadrant if it cannot find a non colliding position

image

config.lua
```diff
@@ -1,5 +1,5 @@
-_DEBUG = false
-_CHEATS = false
+_DEBUG = true
+_CHEATS = true
```
mheguy: You can (and should) set all your config options in `map_selection.lua` when you're testing. :D

map_gen/maps/quadrants/switch_team.lua
```diff
@@ -28,6 +28,8 @@ Global.register(
     end
 )
 
+local player_switched_force = script.generate_event_name()
```
mheguy: I'm confused.. It looks like you created an event to call a local function?

map_gen/maps/test.lua
```diff
@@ -1,13 +0,0 @@
-local b = require 'map_gen.shared.builders'
```
mheguy: Why are you deleting this?

SimonFlapse: Started out as a mistake, because I used that file to make the map gen before I moved it into a folder. This caused a conflict an I ended up deleting it instead of discarding. I can bring it back if we need it.

mheguy: Yes please. It's handy for what it is: a quick way to test worldgen.

SimonFlapse: I use default for that, but that's not the right way I reckon

map_gen/maps/quadrants/switch_team.lua
SimonFlapse: I guess I just wanted to try it out :P But in retro perspective it has no uses since I should use on_player_changed_force

map_gen/maps/quadrants/switch_team.lua
```diff
@@ -93,7 +93,6 @@ local function teleport(event, quadrant)
             nil,
             'Quadrants.quadrant_description'
         )
-        script.raise_event(player_switched_force, {extra = 'data'})
```
mheguy: You could just run `update_gui(nil, true)` here, no?

SimonFlapse: I already fixed that?

mheguy: I should have been more clear:
Your solution was to create a function `changed_force` that calls another function `update_gui`. Why not just call `update_gui` directly?

SimonFlapse: Well I'm not sure why I did it. I just had in mind that I wanted to try to use a custom event

mheguy (approved):

map_gen/maps/quadrants/force_sync.lua
```diff
@@ -109,8 +109,7 @@ local function research_finished(event)
 
     for _, force in pairs(game.forces) do
         if (string.find(force.name, 'quadrant')) ~= nil then
-            force.print(
-                '[img=item/automation-science-pack] New research complete: [technology=' .. technology.name .. ']'
+            force.print({'quadrants.force_sync_research', technology.name}
```
mheguy, line 112: Rather than the tech name, you can take the tech’s localized name.

mheguy (commented):
On phone, but looks good overall.

map_gen/maps/quadrants/force_sync.lua
SimonFlapse: I'm trying to display the technology sprite in chat, it'll convert it to localized string by itself when showing the tooltip for the technology

mheguy: Ah ok, I get you. Cool.

PR #769, opened 2019-02-18
New year, new banner

mheguy:
image


theorangeangle:
This is the type of thing I normally stay out of, but is it just me or does the yellow outline of redmew look a little weird? Maybe we can get Z4rg0n to approve as wekk.

Zr4g0n:
I suggest this, and I'm open to further feedback and changes. 
[Image here](https://media.discordapp.net/attachments/432568711698710540/547176266437230593/unknown.png?width=1203&height=282)

SimonFlapse:
Zr4g0n is proposing something like this:
image

PR #770, opened 2019-02-18
After init, check if loaders should be enabled.

mheguy:
Tested with success in SP.


grilledham (approved):
Maybe after 0.17 comes out and we can see what the state of play is regarding raising events during `on_init`, we could look at raising the technology researched event instead.

PR #773, opened 2019-02-19
Toxic_science_jungle.lua

theorangeangle:
A redux of toxic jungle with more water, different ore distribution and disabled recipes for military, production, and gold science (must be bought at the market) along with flamethrowers.

Fingers crossed I did this right 🤞


plague006:
Note to whoever merges: squash merge and write a sane commit message.

map_gen/maps/toxic_science_jungle.lua
```diff
@@ -0,0 +1,254 @@
...
+
+map = b.change_map_gen_tile(map, 'water', 'water-green')
+map = b.change_map_gen_tile(map, 'deepwater', 'deepwater-green')
+
+map = b.apply_entity(map, enemy)
+
+local function on_init()
+    local surface = RS.get_surface()
+    local player_force = game.forces.player
+    player_force.recipes['military-science-pack'].enabled = false
```
mheguy: Recipes become enabled by research. So once military-2 (or whatever) research is complete, the recipe will be enabled.

map_gen/maps/toxic_science_jungle.lua
```diff
@@ -0,0 +1,254 @@
...
+    player_force.recipes['production-science-pack'].enabled = false
+    player_force.recipes['high-tech-science-pack'].enabled = false -- disable crafting of sciences
+    game.forces.player.technologies['flamethrower'].enabled = false -- disable flamethrower tech
```
mheguy: Earlier you have `local player_force = game.forces.player` where you take the player force and put it in the variable `player_force`. That makes `game.forces.player` redundant here as you can just use `player_force`

map_gen/maps/toxic_science_jungle.lua
```diff
@@ -0,0 +1,254 @@
...
+    player_force.manual_mining_speed_modifier = 2 -- increase mining speed, disabled after military 2 research
+    game.map_settings.enemy_expansion.enabled = true
+
+    -- Set up non-standard market so we can add science packs for purchase while keeping all other items
+    global.config.market.create_standard_market = false
```
mheguy: It's good practice to not touch `global` after the control stage of the data cycle. What this means for you is that you should have this line outside of any function. Probably around line 14 where you're also accessing the `global` table.

The reason you shouldn't be accessing the `global` table after the control stage is that unless you're very sure which order files were loaded in: the variable you're setting might already be used.

More info on the data cycle: https://lua-api.factorio.com/latest/Data-Lifecycle.html

map_gen/maps/toxic_science_jungle.lua
```diff
@@ -0,0 +1,254 @@
...
+    Retailer.set_item('items', {price = 10, name = 'construction-robot'})
+
+    Retailer.set_market_group_label('items', 'Items Market')
+    local item_market_1 = surface.create_entity({name = 'market', position = {0, 0}})
+    item_market_1.destructible = false
+    Retailer.add_market('items', item_market_1)
+end
+Event.on_init(on_init)
+
+table.insert(global.config.player_create.starting_items, {name = 'steel-axe', count = 2})
```
mheguy: This can go at the top with your other config setting (L14)

mheguy (dismissed):

map_gen/maps/toxic_science_jungle.lua
theorangeangle: so i need to disable the recipe when the research completes, rather than just from the map start?

map_gen/maps/toxic_science_jungle.lua
theorangeangle: resolved with 72f1cb1

map_gen/maps/toxic_science_jungle.lua
theorangeangle: i guess it's already in line 14 so i don't need it twice right?

map_gen/maps/toxic_science_jungle.lua
theorangeangle: i'm confused. moving lines 250-252 to below L14 gives two errors:

`accessing undefined variable 'on_init' luacheck(113) [16,15]`
and then `unused function 'on_init' luacheck(211) [154,16]`

map_gen/maps/toxic_science_jungle.lua
mheguy: Yes, but you're in luck. For the map I finished yesterday I wrote a little module to make this exact situation really easy. :D

map_gen/maps/toxic_science_jungle.lua
mheguy: GitHub just shows a bit of context with comments. It was meant only for the bottom line: table.insert...

theorangeangle: ok, did that and it failed 😞

map_gen/maps/toxic_science_jungle.lua
```diff
@@ -252,8 +254,5 @@ local function on_init()
     item_market_1.destructible = false
     Retailer.add_market('items', item_market_1)
 end
-Event.on_init(on_init)
```
mheguy: You need to restore this line

linaori:
In case you want to use the default market without flame weapons, you can do something like this: https://github.com/Refactorio/RedMew/blob/5cf640eecdd02746ac78f676df08db144392ae50/map_gen/maps/toxic_danger_ore_jungle.lua#L41-L46 (might have to add conditions).

This ensures that you don't have to keep a synced list of market items.

theorangeangle:
Does that work the same for adding items as well?

linaori:
You can add/remove items. If you want to add them in specific indices, you have to use `table.insert`

mheguy:
I'm making the following changes before merging:
I'm creating a local `config` for `global.config`
I'm using redmew_surface to ensure enemy expansion is enabled
Lastly I'm moving your on_init command out of the code that is defining the shape of the map

So nothing was broken, no behaviour is changed. Just minor optimizations and structural cleanup.

mheguy:
The fail on travis was due to the PR being merged. Post-merge build: https://travis-ci.org/Refactorio/RedMew/builds/495856156

PR #777, opened 2019-02-20
Find and replace 'in ipairs(' with 'in pairs('

mheguy:


Valansch:
So many ipairs. Any non-factorio-lua dev will think you are crazy :D

plague006:
lol is true

grilledham:
This makes it harder to reason about if the table is an array or not. And the difference in performance between `pairs` and `ipairs` is so small that on some hardware `ipairs` is faster.

Now if you want to turn `ipairs` into `for i = 1, #tbl do local value = tbl[i] end` I can get behind that.

mheguy:
The performance of `ipairs()` and actually a fair bit faster than `pairs()` on my PC..

PR #781, opened 2019-02-20
Switch to functions in table, adjust map_loader accordingly

mheguy:
90% of the credit to @grilledham since he basically told me what to do.

Cases tested:

default with modules
default without modules
diggy with modules
diggy without modules
beach with modules
beach without modules
no require in map_selection (runs as if default is loaded since require 'map_selection' returns true)
no map_selection.lua -- ugly error message as always

### DIFF
--- a/map_gen/shared/map_loader.lua
+++ b/map_gen/shared/map_loader.lua
@@ -3,20 +3,45 @@ local b = require 'map_gen.shared.builders'
 local RS = require 'map_gen.shared.redmew_surface'
 local config = global.config.map_generation
 
+local line = '-----------------------------\n'
+local shape_type = type(shape)
+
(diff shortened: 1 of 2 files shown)


map_gen/shared/map_loader.lua
```diff
@@ -3,27 +3,52 @@ local b = require 'map_gen.shared.builders'
...
 
     for _, m in ipairs(config.terrain_modules) do
         shape = b.overlay_tile_land(shape, m)
     end
 end
-
-if type(shape) == 'function' then
+Debug.print(shape)
+Debug.print(shape_type)
+--- If shape it a function, initialize the generator
```
grilledham: typo on 'it'

grilledham (approved):

PR #784, opened 2019-02-21
Fix/walkabout

### LINKED ISSUE #782: Kill command causes a desync
Came about while testing something else

mheguy:
Fixes #782


grilledham (approved):
Another walkabout fix to add to the tally 😄

mheguy:
Really though, how many is that?

mheguy:
And does this count as 1 or 2? because both of them totally break functionality. :D

PR #787, opened 2019-02-21
[Map] Rock solid

mheguy:
The theme of the map is no underground belts or pipes.
For balance, logistics systems research is disabled.

It's a treeless world of refined concrete underpinned by "obsidian" (dark-lab). Rail and wood are provided at a low cost in the market.

To add extra challenge to the map, cliff generation is turned up.

My hope is unruly spaghetti with splitters essentially playing the role of underground belts. Pipes will be a sweet disaster.

### DIFF
--- a/map_gen/maps/solid_rock.lua
+++ b/map_gen/maps/solid_rock.lua
@@ -23,7 +23,7 @@ ScenarioInfo.add_map_extra_info([[
 - The beasts on this planet are savage and unrelenting.
 ]])
 
-RS.set_map_gen_settings({MGSP.enemy_very_high, MGSP.tree_none, MGSP.cliff_very_high})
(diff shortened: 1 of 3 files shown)


linaori:
assemblers will be the new underground pipes

grilledham:
https://github.com/Refactorio/RedMew/blob/835910c385ec0fba72572eb52ec3150380486ebf/utils/recipe_locker.lua#L32
The recipe unlocker registers the same event twice.

The code looks good, haven't tested it though. 

The concept makes me scream in my sleep.

mheguy:
I submitted the PR and 3 mins later force-pushed the fix for the rebase error (the duplicate on_init function). Still too slow before both of you got to it and saw it. Fucking beasts. :D

mheguy:
RB + FP

mheguy:
This seems to run fine in 17 as both a concept and technically.

PR #788, opened 2019-02-21
Add event toggles to debugger

mheguy:
This adds a page to the Debugger 3001 called "Events".
In it you can toggle any of the game's baked-in events. Toggling the events will trigger prints and log entries detailing the event details/package. From LuaObjects it will also try to display the `name` of the object.

image

In addition to that, it will keep a table in Global with the last 10 event packages of each event type.
image

In that event table, the `name` listed for LuaObjects is sometimes a little.. "weird" and seems to contain more than just the name. As this is a debug tool and still fulfills its purpose: I can personally live with it.

Tested in SP and MP with great success.

### DIFF
--- a/features/gui/debug/event_view.lua
+++ b/features/gui/debug/event_view.lua
@@ -2,6 +2,7 @@ local Event = require 'utils.event'
 local Global = require 'utils.global'
 local table = require 'utils.table'
 local Gui = require 'utils.gui'
+local Model = require 'features.gui.debug.model'
 
 local format = string.format
 local insert = table.insert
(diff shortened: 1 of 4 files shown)


grilledham:
> In that event table, the `name` listed for LuaObjects is sometimes a little.. "weird" and seems to contain more than just the name.

What do you mean?

mheguy:
So I realized 2 things:
1) it wasn't coming from what I had written, it was because I was looking at Global.
2) that it's intentional to help the dev figure out which element it is.

`element = "{LuaGuiElement, name = '84 - features/gui/tasklist:line:61'}",`

I had not seen listings of gui elements before in `Global`.

grilledham (approved):
Given this is just for debug mode, it's pretty easy to approve this.
If we discover that any changes are needed / wanted at a later date we can make a new PR.

PR #790, opened 2019-02-22
Added logic for toggling minable robots

SimonFlapse:
I couldn't find anything in the upcoming feature section of the wiki that robots shouldn't be minable in 0.17

This adds a check everything a player selects a construction robot that:
- Checks is the construction robot is owned by a specific player
  - If not: Sets the construction robot to minable = false
  - If: Sets the construction robot to minable = true

This will practically prevent other players from intentional, or unintentional, mine a players personal construction robots.

You might be able to mine a robot if you select it first and the owner selects it afterwards, and you do not remove your selection. I think this is an edge case that's not worth trying to fix.


features/save_bots.lua
```diff
@@ -0,0 +1,30 @@
+local Event = require 'utils.event'
+local Game = require 'utils.game'
+
+local function preserve_bot(event)
+    local player = Game.get_player_by_index(event.player_index)
```
Valansch: Move this line to line 19 please. Preserve_bot will be called alot.

Valansch (dismissed):
Great idea!
One minor thing.

features/save_bots.lua
SimonFlapse: Yeah I know, but I can't :( The event only passes on the previous selected item, not the new one. So I need the player

Valansch: Oh i missed line 6. Thats a bummer.

features/save_bots.lua
```diff
@@ -0,0 +1,30 @@
...
+    local entity = player.selected
+
+    if entity == nil or not entity.valid then
+        return
+    end
+
+    if entity.name == 'construction-robot' then
+        local logistic_cell = entity.logistic_network.cells[1]
```
grilledham: Can you use <https://lua-api.factorio.com/latest/LuaEntity.html#LuaEntity.logistic_cell> instead?
Also do we need to check if cell is nil, what happens if you select a robot and there are no logistic_cell on the map?

SimonFlapse: I tried initially but didn't seem to work very well. And the player will always only have one cell. My local changes have a check for a nil cell

SimonFlapse: image

mheguy:
Because of the small size and scope, this would be a good fit for `redmew_qol`.

SimonFlapse:
Is it a problem to have it isolated? I could understand if you wanted to reuse the same event to reduce overhead.

mheguy:
Not a problem to have it isolated, nope. Was just a suggestion since it's a good fit.

SimonFlapse:
Updated to 0.17 and moved to Redmew-QoL.lua

Tested on S9.

mheguy (changes requested):
You have two copies of the module in this PR.

features/redmew_qol.lua
```diff
@@ -139,6 +139,45 @@ local loader_check_token =
     end
 )
 
+--- Sets construction robots that are not part of a roboport to unminabe
+-- if the player selecting them are not the owner of them.
+local function preserve_bot(event)
+    local player = Game.get_player_by_index(event.player_index)
+    local entity = player.selected
+
+    if entity == nil or not entity.valid then
```
plague006: As a style choice, I'm wondering why you use `== nil` and `not` both?

mheguy (approved):

PR #795, opened 2019-02-23
Added improved hydra scaling

SimonFlapse:
**Change log**
Added min, max and trigger to config
Added online_player_scale to config

Made hail hydra chances a variable of: 
- number of online players (toggle)
- evolution based on a custom formula that has a slow scaling until evolution 0.8

Current config is in need of balancing, what is shown is just an example to illustrate the changes.

**What is this?**
These changes would make hail hydra better suit all kinds of maps with increased customization.

**What is new?**
We can now customize:
- Hydras minimum chance
- Hydras maximum chance
- Toggle whether player count should affect these chances
- Amount of players needed for configured chances to be true.
  - Above this number further increase the chances
  - Below this number decrease the chances
  - Each player increases or decreases the chance with 1%

**What have been removed**
Evolution_scale didn't do much towards customizing the spawn chances. The equivalent in these changes are:
`{min = 0.2, max = min + evolution_scale}`

**Technical changes**
The formula for calculating the chance is:
`(0.00003*(E*100)^3 + 0.004 * (E*100)^2 + 0.3 * (E*100))*0.01` where `E = force.evolution_factor`

Approx values shown in evolution = (percentage of max):
| 0.25 evolution = 10% | 0.50 evolution = 29% | 0.60 evolution = 45% | 0.75 evolution = 58% |
| 0.80 evolution = 65% | 0.90 evolution = 81% | 1.00 evolution = 100% |

**Yet to do**
- [x] Create public functions to change these settings runtime
- [x] Implement compatibility with the old config using the conversion in `What have been removed`
- [x] Balance the default configuration
- [x] Ensure current maps aren't broken from these changes
- [x] Set max to -1 in config if it's unused
- [x] Make overriding the online_player_scale for a single hydra spawn possible

### DIFF
--- a/config.lua
+++ b/config.lua
@@ -217,18 +217,18 @@ global.config = {
         -- At evolution 0.60 it would spawn with a chance of min + max * (percentage of max) = 1.1
         hydras = {
             -- spitters
-            ['small-spitter'] = {['small-worm-turret'] = {min = 0.2, max = 0}},
-            ['medium-spitter'] = {['medium-worm-turret'] = {min = 0.2, max = 0}},
-            ['big-spitter'] = {['big-worm-turret'] = {min = 0.2, max = 0}},
-            ['behemoth-spitter'] = {['big-worm-turret'] = {min = 0.4, max = 0}},
+            ['small-spitter'] = {['small-worm-turret'] = {min = 0.2, max = -1}},
+            ['medium-spitter'] = {['medium-worm-turret'] = {min = 0.2, max = -1}},
+            ['big-spitter'] = {['big-worm-turret'] = {min = 0.2, max = -1}},
+            ['behemoth-spitter'] = {['big-worm-turret'] = {min = 0.4, max = -1}},
             -- biters
             ['medium-biter'] = {['small-biter'] = {min = 1.2, max = 2}},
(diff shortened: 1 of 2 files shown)


linaori:
What's `max = 0`?

SimonFlapse:
> What's `max = 0`?

It's if you don't want the chance to change. I just didn't want to remove `max = ?` for clarity
`max <= min` <=> `max = min`

linaori:
That sounds odd, perhaps `-1` would be a good alternative?

SimonFlapse:
We could do -1 in the default settings, but I might actually set a maximum setting.
It doesn't matter if it's 0, -1 or anything less than min.

linaori:
> It doesn't matter if it's 0, -1 or anything less than min.

It's the way it's treated, when I see max being lower than min, my brain will wtf for a moment, where as -1 indicates that it is disabled (indexof being a prime example). Additionally could also use "false" instead to turn it off, or a different key.

mheguy:
As this is still WIP, going to tag this as post-0.17

SimonFlapse:
What a nice surprise. No RB for me...

SimonFlapse:
Has been used in Diggy launched 2019-03-01 and Quadrants 2019-03-02

linaori (approved):

PR #797, opened 2019-02-23
fix player_list sort by rank

grilledham:
Error in logs
`features/gui/player_list.lua:550: invalid order function for sorting`


grilledham:
Merged so I can test on crashsite

PR #802, opened 2019-02-23
Customizable tag message

Quadrum1:
Users can create their own messages for player tags

These messages appear once a player joins or leaves a team

**Still a bug: Can't clear tag once assigned! .**

### DIFF
--- a/features/gui/tag_group.lua
+++ b/features/gui/tag_group.lua
@@ -32,6 +32,9 @@ end
 
 local function change_player_tag(player, tag_name, silent)
     local old_tag = player.tag
+	
+	local tag_data = tag_groups[tag_name]
+	
     if tag_name == '' and old_tag == '' then
         return false
     end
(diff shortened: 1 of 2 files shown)


SimonFlapse:
You can't clear tags because the player.tag (old_tag) value has brackets in it.
eg. `[Cat]` instead of the expected `Cat`
You cannot change this, but you could do:
```lua
old_tag = string.sub(old_tag, 2, string.len(old_tag)-1)
local tag_data = tag_groups[old_tag]
```
on line 56, and move the other tag_data back to line 66

Quadrum1:
Just noticed that it IS clearing the tag, just not telling it in chat or removing one from the tag window

mheguy:
Tagging this with 0.17 because we're branching at this point to work on 0.17 integration and won't be merging non-fixes until after that is done. Feel free to keep working on this though and we'll absolutely have a look when things have stabilized. :)

theorangeangle:
@Quadrum1 still interested in this? Now that things have stabilized a bit we might be able to look at it.

grilledham:
#958 supersede this. Sorry I didn't spot this PR when I made #958

PR #804, opened 2019-02-24
Improved performance_scale

SimonFlapse:
It now keeps track of applied modifiers and doesn't override modifiers given from third parties.

It now works with diggy


mheguy:
I'm assuming you tested this and got the results you expected. 17 doesn't seem to have affected anything related to this.

PR #805, opened 2019-02-24
Take offline and invalid players for probation add/remove as well as regular-remove

mheguy:


mheguy:
Should be good to go in 17

SimonFlapse (approved):
LGTM

PR #806, opened 2019-02-24
Add scenario_data_manipulation

mheguy:
Adds a number of debug-only commands and functions for manipulating scenario data.
`dataset-copy` takes a dataset and a destination
`dataset-move` takes a dataset and a destination
`dataset-delete` takes a dataset
All 3 do exactly what you would expect.

`dataset-transform` takes a dataset and a destination as arguments and applies the function located in `global.transform_function` to the entries.

The transformation function is given the table of dataset entries as an argument and a table of entries is expected as a return. Those entries will then be written to the destination (which must not be the same as the original data set). The reason for unique source and destination is strictly paternalistic: because if you make a mistake you lose all your original data. You can always `dataset-move` the output from the transform.

Lastly `dataset-transform-test` runs the entries through the transform function and provides the output in the log as well as `global.transform_results` (and does not copy the results to a dataset).

### DIFF
--- a/control.lua
+++ b/control.lua
@@ -125,6 +125,10 @@ if config.popup.enabled then
     require 'features.gui.popup'
 end
 
+-- Debug-only modules
+if _DEBUG then
+    require 'features.scenario_data_manipulation'
+end
 -- Needs to be at bottom so tokens are registered last.
 if _DUMP_ENV then
     require 'utils.dump_env'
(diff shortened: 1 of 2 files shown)


mheguy:
Closing + reopening to kick travis

mheguy:
Merging after consulting with grilled on discord

grilledham (approved):

PR #807, opened 2019-02-24
Change from using the argument as the name to polling the LuaPlayer for the name

mheguy:
Since commands generally use `game.players[arg.whatever]` to get the `LuaPlayer` we shouldn't rely on the argument being a name as it can just as easily be their index. (There are probably plenty more cases of this)


mheguy:
Going to turn this into a much bigger overhaul of command to take localisedstrings

PR #812, opened 2019-02-24
[17][Enhancement] Drop unnecessary tables.

mheguy:
Changed LuaSurface::regenerate_entity()/regenerate_decorative() to treat "nil" as "all" for the autoplace list.


mheguy:
Tested by manually running commands to regen various chunks ex. `/c game.player.surface.regenerate_decorative(nil, {{0,1}})`

PR #814, opened 2019-02-25
[17] Add error_logging

mheguy:
Writes trapped errors to an external file so that we can find errors more easily (the logs are unruly).

Hooks to pcalls that trap standard errors that would crash our maps: server, command, events, and task.

Has a sort of hack in server.lua to obtain the server's time (requiring server would create a circular dependency).

### DIFF
--- a/features/server.lua
+++ b/features/server.lua
@@ -6,6 +6,7 @@ local Event = require 'utils.event'
 local Game = require 'utils.game'
 local Timestamp = require 'utils.timestamp'
 local Print = require('utils.print_override')
+local ErrorLogging = require 'utils.error_logging'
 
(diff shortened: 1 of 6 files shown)


SimonFlapse:
Sounds nice to get a clean error log. Are we supposed to implement this in other modules? Or any criteria for inclusion?

mheguy:
No other standard modules trap their own errors, but basically if you did want to use this module you use it right alongside your call to `log` to print the error message.

mheguy:
RB + FP

mheguy:
Got a thumbs up from grilled via discord, merging so we can start getting some easy-to-use error logs.

PR #826, opened 2019-02-27
Fix walking_state in walkabout

mheguy:
Maybe it's time to get rid of walkabout and tempban. I'm not sure what purpose they serve.


SimonFlapse:
I've never understood walkabout to be honest. I'm alright with removing it, since I don't know the use case for it. It just seems like wasted time to fix

PR #827, opened 2019-02-27
Fix map gen settings

mheguy:


SimonFlapse:
I can't say much to this. But it looks good to me, I assume you have successfully tested the treat_missing_as_default = false additions. Such was grass, desert and water only.

mheguy:
Water only aka waterworld does not work. I've put in a modding interface request to restore that functionality, but we will have to see how that goes.
As it stands, tetris is the only affected map and it is mentioned in one of the issues.

resources/map_gen_settings.lua
```diff
@@ -255,76 +255,114 @@ return {
...
-            grass = {
-                frequency = 'normal',
-                richness = 'normal',
-                size = 'none'
-            },
-            sand = {
-                frequency = 'normal',
-                richness = 'normal',
-                size = 'none'
+    enable_water = {
```
SimonFlapse, line 328: What is the usage of `enable_water` vs `waterworld` https://github.com/Refactorio/RedMew/blob/21a3b9cea724d4b18f07ff05d995aba4b9e42e99/resources/map_gen_settings.lua#L342

SimonFlapse (approved):
LGTM I don't know if we might need to tweak the preset settings

mheguy:
`waterworld` is/was only water tiles, `enable_water` enables water on, for example, `grass_only` maps.

PR #828, opened 2019-02-27
Change how rotten_apples checks for military techs

mheguy:


SimonFlapse:
It only looks for certain types of military sciences now, not all of them? You could check if a science is using military science packs. and `military`, `military-2` and `stone-walls`.

mheguy:
Yes it only takes techs that give damage boosts. Honestly the whole map might need a rethink since, even without my changes, there are significantly less techs in the "military" category.

@Jayefuu do you have the time to rebalance/figure out a new balance?

mheguy:
Nobody seems to want to fix this in a meaningful way right now so rather than break this in a different way I'll just leave it broken. :D

mheguy:
Or maybe I'll just derpily hit "merge" and have to revert it.

PR #830, opened 2019-02-27
Retaining gui size for popups

SimonFlapse:
image
Instead of:
image

The highlight on the first popup is because of gui debug mode.


grilledham:
Can we move the OK button back to the middle?

SimonFlapse:
Oh yeah sorry, I missed that

SimonFlapse:
image
Fixed close button

grilledham (approved):
This looks good 👍

PR #831, opened 2019-02-27
Fixing info GUI

SimonFlapse:
I can't seem to fix the rules tab and the scroll-bars.

Changing screen resolution and/or changing the gui scale while the gui is open breaks the style. It's fixed by redrawing it.


mheguy:
Rules is still off-center. Map info is displaying properly. Word wrap is disabled everywhere because it does not appear to be working as intended.

PR #832, opened 2019-02-27
Fixing poll gui

SimonFlapse:
I've fixed the styling for the poll GUI
It still desyncs on 0.17.2

It fails heavy mode when it is being closed (After a poll has been created and at least one vote casted). No matter the method.

### DIFF
--- a/features/gui/poll.lua
+++ b/features/gui/poll.lua
@@ -87,8 +87,9 @@ end
 
 local function apply_button_style(button)
     local button_style = button.style
-    button_style.font = 'default-bold'
+    button_style.font = 'default-semibold'
     button_style.height = 26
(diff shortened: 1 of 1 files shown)


mheguy:
This removes the desync risk at the expense of coloring the buttons. I don't understand the module well enough to know the *why* of the desync, just that commenting the code that changes color removes the warnings from heavy mode.

SimonFlapse:
From discord:
>I just tested the poll again after plagues fixes, and it doesn't seem to desync anymore. Do we believe this desync is intentional? Or is it a bug with 0.17.*?

Grilledham suggested we try to create a simple gui to test it.

mheguy:
That would be worth following up on. But for now we need this fix, so I'm merging.

PR #833, opened 2019-02-28
Remove all-tech command

mheguy:
Deprecated by `/cheat` vanilla command which enables cheat mode and grants all research.

### DIFF
--- a/features/admin_commands.lua
+++ b/features/admin_commands.lua
@@ -71,12 +71,6 @@ local function toggle_cheat_mode(_, player)
     Game.player_print('Cheat mode set to ' .. tostring(player.cheat_mode))
 end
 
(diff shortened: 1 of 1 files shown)


SimonFlapse:
So `/cheat` is our current `/hax` + `/all-tech` ? What if we don't want to trigger both of them with `/cheat`?

mheguy:
If you want hax, use hax. If you want all-tech without hax, use cheat and then hax. all-tech was debug-only anyways and there are not many use-cases for being in debug and not wanting hax to justify keeping all-tech around.

SimonFlapse:
Oh okay, we are keeping hax

SimonFlapse (approved):

PR #836, opened 2019-02-28
Update luacheckrc to 0.17

mheguy:


mheguy:
As this is doesn't touch code and already proved itself functional by catching `commands.add_command` I'm merging this.

PR #837, opened 2019-03-01
Score GUI - Converted to rich text

SimonFlapse:
- Reducing the footprint

Tooltip on mouse over:
image

I'm not sure how it'll look when you reach multi-digit numbers

### DIFF
--- a/features/gui/score.lua
+++ b/features/gui/score.lua
@@ -10,6 +10,17 @@ local concat = table.concat
 local main_frame_name = Gui.uid_name()
 local main_button_name = Gui.uid_name()
 
+local descriptions = {
+    {disc = 'Satellites launched', icon = '[img=item.satellite]'},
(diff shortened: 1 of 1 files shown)


grilledham:
This looks good 👍 

I would like to know how it looks when the numbers are filled out, does it become too long?

Also is there an advantage to using rich text over sprites. If we used sprites we could avoid the concat.

SimonFlapse:
Performance wise I don't know. But I think it's easier to implement it using the new image tag since it's just a string. Nothing is stopping us from just doing sprites

SimonFlapse:
image
UI Scaling handles this well.

mheguy:

mheguy:
Taking a second look at it, the hand icon is pretty vague. Would hand + assembler be more clear? Similarly the robot makes it seem like that's a bot count.

PR #838, opened 2019-03-01
Tag - Converted to rich text 

SimonFlapse:
Now prevents creating a tag without an icon.

image
vs
image
 
On map:
image

On player:
image

Issues:
Discord chat:
image
If these changes were to make it, we'd need to remove the tag from discord

### DIFF
--- a/features/gui/tag_group.lua
+++ b/features/gui/tag_group.lua
@@ -22,7 +22,7 @@ Global.register(
 
 local function notify_players(message)
     local players = game.connected_players
-    for i=1, #players do
+    for i = 1, #players do
         local p = players[i]
         if p.valid and not no_notify_players[p.index] then
             p.print(message)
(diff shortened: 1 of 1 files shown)


SimonFlapse:
Might conflict with #802

mheguy:
I think including images in the names would be cool, but I think that in replacing the text with just an image the meaning becomes much less clear. Is the biter tag for someone killing biters?

SimonFlapse:
The biter tag is dog, fish is cat. That's the sprite in the tag gui. Troopers are tanks

mheguy:
It wasn't clear to me that the listing still contained the actual names:
image

I think you should hold off until there is some support on the backend as it makes reading discord and the log nightmarish.

SimonFlapse:
True :D I added it to diggy because I wanted to play test all the changes. But yeah, no merging before the backend can handle it

PR #839, opened 2019-03-01
New GUI to assist in using the new rich text feature

SimonFlapse:
Opening this gui with /rich-text allows you to select entities, tiles, items, virtual-signals or recipes and getting a image to use in text.

Doesn't include various other rich text features like fonts, colors. Images for technology or other sprites such as utility

image

### DIFF
--- a/features/gui/rich_text.lua
+++ b/features/gui/rich_text.lua
@@ -1,6 +1,5 @@
 local Event = require 'utils.event'
 local Gui = require 'utils.gui'
-local Game = require 'utils.game'
 local Command = require 'utils.command'
 
 local main_button_name = Gui.uid_name()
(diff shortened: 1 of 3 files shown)


features/gui/rich_text.lua
```diff
@@ -0,0 +1,195 @@
...
+local function rich_text_command(_, player)
+    local event = {player = player}
+    draw_rich_text(event)
+end
+
+Command.add(
+    'rich-text',
+    {
+        description = 'Opens rich text gui',
+        capture_excess_arguments = false,
```
mheguy, line 191: Just as an aside: capture excess=false is the default so this is redundant. Same with allowed by server.

mheguy:

PR #840, opened 2019-03-01
[Diggy] Updated experience gui

SimonFlapse:
image


SimonFlapse:
I've added some other diggy fixes to this PR

SimonFlapse:
Based on bug reported on discord:
https://discordapp.com/channels/432567222481846283/510210668004245504/553030282031136768

I've added a fix for diggy cheat_mode

Diggy now overrides the cheats table from config.lua when _CHEATS = true and _DEBUG = not true

mheguy:
If you're happy with this, merge at your convenience.

PR #846, opened 2019-03-02
New sprite for info gui

SimonFlapse:


mheguy:
So pretty now! Thank you. :D

PR #848, opened 2019-03-02
Added German

Valansch:
@plague006 show me your google translate skills.


mheguy:
Aw Val your timing sucks. :(

mheguy:
Added to #849

PR #849, opened 2019-03-02
The big localization push

### LINKED ISSUE #606: Localization Support
I was trying to fork and translate the strings at `/locale/en`, but the localization file seems to be not the only string the scenario has.

It would be wonderful if the strings are all in `/locale/*`

Or, at least there could be rule that each `.lua` file should have string on topmost part of the file.

### LINKED ISSUE #801: Fix search_command
It's disabled due to the half-in, half-out implementation of locale at the moment and I wanted to open this issue to make sure it doesn't get forgotten.

mheguy:
Updated and totally rewritten:

This converts `command` to use locale strings as well as all command descriptions. This deprecates the current iteration of command search so it is removed.

What remains to localize (but will not be added to this PR by me): GUI (except info), maps, some features and all map_gen shared.

Added some de locale (courtesy Valansch).

Added an extra script for travis to run on nightly builds which will log the un-translated keys.

Overhauled meltdown's commands.

At multiple points I have wanted small bits of lua as easy-to-run commands. I never added them because there was no good file for it. `features.debug_tools` creates the start of that. `/editor` has made a lot of what I wanted to write obsolete, which is actually great.

Removed several modules and maps:
Venus, nightfall, terraforming
I they require too much maintenance for a map that I would prefer to just rewrite some day.
rusky_pvp
Because we will never use it.
silly_player_names
Because it is very broken and even dangerous to run, one day I would like rebuild something like it at some point.
walkabout, tempban
Require too many bug fixes and exceptions while not providing enough benefit. At some point we will rewrite antigriefing and penalties as a whole, at that point I would like to re-introduce some temporary/timed punishments.

Closes #801
Mostly resolves #606


grilledham:
We could get rid of rusky pvp, I don't think we ever used the map and I don't see us using it in the future.

locale/en/rusky_pvp.cfg
```diff
@@ -1,6 +1,15 @@
+# This file is just an aggregate of the crap rusky_pvp wanted.
+
+[rusky-announcements]
+msg-intro1=Welcome to our Server. You can join our Discord at: discord.gg/dtXMQq6
```
Valansch: This invite is invalid. Just use redmew.com/discord

mheguy: [No Russian](https://en.wikipedia.org/wiki/No_Russian).

mheguy:
I'm removing venus and its component mods for the time being. I'll rebuild them someday but atm they need constant updates and maintenance to keep up to date and I hate seeing them all the time. I've marked down refs if I ever need to consult the source.

Valansch:
Keeping the cfg files in sync is going to be a challenge. While people add /en/ strings the other languages will lag behind. To help to see which keys are missing in /locale/xx/*.cfg you can use this bash command:

```bash
cat $1 | sed 's/=.*//' > diff_file.tmp; cat $2 | sed 's/=.*//' | git diff --no-index -- diff_file.tmp -; rm diff_file.tmp
```
with $1 and $2 being the files to diff.
The command **ignores the translation** and only compares the keys.

To make usage easier you can add a function to your ~./bashrc (Should work for gitbash aswell!) like so:
```bash
localediff() {
    cat $1 | sed 's/=.*//' > diff_file.tmp; cat $2 | sed 's/=.*//' | git diff --no-index -- diff_file.tmp -; rm diff_file.tmp
}
```

Example usage:
localediff ./locale/en/redmew.cfg ./locale/de/redmew.cfg

mheguy:
Giving `silly_player_names` the same treatment

linaori:
@Valansch can we add those to a bin directory in the repository? They are kind of an all-purpose utility used to make our lives easier 😄

Valansch:
@linaori 
@plague006 where talking about adding the script to travis releases. Idk where i would put it in the repo tho.

linaori:
I'd suggest a `/bin` directory, or anything related to utility scripts.

mheguy:
https://github.com/Refactorio/backend_scripts

Honestly even if they're not technically backend, I think it's a good place to store scripts in general.

.travis/check_locale.sh
```diff
@@ -1,7 +1,8 @@
 #!/bin/sh
 
-echo "Chaning working directory to Refactorio/RedMew/locale"
-cd Refactorio/RedMew/locale
+cd ~/build/Refactorio/RedMew || exit
+echo "Changing working directory back to: "
```
Valansch: This should be:
```
cd ~/build/Refactorio/RedMew/locale/ || exit
echo "Changing working directory to: "
```

mheguy:
Command descriptions done. Next round will be hitting modules that print.

mheguy:
I consider this feature-complete in the sense that I don't want to work on more localization at this point.

linaori (approved):

PR #851, opened 2019-03-03
Add generate_event_name to event module

mheguy:
This allows us to add custom events to the debugger while being minimally imposing on devs.


mheguy:
Added bug fix where event_view wasn't checking that changed_state events were specific to its checkboxes. :D

grilledham (approved):

PR #852, opened 2019-03-03
Replaced settings commands with a GUI

linaori:
While waiting for the data from the server:
image

When loaded:
image

image

image

With some dummy fields and error message:
image

Messages send to and from the server:
```
[DATA-GET]31 {data_set:"player_settings",key:"plague006"}
[DATA-SET]{data_set:"player_settings",key:"plague006",value:'{["toast-volume"]=0}'}
```

### DIFF
--- a/config.lua
+++ b/config.lua
@@ -367,6 +367,11 @@ global.config = {
         enabled = true,
         use_remote_server = true,
     },
+    -- enables the redmew settings GUI
+    redmew_settings = {
+        enabled = true,
+        use_remote_server = true,
+    }
 }
 
 return global.config
--- a/config.lua
+++ b/config.lua
@@ -369,8 +369,7 @@ global.config = {
     },
     -- enables the redmew settings GUI
     redmew_settings = {
-        enabled = true,
-        use_remote_server = true,
+        enabled = true
     }
 }
 
--- a/config.lua
+++ b/config.lua
@@ -361,7 +361,12 @@ global.config = {
     -- allows the saving and automatic loading of quickbars between maps
     player_quick_bars = {
         enabled = true
-    }
+    },
+    -- enables the redmew settings GUI
+    redmew_settings = {
+        enabled = true,
+        use_remote_server = true,
+    },
 }
 
 return global.config
--- a/config.lua
+++ b/config.lua
@@ -364,11 +364,6 @@ global.config = {
     },
     -- enables the redmew settings GUI
     redmew_settings = {
-        enabled = true,
-        use_remote_server = true,
-    },
-    -- enables the redmew settings GUI
-    redmew_settings = {
         enabled = true
     }
 }
--- a/features/gui/redmew_settings.lua
+++ b/features/gui/redmew_settings.lua
@@ -9,7 +9,6 @@ local Settings = require 'utils.redmew_settings'
 local Color = require 'resources.color_presets'
 
 local pairs = pairs
-local table_size = table.size
 
 local main_button_name = Gui.uid_name()
 local save_changes_button_name = Gui.uid_name()
@@ -122,8 +121,8 @@ local function draw_main_frame(center, player)
 
     local scroll_pane = settings_frame.add({type = 'scroll-pane'})
     local scroll_style = scroll_pane.style
+    scroll_style.vertically_squashable = true
     scroll_style.maximal_height = 800
-    scroll_style.minimal_height = 35 * table_size(settings)
     scroll_style.bottom_padding = 5
     scroll_style.left_padding = 5
     scroll_style.right_padding = 5
@@ -135,17 +134,12 @@ local function draw_main_frame(center, player)
     local data = {}
 
     for name, setting in pairs(settings) do
-        local caption = name
-
-        if setting.localisation_key then
-            caption = {setting.localisation_key}
-        end
-
         local label = setting_grid.add({
             type = 'label',
-            caption = caption,
+            caption = setting.localised_string,
         })
         label.style.horizontally_stretchable = true
+        label.style.height = 35
 
         local value = Settings.get(player_index, name)
         local input = create_input_element(setting_grid, setting.type, value)
--- a/features/gui/redmew_settings.lua
+++ b/features/gui/redmew_settings.lua
@@ -30,7 +30,7 @@ Global.register(
 )
 
 local on_player_settings_get = Token.register(function (data)
-    local player = game.players[data.key]
+    local player = game.get_player(data.key)
 
     if not player or not player.valid then
         return
@@ -51,7 +51,7 @@ local on_player_settings_get = Token.register(function (data)
 end)
 
 local function player_created(event)
-    local player = game.players[event.player_index]
+    local player = game.get_player(event.player_index)
     if not player or not player.valid then
         return
     end
@@ -75,7 +75,7 @@ local function player_joined(event)
         return
     end
 
-    local player = game.players[event.player_index]
+    local player = game.get_player(event.player_index)
     if not player or not player.valid then
         return
     end
--- a/features/gui/redmew_settings.lua
+++ b/features/gui/redmew_settings.lua
@@ -60,6 +60,12 @@ local function player_joined(event)
         return
     end
 
+    local main_frame = player.gui.center[main_frame_name]
+
+    if main_frame then
+        Gui.destroy(main_frame)
+    end
+
     Server.try_get_data_timeout('player_settings', player.name, on_player_settings_get, 30)
 end
 
--- a/features/gui/redmew_settings.lua
+++ b/features/gui/redmew_settings.lua
@@ -47,7 +47,7 @@ local on_player_settings_get = Token.register(function (data)
 
     local button = player.gui.top[main_button_name]
     button.enabled = true
-    button.tooltip = {'redmew_settings_gui.menu_item_tooltip'}
+    button.tooltip = {'redmew_settings_gui.tooltip'}
 end)
 
 local function player_created(event)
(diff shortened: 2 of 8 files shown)


features/gui/redmew_settings.lua
```diff
@@ -0,0 +1,241 @@
...
+        caption = {'redmew_settings.frame_title'},
+    })
+
+    local settings_frame_style = settings_frame.style
+    settings_frame_style.width = 400
+
+    local scroll_pane = settings_frame.add({type = 'scroll-pane'})
+    local scroll_style = scroll_pane.style
+    scroll_style.maximal_height = 800
+    scroll_style.minimal_height = 35 * table_size(settings)
```
grilledham: I wonder what happens if `minimal_height` > `maximal_height`?
Another way you can handle this is to set `scroll_style.vertically_squashable = true` 
Then the scroll pane takes up as much room as it needs, up to the max height and if you want more minimal height you can use padding.

features/gui/toast.lua
```diff
@@ -14,7 +14,7 @@ local next = next
 local format = string.format
 
 local toast_volume_name = 'toast-volume'
-Settings.register(toast_volume_name, 'fraction', 1.0)
+Settings.register(toast_volume_name, Settings.types.fraction, 1.0, 'toast.toast_volume_setting_label')
```
grilledham, line 16: Is there a reason why the localisation_key isn't passed already in a table like `{'toast.toast_volume_setting_label'}`?
Then we don't have to rebuild the table later when using the key.

features/gui/redmew_settings.lua
```diff
@@ -0,0 +1,241 @@
...
+    local player_index = player.index
+
+    local data = {}
+
+    for name, setting in pairs(settings) do
+        local caption = name
+
+        if setting.localisation_key then
+            caption = {setting.localisation_key}
+        end
```
grilledham: If we make the change to the localisation_key that is suggested below this code could be changed to
`local caption = setting.localisation_key or name`

features/gui/redmew_settings.lua
```diff
@@ -0,0 +1,241 @@
...
+        })
+        label.style.horizontally_stretchable = true
+
+        local value = Settings.get(player_index, name)
+        local input = create_input_element(setting_grid, setting.type, value)
+
+        data[name] = {
+            label = label,
+            input = input,
+            previous_value = value,
```
grilledham, line 140: What's `previous_value` used for, I don't see it being used anywhere?

grilledham (approved):
This looks really good.

Also this passed testing in multiplayer.

This could use some fail safe mechanism for `Server.try_get_data`, such as a timeout. But until we figure that out I'm happy to merge this as is.

mheguy:
Because namespaces are shared across all locale files and there are now two modules with the name `redmew_settings` I made the locale more specific to each.

config.lua
```diff
@@ -310,6 +310,11 @@ global.config = {
     -- enables a command which allows for an end-game event
     apocalypse = {
         enabled = true
+    },
+    -- enables the redmew settings GUI
+    redmew_settings = {
+        enabled = true,
+        use_remote_server = true,
```
mheguy: Without a comment, this can be interpreted in many ways such as an SP run pinging redmew servers for data.

utils/redmew_settings.lua
```diff
@@ -103,26 +103,46 @@ function Public.register(name, setting_type, default)
...
     return setting
 end
 
+---Validates whether a given value is valid for a given setting.
+---@param name string
+---@param value any
+function Public.validate(name, value)
+    local setting = settings[name]
+    if not setting then
+        return format('Setting "%s" does not exist.', name)
```
plague006: Tooltips will happily take localised strings.
On a related note you're missing doc for the return value.

utils/redmew_settings.lua
```diff
@@ -103,26 +103,46 @@ function Public.register(name, setting_type, default)
...
 
     settings[name] = setting
 
     return setting
 end
 
+---Validates whether a given value is valid for a given setting.
+---@param name string
+---@param value any
+function Public.validate(name, value)
```
mheguy, line 121: Missing doc for the return value. You might also want to consider returning 2 values, similar to `setting.callback` since there's no reliable way to check if the data was validated.

utils/redmew_settings.lua
mheguy, line 124: You can return a LocalisedString rather than a string here.

utils/redmew_settings.lua
```diff
@@ -103,26 +103,46 @@ function Public.register(name, setting_type, default)
...
+
 ---Sets a setting to a specific value for a player.
 ---
 ---In order to get a setting value, it has to be registered via the "register" function.
 ---
 ---@param player_index number
 ---@param name string
----@param value mixed
+---@param value any
 function Public.set(player_index, name, value)
```
plague006: Missing return doc

mheguy, line 143: Missing return doc

utils/redmew_settings.lua
```diff
@@ -87,8 +87,8 @@ Public.types = {fraction = 'fraction', string = 'string', boolean = 'boolean'}
 ---
 ---@param name string
 ---@param setting_type string
----@param default mixed
-function Public.register(name, setting_type, default)
+---@param default any
+function Public.register(name, setting_type, default, localisation_key)
```
mheguy, line 92: Missing return doc

mheguy (dismissed):
When loaded in SP this create an unusable and confusing button. 
image

mheguy:
Setting and command feedback are universally given as either console prints or in-gui confirmation. To that end I would suggest either using either and keeping toasts for gameplay feedback or announcements.

mheguy:
I pushed 7a5e6c1 to detect the server's presence and tested 4 cases: launching scenario SP + uploading to server, launching scenario on server + saving to SP.

In all 4 cases the button and gui worked as expected.

Please feel free to change our outright revert 7a5e6c1 though, it was just a proof of concept for myself.

Of note: when download the MP save to SP the scenario will think the server is present, but that shouldn't be any issue as we have many modules that assume the presence of the server and try to pull + push data all the time.

utils/redmew_settings.lua
linaori: By default it's not added in my IDE because it knows the return type. If you want, I can add it, but I'm not sure if it still tells me the correct structure afterwards

utils/redmew_settings.lua
linaori: I wasn't sure if this variant was needed. In theory this shouldn't even reach the GUI, thus I figured it would be overhead to load this as translation. However, depending on the usage of this function (and `set`), it _could_ be returned as error or normal message, but such usages are currently not available anymore (removed with removal of commands)

utils/redmew_settings.lua
linaori: If there was data (a string) -> error, otherwise nil is returned

config.lua
linaori: A different name + docs should solve it, though ideally I'd remove this option altogether. 

Perhaps @grilledham has an idea of how to solve this as ideally I'd like to have the Server module determine whether or not a server is present and use a local storage instead (created on the fly and not persistent between maps). This way our calls will still function but with a local stub. Ideally I'd not have other users fiddle with this setting at all as it brings added complexity

features/gui/redmew_settings.lua
linaori: It _was_ used but is no longer. I'll remove this. It might be interesting for a "reset", but then you'd want the original defaults rather than previous values.

features/gui/redmew_settings.lua
linaori: I will make the register function turn this into a table :+1:

features/gui/redmew_settings.lua
linaori: You're right. When I don't add a minimal height, it doesn't work as it will summon a scrollbar as it doesn't exactly fit. I'll fiddle around with this later

mheguy:
RB+FP for conflict

utils/redmew_settings.lua
mheguy, line 92: Missing doc for new param (reading the code is seems you want people to send a string instead of a table for locale which should definitely be documented since it's atypical)

utils/redmew_settings.lua
mheguy: This issue remains though: we should not be returning plain text strings when they're meant to be feedback. If you want to use `error` I won't resist the use of a string. But otherwise we should be using localised strings.

utils/redmew_settings.lua
mheguy: Yes, please documenting per the styleguide is strongly encouraged.

linaori:
Fixed the merge conflicts

mheguy:
A summary of which issues will and won't be addressed would be helpful.

linaori:
Hmm I thought everything was addressed 🤔 I'll get back to this in the weekend

mheguy:
Could be, I'm just skimming issues and I noticed a bunch of unresolved.

features/gui/redmew_settings.lua
```diff
@@ -0,0 +1,254 @@
...
+    end
+
+    local main_frame = player.gui.center[main_frame_name]
+
+    if main_frame then
+        Gui.destroy(main_frame)
+    end
+end
+
+local function enable_server()
```
mheguy: I'm running into the same issue as you in #854 -- I can catch the server coming online, but don't have a reliable way to catch the server being offline.

linaori:
Tested with @grilledham, including syncing between multiple running games.

PR #853, opened 2019-03-04
luacheckrc tweaks to script and commands

mheguy:
luacheckrc tweaks to raise warnings when registering events through script, or registering a command through commands


mheguy:
RB+FP for conflicts

PR #854, opened 2019-03-04
Add quick bar saving

### LINKED ISSUE #822: Add save/load Action Bar
Currently Action bars / Toolbars are saved on a per map basis. According to rseding we should be able to scan and set action bars with LUA. 

Ideally a player can set-up their action bar, use the save command, and have that action bar any time they join a Redmew scenario automatically.

mheguy:
Restores quick bars between maps automatically.

Saves or removes quick bar storage via commands.

Only registers the event hook if the server is (or was ever) present.

Closes #822

### DIFF
--- a/config.lua
+++ b/config.lua
@@ -357,6 +357,10 @@ global.config = {
             -- whether to only attack on the first launch
             first_launch_only = true
         }
+    },
+    -- allows the saving and automatic loading of quickbars between maps
+    player_quick_bars = {
+        enabled = true
     }
 }
 
--- a/control.lua
+++ b/control.lua
@@ -92,6 +92,9 @@ end
 if config.biter_attacks then
     require 'map_gen.shared.biter_attacks'
 end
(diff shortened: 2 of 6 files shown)


mheguy:
RB+FP for merge conflicts

mheguy:
RB+FP for merge conflicts

mheguy:
~~Will have to iterate on this to account for missing items/mod items, and the like~~

mheguy:
Updated to account for gaps in the saved bars as well as invalid items.

Items are checked against `game.item_prototypes` entries for validity. In cases of non-valid entries the setting of the bar entry is skipped and a player-printed message is sent.

plague006:
Hmm.. I want to expand this to have multiple save slots.

plague006:
Restructured a bit, cleaned up locale.
The clear bars command is currently set to debug only because I don't know if it's a good idea or not but wanted to push forward with what I had to finish this PR.

PR #857, opened 2019-03-04
Added function to cancel a Server.try_get_data requests

grilledham:
There are functions that let people chose there own cancel logic. 

But I only expect people will use `Server.try_get_data_timeout` which works the same as `Server.try_get_data` expect if the timeout is reached the callback is called with `nil`.

I haven't added functions to cancel `Server.try_get_all_data` but if we want that I can add that later in a future PR.


mheguy:
RB+FP

PR #862, opened 2019-03-06
Added player research scaling

SimonFlapse:
The technology price modifier increases with number of connected players.

Only updates on_research_finished to always show the player the correct cost when picking a research.

The module keeps track of its own modifiers so that others can manipulate this value without conflicting.


SimonFlapse:
Initial reactions in terraforming danger ores has been negative
Consider rebranding as: Science gets cheaper with less players

mheguy:
I haven't played with this, but as a gut reaction I don't like the idea of scaling difficulty based on player count in general. I think it incentivizes players treating each other as cost multipliers rather than just as people.

It also creates a perverse incentive to have *less* players online. And it certainly encourages people to want AFK players kicked.

mheguy:
Would appreciate input from others to determine if this is a good idea.

TWLTriston:
Adding my input -- I do not like this in play testing -- especially in danger ores where there was a starting x20 multiplier, seeing that jump to x40 as the server got more popular discourages progress. Prefer to not see this committed as a regular feature. A great project penguin feature however

SimonFlapse:
> Adding my input -- I do not like this in play testing -- especially in danger ores where there was a starting x20 multiplier, seeing that jump to x40 as the server got more popular discourages progress. Prefer to not see this committed as a regular feature. A great project penguin feature however

For perspective, danger ores was configured to be x50 on the dev-branch. I lowered it to x20 and added research scaling.

I could still see it used for increasing research costs on "regular" maps with x1 (or x0.5) science becoming x2/3/4 with 20/40/80 players online or something like that.

Adding to that, I could add an option to increase the science cost of starter research, since they seem a bit low for multiplayer.

PR #865, opened 2019-03-07
Dynamic landfill

SimonFlapse:
With this, landfill will no longer be the plain, old, boring grass-1.

This counts all terrain tiles around the player in a 22x22 bounding box centered around the player. The tile with most counts replaces the grass-1 from the landfill (excluding water tiles)


SimonFlapse:
0.17.10 broke it

features/redmew_qol.lua
```diff
@@ -179,6 +179,85 @@ local function preserve_bot(event)
...
+-- helper function for landfill_detect, creates a bounding box
+-- from a position and a radius
+local function create_bounding_box(position, radius)
+    return {{position.x - radius, position.y - radius}, {position.x + radius, position.y + radius}}
+end
+
+--- Changes landfill tiles from grass-1 to another tile based on the surroundings
+local function landfill_detect(event)
+    local item = event.item
+    if not item then
```
mheguy: This and the following condition can be combined.
```if not item or item.name ~= 'landfill' then```

File: features/redmew_qol.lua [hunk 2]
```diff
@@ -179,6 +179,85 @@ local function preserve_bot(event)
...
+
+    local player = Game.get_player_by_index(event.player_index)
+    local surface = game.surfaces[event.surface_index]
+
+    local position = player.position
+    local area = create_bounding_box(position, search_radius)
+
+    local tiles = event.tiles
+    local num_tiles = #tiles
+    local max_tiles = 4 * (search_radius * search_radius) - num_tiles
```
mheguy: Since `search_radius` is a constant, you can make another constant for `4 * (search_radius * search_radius)` instead of calculating it each time.

features/redmew_qol.lua
```diff
@@ -179,6 +179,85 @@ local function preserve_bot(event)
...
+    local tile = {count = 0}
+    for i, tile_name in ipairs(check_tiles) do
+        local count =
+            surface.count_tiles_filtered {
+            area = area,
+            name = tile_name
+        }
+        local water = string.find(tile_name, 'water')
```
mheguy: Localizing `find` would be beneficial since this may well run a *lot* while players are paving.

features/redmew_qol.lua
```diff
@@ -179,6 +179,85 @@ local function preserve_bot(event)
...
+local function landfill_detect(event)
+    local item = event.item
+    if not item then
+        return
+    end
+    if item.name ~= 'landfill' then
+        return
+    end
+
+    local player = Game.get_player_by_index(event.player_index)
```
mheguy: `Game.get_player_by_index` is deprecated and we should use `game.get_player`

features/redmew_qol.lua
```diff
@@ -179,6 +179,85 @@ local function preserve_bot(event)
...
+    local surface = game.surfaces[event.surface_index]
+
+    local position = player.position
+    local area = create_bounding_box(position, search_radius)
+
+    local tiles = event.tiles
+    local num_tiles = #tiles
+    local max_tiles = num_max_tiles - num_tiles
+    local tile = {count = 0}
+    for i, tile_name in ipairs(check_tiles) do
```
linaori: Is `ipairs` really necessary here? It looks to me like we can also use a normal for loop (which is faster)

SimonFlapse: I've refactored into a normal loop

SimonFlapse:
The new landfill sprite might make this outdated

theorangeangle:
@SimonFlapse will start testing after thanksgiving, unless there are any objections

theorangeangle:
@SimonFlapse considering how pleasant the vanilla implementation of landfill is, do you want to close this issue?

SimonFlapse:
An update to Factorio has made this irrelevant

PR #883, opened 2019-03-21
Biter Corpse Removal

### LINKED ISSUE #608: Corpse cleanup feature?
From discord:
```
[10:36 AM] Newcott - Chewin' Legos: To remove the dead bodies (good for diggy) 
/sc for _,c in pairs(game.player.surface.find_entities_filtered{type='corpse'}) do c.destroy() end
[12:48 PM] plague006: that is a very heavy command to run though, if you do that on a normal/big map it will likely freeze the server for 5-10 seconds
NEW MESSAGES
[1:57 PM] Linaori: Ideally I'd have some sort of % based GC running
[1:57 PM] Linaori: Every biter that dies, has an X percentage of chance to mark the area for cleanup within an X range after X seconds
```

It would be worth looking into a feature where we cleanup corpses after major engagements with biters. Clearing the corpses can significantly help FPS for those with potato PCs.

adrianvanderwal:
Adds biter corpse removal once threshold is reached

closes #608

### DIFF
--- a/features/biter_corpse_util.lua
+++ b/features/biter_corpse_util.lua
@@ -1,0 +1,66 @@
+local Event = require "utils.event"
+local Global = require 'utils.global'
+
+local biter_utils_conf = global.config.biter_utils
+
+local biter_corpses = {}
+
+Global.register(
+    biter_corpses,
+    function(tbl)
+        biter_corpses = tbl
+    end
+)
+
+-- currently no on_corpse_spawned event, using on_entity_died instead
+
+local function biter_died(event)
+    local dying_force = entity.force
+
+    -- ignore player owned entities
+    if dying_force == "player" then
+        return
+    end
+
+    local entity = event.entity
+    local name = entity.name
+    local corpse_names = entity.prototype.corpses
+    
+    -- put corpse name in global list of biter/enemies
+    for c, corpse_name in pairs(corpse_names) do
+        if not biter_corpses[corpse_name] then
+            biter_corpses.insert(corpse_name)
+        end
+    end
+
+    local position = entity.position
+    local radius = biter_utils_conf.radius
+    local area_to_search = {{position.x - radius, position.y - radius}, {position.x + radius, position.y + radius}}
+    local surface = entity.surface
+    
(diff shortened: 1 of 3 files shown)


mheguy (changes requested):
Please test your code (it will error on the death of any entity).
Please look at Travis CI's feedback (the feedback will indicate the cause of the first error you will run into) by checking the details:
image

grilledham:
For the most part the code looks good. The typos that travis pointed out can be easily fixed. Also you should check the entity is valid. Other handlers may run on the same event and make the entity invalid, unlikely in this case but possible so you have to protect against it. 

I do however have concerns about the potential performance impact of such code. Before this is merged it would be good if someone could run some tests to try and measure the performance impact of this code. Which I'll admit wont be easy to do.

 If performance does turn out to be an issue one idea would to be early return every 99 out of 100 times the function is called. Or we could use one of your suggestions in the issue and keep a count of corpses per chunk and clear the chunk when it reaches a threshold.

Another thought is `surface.find_entities_filtered` probably returns entities from the top left to the bottom right of the search area. That might look odd when clearing, we could instead destroy every other corpse to leave some behind.

features/biter_corpse_util.lua
```diff
@@ -15,52 +15,53 @@ Global.register(
...
+         then
+            -- remove the corpses
+            for k, corpse in pairs(
+                surface.find_entities_filtered {
+                    area = area_to_search,
+                    type = 'corpse',
+                    name = biter_corpses,
+                }
+            ) do
+                if corpse.valid and k % 2 == 1 then
```
mheguy: `corpse` is coming from `find_entities_filtered` which will always return valid results, so no need to check validity here

mheguy (changes requested):
L19: `if entity.valie then` suggests the code is still untested. Please inform us when you have tested this PR.

adrianvanderwal:
I have tested this PR on my local machine - barely slows the game down when used on crash site.
The removal of half the corpses is very sudden - may need some tweaking to find numbers that are not so drastic.

PR #888, opened 2019-03-30
feat(player_list): added internationalization

Ku3mi41:
I want translate GUI to russian language, but locale files does not contain all displaing strings. In this reason I want internationalize all GUI mods, so they have ability display different laguages. I start with  "Player list".
That's my first experience with Lua and Factorio modding in total, so give me advice if I do something wrong. I didn't found another way to concatenate translated strings, except call {string} with params, because of that I made a bit refactoring of sorting symbol displaing logic. Is it acceptable way? Should me continue this work?

p.s.> most of casual russian players not very good in English, so I want help them a bit. anyway it's a pretty nice then interface have your native language ;)


linaori:
A lot of files are not yet using the translation system yet, so it's not possible for everything at this moment

Ku3mi41:
> A lot of files are not yet using the translation system yet, so it's not possible for everything at this moment

That's what I mean when told "Internationalization" - replace hardcoded strings by LocalizedString

mheguy:

PR #889, opened 2019-03-31
Revisions to the danger ores scenario

grilledham:
My vs code wouldn't let me checkout tris' branch as there was already a branch named develop so I made a new branch.

### Changes from tris's version

- Lowered stone ratios
- Removed mining productivity techs
- Increased uranium frequency and reduced size.
- Min pollution to trigger a chunk starts at 300 and goes up by 2 with each chunk unlocked up to a max of 2500


map_gen/maps/terraforming_danger_ores.lua
```diff
@@ -8,6 +8,52 @@ local table = require 'utils.table'
...
+)
+
+ScenarioInfo.set_map_description(
+    [[
+Clear the ore to expand the base,
+focus mining efforts on specific quadrants to ensure
+proper material ratios, expand the map with pollution!
+]]
+)
+ScenarioInfo.set_new_info(
```
mheguy, line 42: This info will only be surfaced when not run on redmew servers. When run on redmew servers, the changelog stored server-side will supercede it.

map_gen/maps/terraforming_danger_ores.lua
```diff
@@ -146,13 +224,41 @@ local function ore(x, y, world)
...
-    local index = table.binary_search(weighted_ores, i)
-    if (index < 0) then
-        index = bit32.bnot(index)
-    end
+    local i
+    local index
+    local resource
+
+    if x > 0 and y > 0 then
+      i = math.random() * total_ores_pos_x_pos_y
```
mheguy, line 232: I get there impression there will be many calls to the various `math`, `bit32`, and `table` functions. Would it be worthwhile to localize them?

mheguy (approved):
Looks good overall.

map_gen/maps/terraforming_danger_ores.lua
grilledham: If I wrote it, I'd also use a lookup table rather than string concatenation for the tile names.

PR #890, opened 2019-03-31
Recursive event_name for Event

Cooldude2606:
Some context: while switching over to use the Redmew utils I found a feature that is very useful but was not yet implemented so I made this pull request.

Changes:
- All functions in Event which accept event_name as a parameter will now accept an array of event names.
- All functions in Event which accept event_name as a parameter will now accept the name of an event `'on_tick'` as well as the define of the event `defines.events.on_tick`.
- When a non number value (or a string that resolves to a number see above) is given as the event_name then an error will be thrown.

Note: No changes have been made to event_core; and changes have only been made to function which accept event_name as a parameter within the Event module.

Test Code:
```lua
-- Test for recursion
Event.add({
        defines.events.on_built_entity,
        defines.events.on_robot_built_entity,
        defines.events.script_raised_built
    },
    function(event)
        game.print(serpent.block(event)) -- prints the content of the event table to console.
    end
)
-- Test for string as event_name
Event.add({
        'on_built_entity',
        'on_robot_built_entity',
        'script_raised_built'
    },
    function(event)
        game.print(serpent.block(event)) -- prints the content of the event table to console.
    end
)
```

### DIFF
--- a/utils/game.lua
+++ b/utils/game.lua
@@ -50,7 +50,7 @@ end
 function Game.get_player_from_any(obj)
     local o_type = type(obj)
     local p
-    if type == 'number' then
+    if o_type == 'number' then
         p = Game.get_player_by_index(obj)
     elseif o_type == 'string' then
         p = game.players[obj]


grilledham:
> All functions in Event which accept event_name as a parameter will now accept an array of event names.

There may be some, but off the top of my head I can't think of any case in the Redmew code base where the same handler is used for more than one event. So I'm not convinced accepting an array of event names is useful.


> All functions in Event which accept event_name as a parameter will now accept the name of an event.

Accepting strings for event names seems like a bad idea, wouldn't that break linting? 
Also in the Redmew code base not all events live in define.events so this would encourage a style that doesn't work all the time, making the Event.add function harder to use.

Cooldude2606:
Fair enough if you believe that this will have no use; thank you for reading over it.

PR #894, opened 2019-04-11
added tile_picker feature for better dirt generation in diggy

ChickingWasTaken:
Uses perlin noise for dirt generation so it is nicer to look at


map_gen/maps/diggy/feature/diggy_hole.lua
```diff
@@ -11,6 +11,7 @@ local ScoreTable = require 'map_gen.maps.diggy.score_table'
 local Command = require 'utils.command'
 local CreateParticles = require 'features.create_particles'
 local Ranks = require 'resources.ranks'
+local TilePicker = require 'map_gen.maps.diggy.feature.tile_picker'
```
linaori, line 14: If you disable the feature, what happens?

ChickingWasTaken: It generates terrain like it used to `
    if not enabled then
        return 'dirt-' .. random(1, 7)
    end`

theorangeangle:
@linaori was wondering if i could get a final yes/no on this PR, so that we can clean out some of the old ones!

BlueRock2:
I know this is an old PR but wanted to give some feedback.   I like the option to use perlin noise for the floor tiles.   Just a little sad that the changes create a new tile_picker.lua file to generate the terrain.   Unfortunately, this moves a little further away from the wishlist item (Diggy feedback/todo collection thread #926)  to try and use ore generation system (which is rather robust) and maybe expand it to be able to handle the noise for terrain and rooms (which also use noise) as well.   Granted, the wishlist item came out shortly after this PR was made.

BlueRock2:
Or even better, use when possible tools that are shared by other RedMew maps.  Or add to shared tools if other maps could benefit.   (I still need to learn a lot about what is exactly is available right now)

PR #899, opened 2019-04-25
Map terraforming danger ores april2019

TWLTriston:
Modifications to danger ores scenario, for peer review, not comprehensively play tested.

### DIFF
--- a/map_gen/maps/terraforming_danger_ores.lua
+++ b/map_gen/maps/terraforming_danger_ores.lua
@@ -327,6 +327,40 @@ local function enemy(x, y, world)
     end
 end
 
+local tiles_iron = {
+  [1] = 'grass-1',
+  [2] = 'grass-2',
+  [3] = 'grass-3',
+  [4] = 'grass-4'
+}
+
+local tiles_copper = {
+  [1] = 'red-desert-0',
+  [2] = 'red-desert-1',
+  [3] = 'red-desert-2',
+  [4] = 'red-desert-3'
+}
+
+local tiles_coal  = {
+  [1] = 'dirt-1',
+  [2] = 'dirt-2',
+  [3] = 'dirt-3',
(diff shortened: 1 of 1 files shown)


TWLTriston:
Known issue - biters in unrevealed tiles still spawn post the "win" condition

map_gen/maps/terraforming_danger_ores.lua
```diff
@@ -452,6 +556,59 @@ map = b.fish(map, 0.025)
...
+                'Biters at maximum evolution! Protect the base for an additional 100 rockets to wipe them out forever.'
+            game.print(message)
+        end
+
+        game.forces.enemy.evolution_factor = current_evolution
+        if win_condition_evolution_rocket_maxed > 0 and satellite_count >= (win_condition_evolution_rocket_maxed + 100) then
+            message = 'Congratulations! Biters have been wiped from the map!'
+            game.print(message)
+
+            win_condition_evolution_rocket_maxed = true
```
grilledham: Is this meant to be `win_condition_biters_disabled `? The variable changes from a number to a Boolean which seems odd.

grilledham:
I fixed a problem with the random quadrants. Before if you saved and loaded, the game would crash saying you can't use `math.random` in the load event. If it didn't crash it would have caused a desync anyway.

`win_condition_evolution_rocket_maxed` and `win_condition_biters_disabled` need to be stored in global if you are going to change them at runtime, else the game will desync.
This change still needs to be made.

Otherwise, I don't see anything else that looks problematic.

map_gen/maps/terraforming_danger_ores.lua
TWLTriston: Yes, thank you!

grilledham:
We played this, as far as I can tell it worked, so let's merge.

PR #863, opened 2019-03-07
Quadrants balance update

SimonFlapse:
Tweaked hail hydra to prevent worms from spawning from spitter. This would cause the worms to sometimes spawn outside a turrets range, thus destroying all defences.

Added mapgen settings based on feedback

Added rendering.draw_text for markets to help new comers understand what/where the market is
Included a en, de and da translation of the market name:

Showing danish translation:
image
Says:
English: Market
German: Markt

PR #902, opened 2019-05-02
Fixing for 0.17.35

SimonFlapse:
Fixes based on: https://forums.factorio.com/70188 for 0.17.35 release


SimonFlapse:
I think I've gotten all of them. I have yet to test it. The ones in Quadrant have been fixed in #863 
@plague006 Travis needs to have his defines updated to reflect the renaming of player inventories

grilledham:
I'm going to merge this now so it updates the scenarios on the server. If we find more problems we can make a new PR.

mheguy:
I opened https://github.com/Refactorio/RedMew/issues/904 for changes to luacheckrc based on https://github.com/Refactorio/RedMew/pull/902#issue-275397758, but it seems like grilled got to it? Are there any outstanding changes to be made?

SimonFlapse:
Not that I know of

PR #905, opened 2019-05-06
Updated utils.gui to use mod-gui

Cooldude2606:
Changed top toggle area to use mod-gui by using get_button_flow and also changed the default style of the toggle button to mod_gui.button_style however other style changes are left the same.
Note: the toggle height was changed to 36 rather than 38 to match mod_gui.button_style this may not fit if the other buttons are not using the style.


Cooldude2606:
I needed to move this off my develop branch; see #939

PR #908, opened 2019-05-11
Map/danger bobs ores (WIP)

grilledham:
~Still need to check bob's mod entities for place on ores - bob's miners on ores.~
~Need to update scenario info for the map~
I think this is good now


map_gen/maps/danger_bobs_ores.lua
```diff
@@ -52,8 +52,6 @@ local ores_names = {
     'stone',
     'uranium-ore',
     'bauxite-ore',
-    'cobalt-ore',
-    'gem-ore',
```
grilledham: Still need these, this was to prevent the mod from spawning ores so we have complete control of ore spawn.

PR #912, opened 2019-05-16
Misc fixes

grilledham:


grilledham:
I re enabled the poll focus colors and it seems to be fine.
However the disabled focus colors cause desyncs, see <https://forums.factorio.com/viewtopic.php?f=7&t=70893>

grilledham:
And 0.17.40 fixed the issue above

PR #914, opened 2019-05-20
Rocky Road

theorangeangle:
1. Adds _Rocky Road_: A new map with vanilla generation, but taking advantage of the car_body feature to make things a bit harder. The day/night cycle is also modified, to make solar farms a bit harder than normal. 

2. Fixes a line in car_body so that the redmew surface is used.


map_gen/shared/car_body.lua
```diff
@@ -1,4 +1,5 @@
 local Event = require 'utils.event'
+local RS = require 'map_gen.shared.redmew_surface'
```
grilledham: This line makes no sense, why are you requiring a module you don't use? This what travis is complaining about.

PR #915, opened 2019-05-22
set day/night to bright

ghost:
hope I got that right XD


SimonFlapse:
Thanks for contributing!
For future reference here is what we talked about on discord:
https://discordapp.com/channels/432567222481846283/432570869039104012/580695484138782720

grilledham:
Why did you delete `map_selection.lua.sample `?

ghost:
It should be there still? Unless i did change the name back

ghost:
> Why did you delete `map_selection.lua.sample `?

Its still there just renamed `map-selection.lua`

mheguy:
This will only work if the day/night cycle is also enabled via the config. It also relies on someone to remove the comment tags to get it working.

If you want an optional day/night cycle selection I would propose:

Inserting the requires at the top and adding the custom cycle variable:
```lua
local table = require 'utils.table'
local Config = require 'config'
local bright_cycle = require 'resources.day_night_cycles'.bright

local use_custom_day_night_cycle = false
local sand_width = 512
```

and then adding the functional code inside a conditional after the variables are all declared:

```lua
local perlin_seed_2 = nil

if use_custom_day_night_cycle then
    -- Set the day/night cycle to the "Bright" preset. See resources.day_night_cycles for details.
    Config.day_night = {
        enabled = true,
        use_day_night_cycle = true,
        day_night_cycle = bright_cycle,
        use_fixed_brightness = false
    }
end

RS.set_map_gen_settings(
```

Code untested.

theorangeangle:
@RCTheBlueWolf have you read plague's comment?

PR #916, opened 2019-05-26
Added a tooltip to explain the evolution button

linaori:
During a youtube video I noticed that it was unclear what the number was. In Diggy people thought it was mining speed. 
image

### DIFF
--- a/features/gui/blueprint_helper.lua
+++ b/features/gui/blueprint_helper.lua
@@ -276,7 +276,12 @@ local function player_joined(event)
         return
     end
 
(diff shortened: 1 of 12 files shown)


grilledham (approved):

linaori:
@grilledham perhaps we should add more tooltips for the other buttons as well? I can do that in this PR if needed

grilledham:
Yes, I think adding tooltips to the other buttons would be a good idea.

linaori:
image
image
image
image
image
image
image
image
image

grilledham (approved):

PR #922, opened 2019-05-27
Settings sync for notify: pokes, polls, tasks, tag_groups

grilledham:


features/gui/tag_group.lua
```diff
@@ -665,6 +677,39 @@ Gui.on_custom_close(
 
 Gui.allow_player_to_toggle_top_element_visibility(main_button_name)
 
+Event.add(
+    Settings.events.on_setting_set,
+    function(event)
+        if event.setting_name ~= notify_name then
```
linaori, line 683: This functions looks very similar to the other 2, would it be an idea to put this code in a function that can be shared by all 3?

grilledham: Maybe 😄. There is probably quite a lot of code that could be shared by the guis. But I don't feel like refactoring it atm.

PR #924, opened 2019-05-27
Added a sync message to the settings GUI

linaori:
Not 100% sure if the message is correct, but this way we can show that the settings will be synced between redmew servers. If you have suggestions for the message, post them here.

image


theorangeangle:
> Saving your settings here (on a RedMew-hosted server) will sync automatically to all future maps on RedMew-hosted servers.

Sounds better but might be just me, although the first one is a little wordy.

linaori:
Ideally I'd detect if we were on a redmew server and show a different message, but not sure how reliable that is...

linaori:
What about:

> When playing on a Redmew server, these settings will synchronize to other Redmew servers.

PR #925, opened 2019-05-28
Replaced command for player and chat colors

linaori:
This PR adds a new setting in the GUI for player colors and chat colors. This replaces the `/redmew-color` command. The randomizer has been removed as it's fairly easy to fill in a random color via the GUI for both chat and player color. I've also added a migration system which we can leave in there for a while, which will read the settings from the current colors and copy it to the new player settings storage, explicitly leaving the old values to makes sure old scenarios will still work.

The only thing left to fix is limiting the darkness of colors, which I'm not sure how to fix yet. Setting your chat color to `0 0 0` would make it unreadable. I do not consider this important to fix for this PR as we can moderate this if needed for now (which I highly doubt will happen).

Summary of changes:
 - introduced a `Settings.toScalar` which is used to make a setting value stored as table (for example) compatible with the input forms
 - moved the settings to resources so they don't clutter the logic file
 - `/redmew-color` is removed, syncing is now done by setting it in the GUI, changing the player color with `/color` is still possible and won't sync to the server
 - Settings can now be stored in a different structure than represented in the front-end (color tables in this case)

image

image

image

The above will be saved as:
image

image
image
image

Saved with alpha:
image
image

### DIFF
--- a/config.lua
+++ b/config.lua
@@ -44,6 +44,11 @@ global.config = {
         time_for_trust = 3 * 60 * 60 * 60, -- 3 hours
         everyone_is_regular = false
     },
+    -- allows syncing player colors from and to the server. Disable this if you want to enforce custom colors
+    -- when enabled, /color will also be synced to the player settings
+    player_colors = {
+        enabled = true,
+    },
     -- saves players' lives if they have a small-plane in their inventory, also adds the small-plane to the market and must therefor be loaded first
     train_saviour = {
         enabled = true
--- a/config.lua
+++ b/config.lua
@@ -276,9 +276,6 @@ global.config = {
     donator_commands = {
         enabled = true
     },
-    player_colors = {
-        enabled = true
-    },
     -- adds a command to generate a popup dialog box for players to see, useful for important announcements
     popup = {
         enabled = true
--- a/control.lua
+++ b/control.lua
@@ -27,7 +27,10 @@ require 'features.server_commands'
 require 'features.player_create'
 require 'features.rank_system'
 require 'features.redmew_settings_sync'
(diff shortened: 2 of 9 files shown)


linaori:
Some other points that still need work:
 - `/color` should probably sync the color to the server as well and update the settings
 - The config option should be reverted as PvP scenarios might not want this feature enabled

config.lua
```diff
@@ -276,9 +276,6 @@ global.config = {
     donator_commands = {
         enabled = true
     },
-    player_colors = {
```
SimonFlapse, line 279: Any plans for a new config to disable setting player and chat colors?

linaori: > The config option should be reverted as PvP scenarios might not want this feature enabled

:D

linaori:
Due to certain quirks in the factorio color naming scheme, I've somewhat forced to have Factorio use our naming schema for colors. 

Invalid color name:
image

Valid color:
image

Custom color
image

linaori:
Everything seems to work now. The only issue remaining is that when you do `/color` with an invalid color, it will sync the current player color and chat color to the settings. There's no way around this sadly as we don't know if the command failed or not

linaori:
Tables will now be checked 1 level deep to prevent tables from always saying they have changed: 
image
image

mheguy (commented):
I really like the concept and implementation as demonstrated in the PR's comments. I can't really dig through the code and/or test though, so leaving this as a comment review rather than explicit approval.

grilledham (approved):

grilledham (approved):

Issue #821, opened 2019-02-26
List of broken shit in 0.17


mheguy:
- [x] Walkabout #826 

- [x] Info gui #831 

- [x] Poll gui (create poll)

- [x] All maps that use `RS` to set single types of ground tiles ("grass", "dirt", "sand", "desert" are no longer autoplace_control_names)  #827

- [x] Terraforming uses `acid-projectile-purple` which is gone

- [x] Rotten apples lists and relies on techs that are deprecated #828

- [x] Tetris not getting a waterworld (https://forums.factorio.com/viewtopic.php?f=28&t=65371&p=401110)

Bilka2:
You can make a waterworld with `/c game.create_surface("my-surface", {  property_expression_names = {elevation = -10}}) game.player.teleport({0,0}, "my-surface")`, see #927

PR #927, opened 2019-05-29
Fixed waterworld map_gen_settings

Bilka2:
See #821


grilledham:
I've Tested, Looks good.
Thanks for the PR.

PR #929, opened 2019-05-29
don't show the debugger's gui data in gui data

grilledham:


features/gui/debug/gui_data_view.lua
```diff
@@ -109,11 +109,16 @@ function Public.show(container)
 end
 
 local function draw_element_headers(element_panel, values, selected_index)
+    local copy = {}
+    for k, v in pairs(values) do
+        copy[k] = v
```
linaori, line 114: If `v` is a table, you still have references (not sure what `v` is here)

linaori (approved):

features/gui/debug/gui_data_view.lua
grilledham: v is the data that is stored for the element with Gui.set_data.
That can be seen below on line 121: `for ei, stored_data in pairs(copy) do`
In this case I need a copy of the values table as the for loop below adds to the table as it is reading from it. Without the copy the loop never ends.

Issue #926, opened 2019-05-29
Diggy feedback/todo collection thread


linaori:
With 0.17.x running a while and an increase of popularity, I want to create a collection of feedback/bugs etc so we can track things a bit better. If it has a link, there's an existing issue for it. If it's fixed/handled or not an issue, it will be checked.

### Bugs that affect Diggy
- [x] [2 Errors in retailer](https://github.com/Refactorio/RedMew/issues/879)
- [x] [Error in hail_hydra](https://github.com/Refactorio/RedMew/issues/878)
- [x] [diggy_cave_collapse: changes to on_built_tile](https://github.com/Refactorio/RedMew/issues/815)
- [x] [SimpleRoomGenerator water in Diggy corrupts stress levels](https://github.com/Refactorio/RedMew/issues/552)

### Wishlist
 - [ ] Replace room and terrain generation by the same system that generates ores
 - [ ] The ability to use the new render function to show (per player) what the stress levels are for a certain spot, this requires #552 to be fixed first as it would should incorrect data for some spots otherwise
 - [x] Ability to set enable/disable landfill via a config settings (fixed in #930)
 - [x] Ability to set the initial mining speed (fixed in #930)
 - [x] Merge the Diggy score table into the global score table by making it a simple storage so each scenario can do this
 - [ ] Make it possible to add custom columns (per scenario) to the player list to keep track of stats (makes #887 possible)
 - [x] Localization
 - [x] Localization of score table
 - [ ] [Cave Biomes](https://github.com/Refactorio/RedMew/issues/345)
 - [ ] [Diggy integration with existing mods (primarily Bobs + Angels)](https://github.com/Refactorio/RedMew/issues/301)
 - [ ] [Abandoned outposts(loot base)](https://github.com/Refactorio/RedMew/issues/245)
 - [ ] [Biter caves](https://github.com/Refactorio/RedMew/issues/244)
 - [x] Edit foreman advice on first rock mined, to include why rocks do not give stone.
 - [x] Biter performance late game might cause lag spikes (Partial solutions already implemented)
 - [ ] Ability to always show level and progress in a small window (Like the stats window)
 - [ ] Add something significant to rocket launches https://github.com/Refactorio/RedMew/issues/926#issuecomment-497214884
 - [x] Tweak oil spawn chances https://github.com/Refactorio/RedMew/issues/926#issuecomment-497210469 (fixed in #930)

SimonFlapse:
~~Older issues related to Diggy:~~
[...]

PR #930, opened 2019-05-30
Some diggy tweaks based on feedback

linaori:
Fixes a few feedback points from #926 

 - Added an initial mining speed bonus
 - Added an option to toggle landfill research
 - Slightly increased oil spawn rates

### DIFF
--- a/features/player_create.lua
+++ b/features/player_create.lua
@@ -2,6 +2,7 @@ local table = require 'utils.table'
 local Event = require 'utils.event'
 local Global = require 'utils.global'
 local Info = require 'features.gui.info'
+local pairs = pairs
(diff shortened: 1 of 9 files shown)


SimonFlapse:
LGTM

It's nice with a dedicated setting for tweaking initial mining speed boost, instead of having players turn on cheat mode.

I think players have been unlucky in their quests for oil, because I've liked the current rarity. The increase is minor so seems it's still balanced

SimonFlapse (approved):

linaori:
Yeah, I didn't want to increase it by too much, but do notice that it's relatively rare, despite having high yields.

PR #931, opened 2019-05-30
Builder decoratives

grilledham:


SimonFlapse:
Throttling 1, 2, 1, 3:
image
gives a bit more dense grass compared to 1, 3, 1, 3:
image
While being less dense than 1, 2, 1, 2:
image

Players can turn of decoratives in-game from the graphics menu without restarting the client.

Also disabled game.draw_resource_selection

PR #934, opened 2019-05-30
Rewrite score storage

linaori:
One specific thing that might be nice to have at some point, is to show `1.1K` instead of `1100` in the GUI, but I left that out of this rewrite.

 - Internally all the labels in the GUI for the scoreboard can now be translated
 - Any component can register a new score, which can then be used as either player specific or global
 - I've added a config item for the score config which allows you to dynamically change which ones are shown (Diggy uses it in this PR)
 - Rocket launchers are now tracked via the score updated event
 - GUI only refreshes on demand rather than every X ticks, but will wait 30 ticks before rendering the change, letting accumulate any other score updates in those 30 ticks (roughly 0.5s by default)
 - By firing events when a score has changed, we can also dynamically update other GUI elements such as the player-list if needed

Internally stored as this:
image

Diggy specific changes, the scoretable from diggy is completely removed and integrates in the normal scoreboard now:
image
image
image


linaori:
Also fixes the score table locale and merge of the score table points from #926 /cc @SimonFlapse as you do a lot with translations

linaori:
Build will be fixed later today, I think something is broken in the function that tries to detect whether or not a satellite was launched

grilledham:
Just a thought, but why are we using rich text for the score gui. Why not use a sprite for the sprite and a label for the text?

linaori:
I have no idea, this is what it was originally using and I shamelessly copied it over

SimonFlapse:
I'm glad to see support for localisation

features/gui/score.lua
```diff
@@ -139,7 +135,7 @@ local function global_score_changed(event)
         return
     end
 
-    count = event.new_value
+    local count = event.new_value
```
grilledham: With the change to remove the `new_value` field form the event this no longer works.

utils/score_tracker.lua
```diff
@@ -65,6 +67,10 @@ end
 ---@param score_name string
 ---@param value number|nil number to subtract or add
```
grilledham: Why is `nil` an acceptable value, when would someone want to pass in `nil`?

linaori: Idea was to do +1 in that case, but I changed that before implementing it

grilledham (approved):

utils/score_tracker.lua
```diff
@@ -65,7 +65,7 @@ end
 ---
 ---@param player_index number
 ---@param score_name string
----@param value number|nil number to subtract or add
+---@param value number to subtract or add
 function Public.change_for_player(player_index, score_name, value)
     if not value or value == 0 then
```
grilledham: My preference would be to not silently ignore invalid inputs if the caller has no reason to do that as it hides mistakes.
So I would remove the `not value` check

linaori: This would lead a weird arithmetic error later on where it tries add nil + current

image

linaori: Fair enough, you can trace it back to the cause: 
image

PR #935, opened 2019-05-31
Cleanup player stats

linaori:
- Old player_stats methods are now replaced by the ScoreTracker
 - Using 2k, 4.4m notations for the score
 - Death causes are now translated in the player list

image
image
image

### DIFF
--- a/features/redmew_commands.lua
+++ b/features/redmew_commands.lua
@@ -7,13 +7,9 @@ local Rank = require 'features.rank_system'
 local Donator = require 'features.donator'
 local Color = require 'resources.color_presets'
 local ScoreTracker = require 'utils.score_tracker'
-local get_for_player = ScoreTracker.get_for_player
-local coins_spent_name = 'coins-spent'
-local coins_earned_name = 'coins-earned'
-local player_deaths_name = 'player-deaths'
-local player_console_chats_name = 'player-console-chats'
-local player_items_crafted_name = 'player-items-crafted'
-local player_distance_walked_name = 'player-distance-walked'
+local format_number = require 'util'.format_number
+local player_data_to_show = global.config.redmew_commands.whois.player_data_to_show
+local print_to_player = Game.player_print
 local concat = table.concat
 local tostring = tostring
 local tonumber = tonumber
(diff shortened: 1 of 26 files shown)


features/redmew_commands.lua
```diff
@@ -193,12 +201,12 @@ local function print_player_info(args, player)
...
-        {'format.1_colon_2', 'Coin spent', PlayerStats.get_coin_spent(index)},
-        {'format.1_colon_2', 'Deaths', PlayerStats.get_death_count(index)},
-        {'format.1_colon_2', 'Crafted items', PlayerStats.get_crafted_item(index)},
-        {'format.1_colon_2', 'Chat messages', PlayerStats.get_console_chat(index)}
+        {'format.1_colon_2', 'Distance walked', get_for_player(index, player_distance_walked_name)},
+        {'format.1_colon_2', 'Coin earned', get_for_player(index, coins_earned_name)},
+        {'format.1_colon_2', 'Coin spent', get_for_player(index, coins_spent_name)},
+        {'format.1_colon_2', 'Deaths', get_for_player(index, player_deaths_name)},
+        {'format.1_colon_2', 'Crafted items', get_for_player(index, player_items_crafted_name)},
+        {'format.1_colon_2', 'Chat messages', get_for_player(index, player_console_chats_name)}
```
SimonFlapse: Are these supposed to be localized strings ?

linaori: Yes, I have to dive into that part. I noticed that this part of the Whois isn't translated yet. Technically speaking I have all the translations so I'll try to put them in here before we merge it

SimonFlapse: Thanks 😄

features/redmew_commands.lua
```diff
@@ -168,45 +172,43 @@ local function print_version()
...
         print_no_target(target_ident)
         return
     end
 
-    local info_t = {
-        'redmew_commands.whois_formatter',
-        {'format.1_colon_2', 'Name', target_name},
-        {'format.single_item', target.connected and 'Online: yes' or 'Online: no'},
-        {'format.1_colon_2', 'Index', target.index},
-        {'format.1_colon_2', 'Rank', Rank.get_player_rank_name(target_name)},
```
mheguy, line 189: I'm glad to see this get a refactor. This was a half-measure when ranks were first localized and always kind of irked me. 😄

PR #936, opened 2019-06-01
Crowdin sync 2019-06-01

SimonFlapse:
Danish: 33%
German: 77%
Czech: 64%
Finnish: 8%
French: 34%
Korean: 6%
Norwegian: 85%
Brazilian: 20%
Russian: 24%
Spanish: 1%


mheguy:
It seems like you're doing these syncs by hand? I had a look at https://support.crowdin.com/github-integration/ and it looks like it will save you a lot of time in the long term.

SimonFlapse:
> It seems like you're doing these syncs by hand? I had a look at https://support.crowdin.com/github-integration/ and it looks like it will save you a lot of time in the long term.

My main problem is the way our locale files are structured. I've written a Java application for converting between cfg and a csv format that crowdin supports and back again. That part is mainly automated though, I just need to run it and copy the files to a new branch.

PR #938, opened 2019-06-01
Wall breach

SimonFlapse:
Wall breach is a feature that allows for wall reinforcement using walls, but at the cost of big and behemoth biters dealing more damage to the wall structures if they are not reinforced enough

If the concept is appealing it could be expanded, with tiles giving further reinforcements.

Big biters deal extra damage to 1 row walls, and near normal damage to 2 row walls

Behemoth biters deal extra damage to 1 row walls and 2 row walls and regular damage to 3 row walls.

Creates alerts if the walls are being breached.


features/wall_breach.lua
```diff
@@ -0,0 +1,94 @@
...
+        return
+    end
+
+    local cause_name = cause.name
+    local damage = entity.damage
+    local force = event.force
+    if not valid(force) then
+        return
+    end
+    local strength = wall_strength(entity)
```
linaori, line 69: This is going to be really heavy on performance

SimonFlapse: I've changed the implementation, I hope that is less heavy on performance

theorangeangle (approved):
Looks great!  Hopefully this adds a little more difficulty to crash site. I would like to get someone who knows LUA better than I do to also review, but otherwise it is good for merge IMO. 
(Although I do not know what quite to look for in terms of potential desync issues)

grilledham:
Using the entity damage event could make this very performance heavy. I'm not sure what to suggest though.

grilledham:
Maybe the way forward with this is to have it off by default, but let some maps opt in. Some scenarios with less biters (diggy, terraforming danger ores) might be fine with this.

SimonFlapse:
Closing because the this idea seems to impact performance more than what we benefit

PR #939, opened 2019-06-02
Updated utils.gui to use mod-gui

Cooldude2606:
Sorry for the second pr I needed to move it onto a new branch so I could pull updates. See #905

### DIFF
--- a/utils/gui.lua
+++ b/utils/gui.lua
@@ -2,6 +2,7 @@ local Token = require 'utils.token'
 local Event = require 'utils.event'
 local Game = require 'utils.game'
 local Global = require 'utils.global'
+local mod_gui = require 'mod-gui'
 
 local Gui = {}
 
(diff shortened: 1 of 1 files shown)


grilledham:
Sorry that I've been rather slow to respond to this / #905

But I do have some thoughts on this. The Gui module as it stands is just meant as a bunch of optional helper methods. It aims to be unopinionated about how people make guis. So whether a module uses the mod-gui button flow I think is up to the implementer.

That said I would accept making some changes to help with using the mod-gui button flow.

```lua
function Gui.get_top_element_flow(player)
    player = Game.get_player_from_any(player)
    return mod_gui.get_button_flow(player)
end
```
Could be changed to 
```lua
Gui.get_top_element_flow = mod_gui.get_button_flow
```
To prevent people needing two imports.


The logic inside <https://github.com/Refactorio/RedMew/pull/939/files#diff-ff2aeeff7610429fb3e8de55da65c1e2R233> could be changed to recursively check all elements in the top flow, not just direct children of top or mod_gui.get_button_flow.

PR #943, opened 2019-06-04
First shot at #608 Corpse Cleanup

Starholme:
This is a revival and partial re-write of #883 that was abandoned.

It looks at corpses within a small range of any new corpse, and randomly removes to get under the limit. Although it has to run every time an entity dies, it only touches a small area (36 tiles), so it should be fairly low impact. The number of corpses and radius are adjustable, but these values seem reasonable. The 'after' screenshot is using these settings.

I've tested what I can think of, but I'm open to suggestions.

Before:
Before
After:
After

### DIFF
--- a/features/biter_corpse_util.lua
+++ b/features/biter_corpse_util.lua
@@ -2,73 +2,53 @@
 local Event = require 'utils.event'
 local Global = require 'utils.global'
 local table = require 'utils.table'
+local fast_remove = table.fast_remove
+local pairs = pairs
 
 local biter_utils_conf = global.config.biter_corpse_util
 
-local corpse_chunks = {}
-
 -- Factorio removes corpses that hit 15 minutes anyway
 local max_corpse_age = 15 * 60 * 60
+local kills_per_cleanup = 500
+
+-- x -> y -> array of {entity, tick}
+local corpse_chunks = {}
+local cleanup_count = {kills_per_cleanup}
 
 Global.register(
     {
-        corpse_chunks = corpse_chunks
+        corpse_chunks = corpse_chunks,
+        cleanup_count = cleanup_count
     },
     function(tbl)
         corpse_chunks = tbl.corpse_chunks
+        cleanup_count = tbl.cleanup_count
     end
 )
 
 -- cleans up the stored list of corpses and chunks
 local function remove_outdated_corpses(now)
     -- loop each stored chunk
-    for cci, corpse_chunk in pairs(corpse_chunks) do
-        local count = corpse_chunk.count
+    for x, column in pairs(corpse_chunks) do
+        for y, corpses in pairs(column) do
+            local count = #corpses
 
-        -- loop stored corpses
-        local corpses = corpse_chunk.corpses
-        local i = 1
-        local corpse = corpses[i]
-        while corpse ~= nil do
-            if corpse.tick < now then
-                table.fast_remove(corpses, i)
-                count = count - 1
-            else
-                i = i + 1
+            for i = count, 1, -1 do
+                if corpses[i].tick < now then
+                    fast_remove(corpses, i)
+                    count = count - 1
+                end
             end
-            corpse = corpses[i]
+
+            if count == 0 then
+                column[y] = nil
+            end
         end
 
-        corpse_chunk.count = count
-
-        -- remove tracked chunk if no corpses
-        if count < 1 then
-            corpse_chunks[cci] = nil
+        if #column == 0 then
+            corpse_chunks[x] = nil
         end
     end
-end
-
---Remove extra corpses that are in this area
-local function corpse_cleanup(hash_position)
-
-    local corpse_chunk = corpse_chunks[hash_position]
-    local count = corpse_chunk.count
-    local corpses = corpse_chunk.corpses
-    local num_to_remove = count - biter_utils_conf.corpse_threshold
-
-    -- remove enough entities to be under the threshold
-    for i = 1, num_to_remove do
-        local corpse = corpses[i]
-        if corpse.entity.valid then
-            corpse.entity.destroy()
-        end
-
-        table.fast_remove(corpses, i)
-        count = count -1
-    end
-
-    corpse_chunk.count = count
-
 end
 
 local function biter_died(event)
@@ -88,35 +68,50 @@ local function biter_died(event)
     local tick = event.tick
 
     -- Chance to clean up old corpses and chunks
-    if tick % 60 == 0 then
+    local cuc = cleanup_count[1] - 1
+    if cuc <= 0 then
         remove_outdated_corpses(tick)
+        cleanup_count[1] = kills_per_cleanup
+    else
+        cleanup_count[1] = cuc
     end
 
-    --Calculate the hash position
-    local x = entity.position.x - (entity.position.x % biter_utils_conf.chunk_size)
-    local y = entity.position.y - (entity.position.y % biter_utils_conf.chunk_size)
-    local hash_position = x .. "_" .. y
+    --Calculate the chunk position
+    local chunk_size = biter_utils_conf.chunk_size
+    local pos = entity.position
+    local x, y = pos.x, pos.y
+    x = x - (x % chunk_size)
+    y = y - (y % chunk_size)
 
     -- check global table has this position, add if not
-    if corpse_chunks[hash_position] == nil then
-        corpse_chunks[hash_position] = {
-            count = 0,
-            corpses = {}
-        }
+    local corpse_chunks_column = corpse_chunks[x]
+    if corpse_chunks_column == nil then
+        corpse_chunks_column = {}
+        corpse_chunks[x] = corpse_chunks_column
     end
 
-    -- get and increment this chunk, add this entity
-    local corpse_chunk = corpse_chunks[hash_position];
-    local count = corpse_chunk.count + 1
-    corpse_chunk.count = count
-    corpse_chunk.corpses[count] = {
+    local corpses = corpse_chunks_column[y]
+    if corpses == nil then
+        corpses = {}
+        corpse_chunks_column[y] = corpses
+    end
+
+    -- Add this entity
+    local count = #corpses + 1
+    corpses[count] = {
         entity = entity,
         tick = tick + max_corpse_age
     }
 
-    -- Call cleanup if above threshold
-    if corpse_chunk.count > biter_utils_conf.corpse_threshold then
-        corpse_cleanup(hash_position)
+    -- Cleanup old corpse if above threshold
+    if count > biter_utils_conf.corpse_threshold then
+        local old_entity = corpses[1].entity
+
+        if old_entity.valid then
+            old_entity.destroy()
+        end
+
+        fast_remove(corpses, 1)
     end
 end
 
--- a/features/biter_corpse_util.lua
+++ b/features/biter_corpse_util.lua
@@ -91,8 +91,8 @@ local function biter_died(event)
     end
 
     --Calculate the hash position
-    local x = entity.position.x - (entity.position.x % biter_utils_conf.radius)
-    local y = entity.position.y - (entity.position.y % biter_utils_conf.radius)
+    local x = entity.position.x - (entity.position.x % biter_utils_conf.chunk_size)
+    local y = entity.position.y - (entity.position.y % biter_utils_conf.chunk_size)
     local hash_position = x .. "_" .. y
 
     -- check global table has this position, add if not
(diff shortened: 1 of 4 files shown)


features/biter_corpse_util.lua
```diff
@@ -0,0 +1,47 @@
...
+
+local biter_utils_conf = global.config.biter_corpse_util
+
+local random = math.random
+
+-- currently no on_corpse_spawned event, using on_entity_died instead
+
+local function biter_died(event)
+    local entity = event.entity
+    if entity.valid then
```
linaori: Can you turn those into guard clauses to reduce nesting?
```lua
if not entity.valid then
    return
end
```
```

File: features/biter_corpse_util.lua [hunk 2]
```diff
@@ -0,0 +1,47 @@
...
+        local dying_force = entity.force
+        -- ignore player owned entities
+        if dying_force == 'player' then
+            return
+        end
+
+        local surface = entity.surface
+
+        -- More than the desired number of corpses?
+        if surface.count_entities_filtered {
```
linaori: Same in regards of the guard clause

features/biter_corpse_util.lua
```diff
@@ -0,0 +1,47 @@
...
+local random = math.random
+
+-- currently no on_corpse_spawned event, using on_entity_died instead
+
+local function biter_died(event)
+    local entity = event.entity
+    if entity.valid then
+        local dying_force = entity.force
+        -- ignore player owned entities
+        if dying_force == 'player' then
```
grilledham: I don't think this works unless you do `dying_force.name == 'player'

features/biter_corpse_util.lua
grilledham: The table that is passed to `count_entities_filtered ` is the same as the table that is passed to `find_entities_filtered ` can we make a `local filter = {...}` and reuse it?

features/biter_corpse_util.lua
```diff
@@ -0,0 +1,47 @@
...
+                } > biter_utils_conf.corpse_threshold then
+            -- Get the actual entities
+            local corpse_list = surface.find_entities_filtered {
+                    position = entity.position,
+                    radius = biter_utils_conf.radius,
+                    type = 'corpse',
+                    force = 'neutral'
+                }
+
+            local num_to_remove = #corpse_list - biter_utils_conf.corpse_threshold
```
grilledham: Just because this code might potentially run quite a lot can I request the micro optimisation of making a `local count = #corpse_list`?

linaori (changes requested):
I like the idea of it being random, but this is going to be a really heavy script. What if instead of doing this every biter death, the position is marked for cleanup at a later point? It would roughly translate to this:

1) a biter dies & entity.valid
2) 5% chance of garbage collection
3) is the location (+ range) already marked for cleanup, stop
4) set down marker for this position (+ range)
5) register event to clean location in X ticks (say 9000 ticks)
6) do the cleanup as you have here (random is fine I guess)

Another option is to keep trick of all actual positions of entities that died and at what tick, then you can calculate back which entities should be removed on those positions

grilledham (commented):
Thanks for the PR.

I've suggested one bug fix and 2 micro optimisations. 

Thinking about it further would it make more sense to only listen for `dying_force.name == 'biter'` rather than not `'player'`? If for example trees being grenaded counts as neutral force dying.

Also what happens when non biter entities on the enemy force die, like say when outpost turrets are destroyed in crashsite. Do we want that triggering the corpse removal logic?

Performance wise, it's hard to predict what can cause bad performance without trying it. So I'm willing to try this and see what happens.

features/biter_corpse_util.lua
Starholme: Is there a list of these values somewhere? I'm sure there has to be, I just can't find anything...

features/biter_corpse_util.lua
Starholme: I tried it like this and it didn't work. Just found every entity:
```
local filter={
                position = entity.position,
                radius = biter_utils_conf.radius,
                type = 'corpse',
                force = 'neutral'
                }
surface.count_entities_filtered {filter}
```
Seems like it should be possible though. Any suggestions?

features/biter_corpse_util.lua
grilledham: <https://lua-api.factorio.com/latest/LuaForce.html>

<https://lua-api.factorio.com/latest/LuaForce.html#LuaForce.name> for the name property

<https://lua-api.factorio.com/latest/> for the whole modding api

features/biter_corpse_util.lua
grilledham: I was suggesting this
```lua
local filter = {
    position = entity.position,
    radius = biter_utils_conf.radius,
    type = 'corpse',
    force = 'neutral'
}
if surface.count_entities_filtered(filter) > biter_utils_conf.corpse_threshold then
    -- Get the actual entities
    local corpse_list = surface.find_entities_filtered(filter)

    -- rest of code
end
```
Which I think should work

features/biter_corpse_util.lua
Starholme: Done, now checks that they are part of the default 'enemy' force.

features/biter_corpse_util.lua
Starholme: Thanks, I knew it should be possible, but I'm pretty new to lua. Mostly a c# guy.
Should be resolved in the latest commit.

Starholme:
> I like the idea of it being random, but this is going to be a really heavy script. What if instead of doing this every biter death, the position is marked for cleanup at a later point? It would roughly translate to this:
> 
> 1. a biter dies & entity.valid
> 2. 5% chance of garbage collection
> 3. is the location (+ range) already marked for cleanup, stop
> 4. set down marker for this position (+ range)
> 5. register event to clean location in X ticks (say 9000 ticks)
> 6. do the cleanup as you have here (random is fine I guess)
> 
> Another option is to keep trick of all actual positions of entities that died and at what tick, then you can calculate back which entities should be removed on those positions

I'm concerned that we will end up with one 'big' slowdown every 9000 ticks, instead of a 'small' slowdown on each death. I could implement point 2 though, just ignore x% of the calls. Would avoid the overhead of counting entities each time.
I don't think we can actually find out which would work better without building and testing both.

features/biter_corpse_util.lua
```diff
@@ -0,0 +1,53 @@
...
+
+    local filter = {
+        position = entity.position,
+        radius = biter_utils_conf.radius,
+        type = 'corpse',
+        force = 'neutral'
+    }
+
+    -- More than the desired number of corpses?
+    if not (surface.count_entities_filtered(filter)  > biter_utils_conf.corpse_threshold) then
```
grilledham: Rather than `not` you could invert `>` to `<=`.
I personally find it reads better that way.

grilledham (approved):
While I think linaori suggestion could be more performant, I'm willing to give this a try and see what happens.

features/biter_corpse_util.lua
Starholme: Good point. Done.

linaori:
The thing I'm most worried of, is that we will do 2 surface lookups for a big portion of death events. Diggy stress map proves that it's not a performance issue to store a massive table of coordinates and in this scenario we'd only store 0~20 max or so. Every X ticks 2 surface lookups with a random to kill corpses is far more performant than doing this every few deaths.

It will also feel more natural to not immediately see corpses disappear 😄

mheguy (approved):
Reviewing since I was requested:
I'm too out of the loop to have an informed opinion on what the current scenario can and can't handle performance-wise.
With that said, I like seeing initiative and someone willing to write code. I think this definitely deserves a shot and can be modified if needed.

Starholme:
Ok, complete rewrite. It's using the basic structure suggested by @linaori above.

Store each corpse, from the post_entity_died event, into a table grouped by the position.
If a position has more than the threshold, remove some.
Once and a while, trim the table by removing any old corpses and empty positions.
Does zero surface api calls.

If you are trying it out, it's most obvious to see the behavior if you either increase the chunk size or reduce the corpse threshold. At the current settings you really need to be looking to see that it's removing much.

@grilledham @plague006 Please review again as the code is entirely changed.

grilledham:
This works a bit differently to how I imagined it.

 `entity.position.x % biter_utils_conf.chunk_size`
Calculating the hash position using a module strikes me as odd. Wouldn't that mean every n tiles counts as being in the same chunk? What about:
`math.floor(entity.position.x / biter_utils_conf.chunk_size)`

I'm not sure if the `remove_outdated_corpses()` function is worth it. I think the `corpse_cleanup()` function is sufficient to do the clean up. What are your thoughts on this?

I haven't tested this yet, can you recommend a good way to test it?

Starholme:
> This works a bit differently to how I imagined it.
> 
> `entity.position.x % biter_utils_conf.chunk_size`
> Calculating the hash position using a module strikes me as odd. Wouldn't that mean every n tiles counts as being in the same chunk? What about:
> `math.floor(entity.position.x / biter_utils_conf.chunk_size)`
> 
> I'm not sure if the `remove_outdated_corpses()` function is worth it. I think the `corpse_cleanup()` function is sufficient to do the clean up. What are your thoughts on this?
> 
> I haven't tested this yet, can you recommend a good way to test it?

With the full line being entity.position.x - (entity.position.x % biter_utils_conf.chunk_size), it makes sense.
Example: x = 47, so 47 - (47 % 3) = 47 - 2 = 45
Essentially gives you the same result as the math.floor you mention.

corpse_cleanup is responsible for removing excessive corpses from the map.
remove_outdated_corpses is all about pruning the stored list, so we don't hang onto extra entities or chunks when they no longer make sense (ie: factorio has removed the corpse because it is more than 15 minutes old)

Testing for me was like this:
- Start new scenario, set player to god, research all techs, set always day
- Find a biter nest, plop down a few turrets full of ammo
- Crank up the game speed
- See how things go
To test the remove_outdated_corpses, you'll need to remove the first set of turrets, go find a nest somewhere else and set them up. Probably want to add a print statement when it removes an empty corpse_chunk on line 46. If you print the key, it will be the x and y.

grilledham:
> With the full line being entity.position.x - (entity.position.x % biter_utils_conf.chunk_size), it makes sense.

Right you are, my mistake.

As a suggestion, what about we only keep the corpse count for each chunk. Then once it exceeds a limit we get all the corpses for the chunk, remove half and update the count.

That way we don't need to clear out expired corpses. Plus storing data in global increases save / load time.

Thoughts?

grilledham:
For the diggy map we had to store lots of data by position / chunk.
So we tested 3 different ways to do that:
 - string hash (what you are doing)
 - number hash
 - table of tables

Turns out table of tables is the fastest.

Starholme:
@grilledham I think the big difference here is that we only ever expect to see a few dozen 'chunks', each with less than 20 corpses. The amount of data is never very large, with a worst case being something like 100 chunks * 20 corpses = 2000 total iterations.

The tradeoff comes down to mine requiring more looping vs yours requiring more surface api calls (because you only have the count, you have to check the map)

It's a little difficult to benchmark some of these things, hard to get a concrete answer.

Starholme:
On discord, @linaori mentioned that this can be merged

grilledham:
@Starholme 

There are a few micro optimisations that I see no harm in adding, like making some local variables where values are used more than once. Also I would like to turn the string hash into a table of tables.

Would you like me to commit the changes directly into this PR, or would you prefer I outline the changes and you make them?

Starholme:
@grilledham Go ahead and make the changes

grilledham:
I ended up playing around with this quite a bit.
Changes I made:
- Covert hash to table of tables
- Used locals for repeated values.
- Noticed that `corpse_cleanup` would only ever clean up one corpse per call, so I inlined the logic.
- Changed the old corpse cleanup logic to be based on biters killed rather than if a biter was killed in the right tick. This removes a problem where if lots of biters are killed on the right tick (say a grenade) the cleanup logic would be called multiple times in the same tick.
- Changed the corpse_threshold from 10 to 3. This is just my personal opinion, if someone wants to change the config value they can.
- Renamed the module to `biter_corpse_remover`

I think a good way to test this is to put the following in map_selection.lua
```lua
_DEBUG = true
_CHEATS = true

local spawners = b.throttle_world_xy(b.full_shape, 1, 8, 1, 8)
spawners = b.subtract(spawners, b.circle(32))
spawners = b.entity(spawners, 'biter-spawner')

map = b.apply_entity(b.full_shape, spawners)

return map
```

linaori (approved):

PR #947, opened 2019-06-08
Cutscene Controller

SimonFlapse:
A new cutscene controller to take advantage of cutscenes.

Includes a helper module to make use of rendering during cutscenes.

Implemented for Diggy (Added sound effect)

Features:
**Cutscene controller:**
- [x] Callback function triggered on the cutscene event
- [x] Save waypoints and callback functions as a preset
- [x] Keeps track of all running cutscenes
- [x] Keeps track of disconnected characters 
- [x] Handle players that leave mid cutscene
- [x] Waypoint injection to running cutscenes by restarting them.
- [x] Disconnect players from character before cutscene is started
- [x] Restoring character when cutscene is over, terminated or if player left
- [x] Fix players not being able to interact with any entities (But able to move around)
- [x] Termination that connects players back to their characters
- [x] Allows to register renderings to current running cutscenes so that /skip can destroy them.
- [x] Added a termination callback function that runs if the cutscene is terminated (Cutscenes that have been updated will terminate at the end)
- [x] /skip command to exit cutscene
- [x] skip button to exit cutscene
- [x] /replay command to replay introduction cutscene
- [x] Prevents starting a new cutscene when another one is already running and controlled by the cutscene controller

**Rendering:**

- [x] Draw_text in a "dynamic" way, that adjusts to players resolution and zoom level while preserving the intended look.
- [x] Draw_multi_line_text, that handles multiline text in a "dynamic" way.
- [x] Draw_rectangle in a "dynamic" way that adjusts to players resolution and zoom level while preserving the intended look.
- [x] Blackout, that fills the entire screen with a color.
- [x] Allow passing through parameters
- [x] Returns index or indices for created rendering elements

**Radio**

- [x] Added a radio GUI when _DEBUG = true
- [x] Open radio GUI with /radio command
- [x] Play sounds by pressing play
- [x] Allow copying sound path from GUI

**Todo**
- [x] Localisation
- [x] Added usage of Debug.print() and code cleanup
- [x] Streamline rendering
- [x] Remake calculations for dynamic changes to positions based on players screen resolution
- [x] Remake calculations for dynamic changes to position and scale based on player zoom
- [x] New calculations for converting default font to pixel in height and length. (Might need official informations about font size)


features/redmew_commands.lua
```diff
@@ -212,6 +212,15 @@ local function print_player_info(args, player)
     end
 end
 
+local function skip_cutscene(_, player)
+    if not player or not player.valid then
+        return
+    end
+    if player.controller_type == defines.controllers.cutscene then
+        require 'features.cutscene.cutscene_controller'.terminate_cutscene(player.index)
```
mheguy: You can't `require` after the data stage so I'm not sure what advantages this has over a typical require at the top of the file?

locale/en/redmew_command_text.cfg
```diff
@@ -21,6 +21,7 @@ server_time=Prints the server's time.
 seeds=List the seeds of all surfaces
 redmew_version=Prints the version of the Redmew scenario
 whois=provides information about a given player, admins can see the inventory of a player by adding "yes" as a second argument
+skip=Stops current running cutscene
```
mheguy: "Skips the current cutscene" might be better phrasing

features/redmew_commands.lua
SimonFlapse: No advantage, I was just too tired to put it in the right place

locale/en/redmew_command_text.cfg
SimonFlapse: I thought of that, but that makes it sound as if it jumps to the next waypoint.

linaori (approved):
Looks awesome!

SimonFlapse:
Additions to consider:

- [ ] Countdown, showing seconds till cutscene is finished
- [x] Powerpoint like control. Allowing to go to previous and next waypoint
- [ ] Settings to disable introduction cutscene for regulars

theorangeangle (approved):
I see no reason not to approve? We have waited on this for a while and it would be nice to have the cutscene features for future map modifications.

theorangeangle:
Follow up:

-We can do some beta testing this week/weekend if you want to ensure there is stability
-Once merged, we need to write-up an announcement to add in for the scenario data and version history.

theorangeangle:
Note: replay command does not work while on probation.

PR #950, opened 2019-06-18
[WIP] Diggy Experimental (foundation for multi surface diggy)

SimonFlapse:
Based off `Dark Forest` by `Bienenstock`

Changes to Diggy, allowing other entities than rocks.

image

### DIFF
--- a/features/player_stats.lua
+++ b/features/player_stats.lua
@@ -133,7 +133,12 @@ local function robot_built_entity()
 end
 
 local function biter_kill_counter(event)
-    if event.entity.force.name == 'enemy' then
+    local entity = event.entity
+    if not entity or not entity.valid then
+        return
+    end
+
+    if entity.force.name == 'enemy' then
         change_for_global(aliens_killed_name, 1)
     end
 end
--- a/map_gen/maps/diggy/feature/coin_gathering.lua
+++ b/map_gen/maps/diggy/feature/coin_gathering.lua
@@ -59,6 +59,10 @@ function CoinGathering.register(config)
 
     Event.add(defines.events.on_entity_died, function (event)
         local entity = event.entity
+        if not entity or not entity.valid then
+            return
+        end
+
         local force = entity.force
         if force.name ~= 'enemy' or random() > alien_coin_drop_chance then
             return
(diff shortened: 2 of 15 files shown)


theorangeangle:
Do you want to beta test this sometime? Or just throw in the scenario/run publicly

SimonFlapse:
We can do a public run, and merge if it goes well.

I'm still a bit conflicted regarding how to keep diggy and diggy experimental updated with minimal manual work. Right now i'm referencing all diggy files from diggy experimental, and replacing the scripts that has changed for diggy experimental.

theorangeangle:
@grilledham should we implement different “modes” for Diggy in the same way it was done for crash site?

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

grilledham:
I don't really know the Diggy code very well, so I don't know how easy it would be. But multiple modes / presets for Diggy would make the most sense to me. 
We could even add it as a scenario on the web panel, like there is for the two crashsite variants.

theorangeangle:
I'm going to start testing on thursday in an effort to clear out some of the old PRs and add maps to the scenario.

BlueRock2:
Hello I am very curious about this diggy version.   Does it completely do away with the ceiling support system?   Or are there separate indoor/outdoor areas?   I would be happy to help test or give feedback.

SimonFlapse:
@BlueRock2 The current state of it has ceiling support disabled. But I don't see why it couldn't be supported. The PR is stale though, due to lack of time on my end. My vision is to use this for a multi surface Diggy. Where this would be the overworld, and the classic diggy as the underground.

BlueRock2:
Oh how I wish trains could travel between surfaces...  Would love to have indoor and outdoor areas in diggy.   Other than just supports and scenery could do some interesting things to change up the rules by being indoors vs. outdoors.

SimonFlapse:
I'm not sure, but you could mix together some logic to copy a train and its contents between surfaces. You just need to figure out where the cave entrance would be. (Maybe limited map size in a single direction/part of the map kept as void. Do you could like around it)

PR #955, opened 2019-06-20
Highlight top buttons

### LINKED ISSUE #364: Highlight menu items that are active
I honestly have no idea how this is supposed to work, but it would be a really awesome feature. Just now I was fanatically clicking all of them to find the one I closed

grilledham:
When the gui is open highlight the top button
image

Closes #364

The sprite does grow slightly in size when highlighted. Not sure why I tried setting the padding the same before and after and didn't change it. But maybe it's not a problem

### DIFF
--- a/features/gui/info.lua
+++ b/features/gui/info.lua
@@ -79,7 +79,8 @@ end
 local changelog_callback = Token.register(process_changelog)
 
 local function prepare_title()
-    local welcome_title = [[
+    local welcome_title =
+        [[
 111111  1111111 111111  111    111 1111111 11     11
 11   11 11      11   11 1111  1111 11      11     11
 111111  11111   11   11 11 1111 11 11111   11  1  11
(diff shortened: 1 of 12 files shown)


features/gui/info.lua
```diff
@@ -652,13 +661,20 @@ end
...
+    local center = gui.center
     local main_frame = center[main_frame_name]
+    local main_button = gui.top[main_button_name]
 
     if main_frame then
         upload_changelog(event)
         Gui.destroy(main_frame)
+        main_button.style = 'icon_button'
     else
+        main_button.style = 'selected_slot_button'
```
linaori, line 677: it's easy to forget adding this when you add a button to the top, it's also error prone as it's pretty much copy-paste. Would it be possible to create a small API where we can talk to, to use this "toggle" feature as it's more than just rendering a window or not right now? 

Perhaps @Cooldude2606 has an idea for this as he made something really nice for something like this already

grilledham: I agree this is not ideal.
The problem comes down to how does the button know if it should be in the toggled state or not.
I don't have a solution to that. But I'm open to discuss ideas.

Cooldude2606: Toggled states are hard, I have one instance in my code when I use them, the way in which I check if it should be toggled is by using my store module; but the general idea I can still explain.
If you have a table that stores the current "state" of the toggle then you can place the button inside a "boarderless flow" (flow with no padding) and set either its name or caption (both work) to be the key to look up in the "toggle table" that you store and get the state for that button. See my player list action button for my one example.
A note on it chaning size: it isnt chaning size it is just that the slot buttons are made to go next to each other where as the other button styles have a built-in effect which makes them appear to pop out slightly. To make them appear the same size try putting it in a flow, adding some padding but then making the button smaller.

grilledham: I refactored the modules that had multiple ways in which the main frame could be closed to have a single `close_main_frame` function. Without a non minor change to way we write the gui code I don't think we can do any better.

Regarding the icon size change:
I'm not sure if I misunderstood you cooldude, but the button itself stays the same size, it's the sprite inside the button that changes size. 

I can live with the sprite changing size, however if anyone wants to look into it they can.

Cooldude2606: Sorry, I thought you meant that the size the the button looked was different without the size being different. I have had this issue before which is why I thought you meant this.

PR #956, opened 2019-06-20
fix desync with active turret delay

grilledham:
I'm pretty sure this fixes the desyncs we saw today on crashsite. 
I've tested on s9 without theses changes and I can cause a desync.
With these changes I cannot.


grilledham:
As this is a hotfix I'm going to merge this now

Issue #906, opened 2019-05-06
Miscellaneous Map Modifications


theorangeangle:
With 0.17.35 (and 0.17 to an extent) breaking all current saves available at redmew.com/saves, I have started going through all old maps in \map_gen\maps\, noting any issues as they are run for public playthroughs. 

Most of these are small changes that can be fixed easily with one or two lines of code, and are a great start for any new contributors. 

Along with running the maps for any issues, it is also a good time in my opinion to start writing descriptions and information for each map, to eventually write on the website as a listing for all maps.

### If you were linked here as "easy ways to contribute to the redmew code" or something similar, pick a comment not marked as resolved and read it. Once you have chosen a map, make the changes and [open a PR](https://github.com/Refactorio/RedMew/pulls).

theorangeangle:
Template for map testing:

**Map Name**

*Version Tested*

*Codebase/Feature Issues* Issues with features/modules not directly related to map generation, map difficulty being very unbalanced, or anything that doesn't fit in map generation issues below.

*Map Generation Issues* Issues with resource distribution, terrain not looking like it is supposed to, parts of the map being unreachable/unplayable, etc.

*Short Description* (for map info tab in-game) A couple sentences that describe the map and its changes that make it different from a normal, vanilla playthough. This section is typically where any jokes, puns, etc go along with any rich text, which can often help describe map features easier (using icons in danger_ores to show what is allowed versus a text list is a good example of rich text)

*Extra Information* (for map info tab in-game) A list of changes, put simply and with one change to a line. Things listed here are core changes (hail hydra enabled), not little things or information that could change based on the map seed.

theorangeangle:
**Bacon Islands**

*Version*: 0.17.36 

*Codebase Issues*: Map preview image missing, landfill not disabled

*Map Generation Issues*: N/A

**Update** #907 should fix this.

theorangeangle:
**Lines and Balls**

*Version:* 0.17.36

*Codebase Issues*: N/A

*Map Generation Issues*: All balls generate the same ore pattern

**Update** #907 should fix this.

theorangeangle:
**Beach**

*Version*: 0.17.37

*Codebase Issues*: I changed the [day/night cycle](https://github.com/Refactorio/RedMew/wiki/Day-Night-cycle) to the bright setting in config.lua, people seemed to be enjoying the change 
 
Missing map preview image 
Market might need rebalance?

*Map Generation Issues*:

[...]

PR #979, opened 2019-09-09
Changes to Maltease Crossings as per #906

Jayefuu:


map_gen/maps/maltease_crossings.lua
```diff
@@ -24,16 +56,6 @@ local function value(base, mult, pow)
     end
 end
 
-local function no_trees(world, tile)
```
SimonFlapse: If you still want to remove trees you can use the redmew surface along with this preset:

https://github.com/Refactorio/RedMew/blob/c6fd8ec4ea7002ed8057dc3cad7a6df6d987ecb9/resources/map_gen_settings.lua#L235

Jayefuu: Does that remove all trees on the surface or can you apply it to a shape? I only wanted to remove trees from particular islands which I think is why Plague didn't add it to the map_gen_settings section of this map when he did a pass through.

SimonFlapse: It removes it from the mapgen, you would then have to apply trees yourself, when needed.

Jayefuu: Ok that's what I thought. Then in this case the no_trees function is more appropriate because it's only 4 small islands I want them removed from to allow worms to be spawned without colliding with trees.

Jayefuu: Maybe we should add no_trees to the builders?

grilledham: Along with the no_enemies and no_ores effects

SimonFlapse: @grilledham would that make sense? Or do we already have a way of removing entities from the map gen through the builder?

SimonFlapse: > `@grilledham` would that make sense? Or do we already have a way of removing entities from the map gen through the builder?

You beat me to it :)

Jayefuu: Nothing currently in the builder to that effect, though several of the maps I wrote have similar functions that I copied from grilledham's earlier maps, so they certainly got enough reuse to go in builders.

SimonFlapse (approved):
If you're good with not being able to remove trees, then it LGTM

map_gen/maps/maltease_crossings.lua
SimonFlapse: The builders now have the functionality to remove trees (Added in https://github.com/Refactorio/RedMew/pull/979/commits/b6bcc70194cf840a8936255867f5a0ff36fd08dc)

theorangeangle:
scheduled for beta test on sunday (4/4)

theorangeangle:
Map might need some balance changes but in terms of "does it work as intended" the answer is yes

PR #980, opened 2019-09-11
Added rail_grid map preset and preview

Jayefuu:
A map preset with grass rail grid to enforce rail blocks. Played and tested with modded but also suitable for vanilla.


map_gen/maps/rail_grid.lua
```diff
@@ -0,0 +1,158 @@
...
+local tile_map ={
+    ['grass-1'] = 'dirt-1',
+    ['grass-2'] = 'dirt-2',
+    ['grass-3'] = 'dirt-3',
+    ['grass-4'] = 'dirt-4',
+}
+
+-- replace grass tiles with dirt so that the rail grid is much
+for old_tile, new_tile in pairs(tile_map) do
+    map = b.change_map_gen_tile(map, old_tile, new_tile)
```
grilledham: I want to add a function to builders that would make this easier/ more efficient.

Jayefuu: I'm not sure why Soggs changed it to this. Previously I added a function to RS map_gen_settings that allowed everything but grass. Something to do with the RSO starting areas we needed for the modded run maybe meant we had to change to changing map gen tiles instead?

theorangeangle:
Do you have ideas for a description (and bullet-point info)?

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

Jayefuu:
> 
> 
> Do you have ideas for a description (and bullet-point info)?
> 
> Sent with [GitHawk](http://githawk.com)

See line 27? I added a map description already.

theorangeangle:
Ok, will copy/paste them onto the wiki now :D 

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

map_gen/maps/rail_grid.lua
```diff
@@ -0,0 +1,154 @@
...
+pattern.
+
+You cannot place rail on any tile type except landfill. There is space at the
+grid intersections for junctions and turnarounds. There is space for
+two stations on each side of the grid.
+]]
+)
+
+
+RS.set_spawn_position({x = 20,y = 20})
```
grilledham: Is there a reason this is needed over say translating the map?
The advantage of translating the map is it will work even if another script disables the redmew surface.

Jayefuu: No, personally I always do map = b.translate(map,20,20). Soggs added the spawn position change though. I suppose one advantage is its clear from the function name why the line was added.

map_gen/maps/rail_grid.lua
```diff
@@ -0,0 +1,154 @@
...
+        local entity = event.created_entity
+        if not entity or not entity.valid then
+            return
+        end
+        local name = entity.name
+
+		local ghost = false
+        if name == 'tile-ghost' or name == 'entity-ghost' then
+            ghost = true
+			ghost_name = entity.ghost_name
```
grilledham: `ghost_name` is a global, you don't want to be doing that.
I suggest changing the line to `name = entity.ghost_name`
Then the logic below becomes simpler.
```lua
if (name ~= 'straight-rail') and (name ~= 'curved-rail') then
			if not ghost then
			    return
			elseif (ghost_name ~= 'straight-rail') and (ghost_name ~= 'curved-rail') then
				return
			end
        end
```
to
```lua
if (name ~= 'straight-rail') and (name ~= 'curved-rail') then
    return
end
```

map_gen/maps/rail_grid.lua
```diff
@@ -0,0 +1,154 @@
...
+                    break
+                end
+            end
+        end
+        if status == true then
+            return
+        else
+            --destroy entity and return to player
+            if not p or not p.valid then
+               return
```
grilledham: I think if you are going to do this check it would make sense to do it before you access `p.surface` above.

map_gen/maps/rail_grid.lua
```diff
@@ -0,0 +1,154 @@
...
+			end
+        end
+
+        -- Check the bounding box for the tile
+        local status = true
+        local area = entity.bounding_box
+        local left_top = area.left_top
+        local right_bottom = area.right_bottom
+        local p = game.get_player(event.player_index)
+        --check for sand under all tiles in bounding box
```
grilledham: Comment is wrong, should say it's checking for a tile that is not landfill.

map_gen/maps/rail_grid.lua
grilledham: Yeah I think it had something to do with RSO, but you would have to ask Soggs for the details.

grilledham (changes requested):
I think this is mostly good. A few small changes to the prevent rail logic.

Also we will have to fix the travis linting issues. Ask if you need help with that.

Then this should be good to go.

Jayefuu:
Thanks, those all make sense. I will try get to them soon.

grilledham:
I refactored this. 
I made a rail_grid directory and separated the parts into different files, this should make reuse of the code easier.

grilledham (approved):

PR #981, opened 2019-09-11
Updated map descriptions and landfill tech from maps named in issue #906

Jayefuu:


map_gen/maps/grid_islands_rotated.lua
```diff
@@ -177,4 +197,14 @@ map = b.any {map, paths, sea}
 map = b.change_map_gen_collision_tile(map, 'water-tile', 'grass-1')
 map = b.rotate(map,degrees(45))
 
+
+-- Disable landfill technology
+local function on_init()
+    local player_force = game.forces.player
+    player_force.technologies['landfill'].enabled = false           -- disable landfill
```
SimonFlapse, line 204: You removed this part:
> -- For best balance run the following commands after map generation:
> -- /silent-command game.forces["player"].technologies["landfill"].enabled = false
> -- /silent-command game.forces.player.character_running_speed_modifier = 1.5
> -- /silent-command game.difficulty_settings.technology_price_multiplier=2

Should we include it in the on_init() function?

```suggestion
    player_force.technologies['landfill'].enabled = false           -- disable landfill
    player_force.character_running_speed_modifier = 1.5
    game.difficulty_settings.technology_price_multiplier = 2
```

Jayefuu: I think I tested that and it didn't work.  I didn't have time to investigate why.

SimonFlapse: I just double checked them, and they should work unless it's being set else where

Jayefuu: No idea what I was doing wrong then. If you've tested the changes then they should be committed.

SimonFlapse:
Does the other grid island also need the technology and running speed increase?

SimonFlapse (approved):

map_gen/maps/gears.lua
```diff
@@ -14,12 +14,24 @@ RS.set_map_gen_settings(
 
 ScenarioInfo.set_map_name('Gears')
 ScenarioInfo.set_map_description(
-[[
-Am I inside an assembling machine?
-
+    [[
+Large gear shaped islands
+
```
SimonFlapse: To fix the whitespace
```suggestion
```

theorangeangle (approved):

Issue #957, opened 2019-06-21
Crash Site Feedback/todo collection thread 


theorangeangle:
Similar to our Diggy feedback issue, we have accumulated a lot of feedback on Crash Site since the last major updates. With autofill being added scenario-wide, it might be time to add some new features and fix existing "exploits" that break the game balance a little.

**Bugs, Refactoring and Balance Issues**
- [ ] Keeping one turret alive at an outpost keeps it relatively defended from biter pathfinding
- [x] Shooting a power pole at an outpost makes destroying the laser turrets much easier( _fixed by #959_)
- [ ] Late-game is too easy after MK2 armor
- [ ] ~~Hail Hydra [causes lag issues](https://clips.twitch.tv/CleanAverageCheeseArsonNoSexy) when nuking spawners (late-game spawner clearing)~~
- [ ] Rename the different crash site modes as "biomes" or "planets".

**Wishlist/Suggested Ideas**

- [x] Spawn combat robots in late-game outposts as a defense mechanism (#659)
- [ ] More random events in general
- [ ] Change the furnace at spawn to a damaged assembler if possible?  
  - Cause all outpost to pause production for x time to add a bit of a resource crunch
  - A popup shows up saying "the rocket is falling back to earth. abort mission!" and then random artillery strikes start happening in the base
  - "Supply Drops" with the wreckage item
- [ ] Respawn timer shouldn't be as long or maybe timer starts off small and increases per death or maybe based off evolution
- [ ] Decrease the timer for refilling outpost turrets, notably flamethrower turrets
- [ ] New outposts that are different than the standard "wall with assembler or chest", notably some suggested by [Naquada](https://discordapp.com/channels/432567222481846283/478339417245876256/571808225880440844)
- [ ] More coins in the one-off "loot" outposts
- [x] An option in the market to "reveal" an outpost with a map marker and/or chunk reveal
- [x] Vote-restart, win condition, or some semi-automated way of restarting the map
- [ ] Making assemblers destructible once the outpost is captured 
- [ ] Not allowing assemblers to operate without the outpost captured
- [ ] Some balancing for if the person is in SP/ an option in the config to set an "easy" mode to give solo/co-op players a fighting chance 
- [x] Add cutscene tutorial for joining the game the first time, similar to Diggy.

**Jayefuu's To Do list:**
- [ ] Make nuclear outposts harder to capture by changing the turrets to a different type/mix
[...]

Issue #977, opened 2019-09-05
Map Idea - Escape Earth


HySpeed:
# Map Idea: Escape Earth

## Concept

This scenario combines these elements:
* **Game Map of Earth**: Factorio World / World Map scenarios
* **Outposts & resource**: Crash Site scenario
* **Market purchases & upgrades**: Crash Site scenario
* **Static Rocket Silo location**: Frontier scenario

## Player Description
You have returned to Earth and find it overrun with aliens.  The goal is to escape Earth, but the technology to build a rocket silo has been lost.  A Rocket Silo is located in South America -- the landing site is in Africa.  To get to the silo, the aliens have to be battled and a path forged across the world.  Surface resources are gone.  Resources are produced by mines that must be captured and powered.  
[optional]
The scans before landing showed a possible alternate rocket silo on a remote island in the South Pacific - getting resources to that location will be difficult because of limited throughput. 

### Earth Map
The Earth map used by the Factorio World scenario is scalable.  Using this technology, a world map can be created, modified in the editor, then exported for use as a static map.  The optimal size is unknown (noted below in 'Tricky Parts').  Also, see below about Land Fill.
* In the map editor, a land bridge must be created between northeast Russia and Alaska (Arctic circle).  This bridge will be wide enough for train lines and power connections. (e.g. minimum 6 tiles).
* In the map editor, Central America / Panama must be wide enough for train and power connections also.
* If the 'alternate rocket silo' is included, a narrow land bridge must be created in the editor which allows only a single belt and power connection to reach Australia.
  * The concept of the alternate silo is that players could use this, but because it is only a single belt, it should take longer to build and launch a rocket from that silo.  The benefit is that the players do not have to conquer the American continent.

## Outposts
The outposts of Crash Site will be used to provide basic resources only (e.g. no assembled items) - (coal, stone, iron ore, copper ore, uranium ore, oil, water, etc)
(If water is provided through an outpost, the offshore pump could be removed.  The benefit is neglibable, but the concept of tweaking another feature of the game is worth considering.)
* Initial outposts - those within the base - would be limited to level 1 and not upgradable.  If possible, they could be less than level 1.  

## Map Layout Ideas
* MLI1: The starting base is the middle of Africa - South Africa is a 'dead end', but would serve well for additional resource outposts that would get the player going in the early-to-mid game.  Their upgradability should be limited (e.g. level 4).
[...]

PR #982, opened 2019-09-14
Added adjustable Respawn Time to CrashSite (#957)

HySpeed:
"Respawn timer shouldn't be as long or maybe timer starts off ~small and increases~ per death or maybe based off evolution"
- Implementation is based off evolution
- I chose to not cache global.config

Based on conversation with GrilledHam:
- Respawn delay is to prevent / punish early game suicide runs to clear nests.

This implementation decreases the respawn time from default of 60 seconds to a minimum of 10 seconds, by 5 seconds per 10% of evolution.

Added enable (default false) to control adjustable respawn delay.  When disabled, respawn is 60 seconds.

### DIFF
--- a/config.lua
+++ b/config.lua
@@ -206,14 +206,16 @@ global.config = {
                 }
             }
         },
-        -- settings for controlling length of delay before player respawn
+        -- settings for controlling length of delay before player respawn - the delay goes *down* over time
         -- respawn time is determined by biter progression:
-        --   min_time + (increment_amount * biter_progression )
+        --   respawn_delay = max_time - (decrement_amount * biter_progression)
+        --   respawn_time = max( min_time, respawn_delay )
         -- min_time default is 10 seconds
-        -- increment_amount default of 3000 will add ~5 seconds per 10% evolution and max at 60 seconds
+        -- decrement_amount default of 3000 will remove ~5 seconds per 10% evolution
         player_respawn_time = {
           min_time = 600,
-          increment_amount = 3000
+          max_time = 3600,
+          decrement_amount = 3000
         },
     -- spawns more units when one dies
     hail_hydra = {
--- a/config.lua
+++ b/config.lua
@@ -213,6 +213,7 @@ global.config = {
         -- min_time default is 10 seconds
         -- decrement_amount default of 3000 will remove ~5 seconds per 10% evolution
         player_respawn_time = {
+          enabled = false,
           min_time = 600,
           max_time = 3600,
           decrement_amount = 3000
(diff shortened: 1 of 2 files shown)


grilledham:
I personally don't think it makes sense to reduce the respawn time based on evolution.
Evolution is used to make the game harder and here we are using it to arguably make the game easier.
Personally I think that confuses the purpose of evolution.

Zahir-Khan:
Gameplay-wise I feel the current value of 1 min does become too much sometimes as you need to almost always travel far to recover the body but it's worth the wait and it forces the person to be careful. I would prefer providing minutes to players so they can choose to use those minutes to reduce their respawn times. Keep a floor respawn time of 10s. Minutes could be earned by staying online and of course, in the future, we could provide unlimited minutes to the contributors.

grilledham:
I think adding an upgrade to the spawn market that reduces spawn time for the whole team would be my preferred way to more forward with this.

I'm open to more discussion / ideas though.

HySpeed:
I'm good with making it a purchase. 
The bonus for me is that I would learn how to add an item to the market, which I want to learn as part of #977 .

What numbers are you thinking would be good?  We need:

- Initial Price (coins)
-  Increase in price per purchase (coins)
- Max level (market level)
- Change per purchase (seconds)

I'm thinking we keep these values

- Minimum time (currently 10 seconds)
- Initial [maximum) time (currently 60 seconds)
- Enable / Disable

Which of the above values would be stored in the Root Config File? 
Is there a Market Configuration File? (I haven't looked.)

One of the earlier comments was a penalty for quick deaths.  (e.g. If last respawn was less than X minutes ago, respawn_delay is max_respawn_delay.)  I'd like to add this, but I'd need to know if the Factorio or RedMew code keeps track of time since last spawn for players.  I wouldn't want to add a table to track that for just this rarely-used feature.

thanks

linaori:
You can fetch the market group a market belongs to and then use `Retailer.set_item` to add it. To update the price you can either completely replace the item, or look it up with `Retailer.get_items` and change the price. Note that the latter doesn't trigger a GUI refresh, so you're safer replacing the item (use a copy, not a reference to the same table). 

**Edit** Additionally you can also make it purchasable once, and add an event listener to add the new one. You can also add all of the upgrades from the start and only enable them when the previous is bought.

Jayefuu:
@HySpeed are you still working on this? It looks like a really good feature but has been quiet for a while.

HySpeed:
> @HySpeed are you still working on this? It looks like a really good feature but has been quiet for a while.

@Jayefuu I did not work on it any farther.  It was a bit too high of a technical hurdle for me to figure out.
Anyone else is welcome to continue and implement it.

Issue #552, opened 2018-12-15
SimpleRoomGenerator water in Diggy corrupts stress levels


linaori:
**Describe the bug**
Opening a room and placing water will not properly destroy rocks, causing the stress map to not be updated, meaning the room collapse prevention system won't spawn a new rock

**Expected behavior**
A rock should be spawned either way

**Steps to reproduce**
Open a room that would've collapsed

**Screenshots**
Dirt is working as intended: 
![image](https://user-images.githubusercontent.com/1754678/50043471-0d8c8f80-0075-11e9-8da4-4c7b0c0c5bf0.png)

Water is not working as intended:
![image](https://user-images.githubusercontent.com/1754678/50043473-1aa97e80-0075-11e9-81a2-79cee17b8ca7.png)

**Map**
Diggy - latest

**Additional context**
n/a

linaori:
Here's a clear example where something went wrong because a rock was removed as it should've collapsed, but was prevented: 
![image](https://user-images.githubusercontent.com/1754678/50043517-c521a180-0075-11e9-80b0-94e1a8402a02.png)

PR #983, opened 2019-09-14
Fix stress corruption caused by water #552

curiosity-a:


map_gen/maps/diggy/template.lua
```diff
@@ -41,28 +42,48 @@ local function insert_next_tiles(data)
...
                 end
             end
+            if string_find(new_tile.name, "water", 1, true) then --maybe check prototype's collision mask instead?
+                local entities = surface.find_entities{{tile_pos.x, tile_pos.y}, {tile_pos.x + 1, tile_pos.y + 1}}
+                for _, entity in pairs(entities) do
+                    if entity.teleport({entity.position.x + teleport_offset, entity.position.y}) then
+                        teleported_count = teleported_count + 1
+                        teleported[teleported_count] = entity
+                    else
+                        print("WARNING: entity " .. entity.name .. " failed to teleport out of the way at " .. serpent.line(entity.position))
```
grilledham: What does `print` refer to? If it's `G.print` then you probably didn't want that. Maybe use `log` instead.

map_gen/maps/diggy/template.lua
```diff
@@ -41,28 +42,48 @@ local function insert_next_tiles(data)
...
             if new_tile.name ~= 'out-of-map' then
-                local current_tile = get_tile(new_tile.position.x, new_tile.position.y)
+                local current_tile = get_tile(tile_pos.x, tile_pos.y)
                 if current_tile.name == 'out-of-map' then
                     void_removed_count = void_removed_count + 1
                     void_removed[void_removed_count] = {surface = surface, position = current_tile.position}
                 end
             end
+            if string_find(new_tile.name, "water", 1, true) then --maybe check prototype's collision mask instead?
+                local entities = surface.find_entities{{tile_pos.x, tile_pos.y}, {tile_pos.x + 1, tile_pos.y + 1}}
```
grilledham: To be consistent with the other surface calls, we could make a local `find_entities`

map_gen/maps/diggy/template.lua
```diff
@@ -41,28 +42,48 @@ local function insert_next_tiles(data)
...
+                for _, entity in pairs(entities) do
+                    if entity.teleport({entity.position.x + teleport_offset, entity.position.y}) then
```
grilledham: I think you can just call this as `entity.teleport(teleport_offset)`
Then do the same thing below.

Saves making a table, but arguably your way is clearer.

map_gen/maps/diggy/template.lua
```diff
@@ -41,28 +42,48 @@ local function insert_next_tiles(data)
...
-                local current_tile = get_tile(new_tile.position.x, new_tile.position.y)
+                local current_tile = get_tile(tile_pos.x, tile_pos.y)
                 if current_tile.name == 'out-of-map' then
                     void_removed_count = void_removed_count + 1
                     void_removed[void_removed_count] = {surface = surface, position = current_tile.position}
                 end
             end
+            if string_find(new_tile.name, "water", 1, true) then --maybe check prototype's collision mask instead?
+                local entities = surface.find_entities{{tile_pos.x, tile_pos.y}, {tile_pos.x + 1, tile_pos.y + 1}}
+                for _, entity in pairs(entities) do
```
grilledham: For new code we seem to have adopted the style of using `for i = 1, #entities` over using `pairs`.

grilledham (approved):
I think this is a decent solution to the bug. 
Someone who is more experienced with the diggy code might be able to give a more meaningful review though.

linaori (approved):

grilledham:
I'm merging now, thanks for the PR.
Closes 552

PR #985, opened 2019-09-29
Feature - Combat bots spawn when player kills a turret.

Zahir-Khan:


grilledham:
Thanks for the PR.

Can I ask why this uses the `on_entity_damaged` event rather than the `on_entity_died` event? The former is more expensive.

map_gen/maps/crash_site/entity_damaged_events.lua
```diff
@@ -0,0 +1,72 @@
...
+local bot_spawn_whitelist = {
+    ['gun-turret'] = true,
+    ['laser-turret'] = true,
+    ['flamethrower-turret'] = true
+}
+
+local spawn_bot_def =
+    Token.register(
+        function(parm)
+            local entity = parm.event.entity
```
grilledham: Need to check the entity is still valid before doing anything with the entity, else this can throw an error.
```lua
if not entity.valid then
    return
end
```
Is how we normally do it.
You also need to do this for `spawn_bot_des` and for `on_entity_damaged` event.

map_gen/maps/crash_site/entity_damaged_events.lua
```diff
@@ -0,0 +1,72 @@
...
+        local entity_force = entity.force
+        local repeat_cycle = 1 -- the number of times a squad of robots are spawned default must be 1
+        local ef = game.forces.enemy.evolution_factor
+         -- remove 'character' condition so that attack form any entity such as artillery will cause the bots to spawn.
+        if bot_spawn_whitelist[entity_name] and entity.force.name == 'enemy' and cause.name == 'character' and entity.health == 0   then --Conditions are prioritised for performance
+            if ef > .95 then
+                repeat_cycle = 2
+            end
+            if entity.name == 'gun-turret' then
+                local parm = {event = event}
```
grilledham: Could you pass the event table as is? Saves having to make an extra table and then getting the event back out again later.

map_gen/maps/crash_site/entity_damaged_events.lua
```diff
@@ -0,0 +1,72 @@
...
+                for i = 1, repeat_cycle, 1 do
+                    --Wonky Code because set_timeout_in_ticks doest work beyond 10 ticks for some reason and doesnt work if same funciton is called multiple times in the same tick.
```
grilledham: I'm not sure what is going on here with this comment. Have you tested with `_DEBUG = true` in map_selection.lua to see if something is throwing an error.

grilledham:
I haven't play tested this yet, so I'm not sure what the difficulty is like, but we could consider limiting this to mid - late game turrets.

Zahir-Khan:
> Thanks for the PR.
> 
> Can I ask why this uses the `on_entity_damaged` event rather than the `on_entity_died` event? The former is more expensive.

Because on_entity_died event for turrets doesn't provide 'cause' which is necessary to know which player killed the turret. See the forum post below. More info on the discord dev channel.

[https://forums.factorio.com/viewtopic.php?f=23&t=48353&p=419260&hilit=on_entity_died+cause#p419260]

I have a feeling both the on_entity_damaged and on_entity_died event work the same internally - regardless of which is more expensive. I have optimized the conditions so that one condition check occurs most of the time. In the grand scheme of things, this check occurs very rarely throughout the game.

Zahir-Khan:
> I haven't play tested this yet, so I'm not sure what the difficulty is like, but we could consider limiting this to mid - late game turrets.

This is currently being tested on server 9 - Crashsite beta. Haven't noticed any problems so far in early game or late in my testing, they don't do more damage than a few biters and spitters chasing a player throughout the evolution cycle. There are multiple factors I took into account before settling in on a modifier too many to summarise in text. Need to schedule a discord voice with playtesting there is definitely room for tweaking. The mid-late game implementation is my intention too. I have kept the early game easy so that players don't feel a difference in difficulty from before.

map_gen/maps/crash_site/entity_damaged_events.lua
Zahir-Khan: I noticed, I forgot to add it. It has been added to on_entity_damaged but should I add it to spawn_bot_des/def too? since its the same entity info being passed along. It would be checking twice.

map_gen/maps/crash_site/entity_damaged_events.lua
Zahir-Khan: It is a legacy comment was having issues with bots, not spawning/spawning sometimes if I set the tick value more than 10 ticks could be a problem with that entity not being valid anymore as it gets destroyed. Doesn't matter now as the current implementation works flawlessly.

map_gen/maps/crash_site/entity_damaged_events.lua
grilledham: You have to add it here too.
The reason is the entity could become invalid in the ticks from when you last checked.

map_gen/maps/crash_site/entity_damaged_events.lua
```diff
@@ -0,0 +1,74 @@
...
+    defines.events.on_entity_damaged,
+    function(event)
+        local entity = event.entity
+        local entity_name = entity.name
+        local entity_pos = entity.position
+        local cause = event.cause
+        local entity_force = entity.force
+        local repeat_cycle = 1 -- the number of times a squad of robots are spawned default must be 1
+        local ef = game.forces.enemy.evolution_factor
+        if not entity.valid then
```
grilledham: You need to do the `entity.valid` check before you access any properties of the entity.

grilledham:
> Because on_entity_died event for turrets doesn't provide 'cause' which is necessary to know which player killed the turret. See the forum post below. More info on the discord dev channel.

Well that does make it more difficult. 
I'm still not sure about using entity_damage event though.
Maybe we can look for nearby players on entity died and send the bots towards them?

map_gen/maps/crash_site/entity_damaged_events.lua
Zahir-Khan: I havnt understood what you mean.

Zahir-Khan:
> > Because on_entity_died event for turrets doesn't provide 'cause' which is necessary to know which player killed the turret. See the forum post below. More info on the discord dev channel.
> 
> Well that does make it more difficult.
> I'm still not sure about using entity_damage event though.
> Maybe we can look for nearby players on entity died and send the bots towards them?

Yes I understand, we are using the surface.find_entities_filtered to determine artillery targets. This method is highly expensive with the frequency at which bot spawning must occur - we are defining a large area (at least 36sq rocket launcher range) in which the game checks if each and every entity within this area matches the filter we set then it must decide which one of the character entities it must target then get its position and not only the position we need a reference to this entity itself to spawn the bots they cant spawn without a target entity. I have noticed heavy lagg on crash site server when just a few players move around within the enemy arty range.

The way I have ordered the checks for entity_damaged_event, most of the time a check occurs only once - every time an entity gets damaged the first check is with respect to health == 0, this check is the baseline and for any new feature added there is always a cost. The second check is if its enemy, so here all player entities are filtered out. The Third check is the whitelist (enemy doesn't have a huge number of entities so by this time its almost as if the function within is being run). 
I'm using a series of 'AND' interpreter will not check all of them it checks one by one so most of the time the cost is 1-2 checks each time something gets damaged at most its 3.

Zahir-Khan:
> > Because on_entity_died event for turrets doesn't provide 'cause' which is necessary to know which player killed the turret. See the forum post below. More info on the discord dev channel.
> 
> Well that does make it more difficult.
> I'm still not sure about using entity_damage event though.
> Maybe we can look for nearby players on entity died and send the bots towards them?

Yes I understand, we are using the surface.find_entities_filtered to determine artillery targets. This method is highly expensive with the frequency at which bot spawning must occur - we are defining a large area (at least 36sq rocket launcher range) in which the game checks if each and every entity within this area matches the filter we set then it must decide which one of the character entities it must target then get its position and not only the position we need a reference to this entity itself to spawn the bots they cant spawn without a target entity. I have noticed heavy lagg on crash site server when just a few players move around within the enemy arty range.

The way I have ordered the checks for entity_damaged_event, most of the time a check occurs only once - every time an entity gets damaged the first check is if the entity is enemy,  here all player entities are filtered out. The second check is health ==0. The Third check is the whitelist (enemy doesn't have a huge number of entities so by this time its almost as if the function within is being run).
I'm using a series of 'AND' interpreter will not check all of them it checks one by one before moving to another so most of the time the cost is 1 check each time something gets damaged at most its 2.

map_gen/maps/crash_site/entity_damaged_events.lua
```diff
@@ -0,0 +1,81 @@
...
+        end
+        local entity_name = entity.name
+        local entity_pos = entity.position
+        local cause = event.cause
+        local entity_force = entity.force
+        local repeat_cycle = 1 -- the number of times a squad of robots are spawned default must be 1
+        local ef = game.forces.enemy.evolution_factor
+         -- remove 'character' condition so that attack form any entity such as artillery will cause the bots to spawn.
+         -- Conditions are prioritised for performance do not change the order for first 3
+        if entity.force.name == 'enemy' and entity.health == 0 and bot_spawn_whitelist[entity_name] and cause.name == 'character' then
```
linaori: Can you try to reduce the amount of assignments to a bare minimum? Right now if this `if` is hit, there are a bunch of assignments that are unused, of which some are expensive.

`entity.force.name` should be `entity_force.name`

map_gen/maps/crash_site/entity_damaged_events.lua
```diff
@@ -0,0 +1,81 @@
...
+            if ef > .95 then
+                repeat_cycle = 2
+            end
+            if entity.name == 'gun-turret' then
```
linaori: `entity.name` should be `entity_name`, same for `entity.name == 'laser-turret'` below

map_gen/maps/crash_site/entity_damaged_events.lua
```diff
@@ -0,0 +1,81 @@
...
+            elseif entity.name == 'laser-turret' then
+                local parm = {event = event}
+                for i = 1, repeat_cycle, 1 do
+                    entity.surface.create_entity{name = 'defender', position = entity_pos,target= cause,force = entity_force}
+                    Task.set_timeout_in_ticks(4,spawn_bot_des,parm)
+                    Task.set_timeout_in_ticks(6,spawn_bot_def,parm)
+                end
+            else
+                for i = 1, repeat_cycle, 1 do
+                    entity.surface.create_entity{name = "distractor-capsule", position = entity_pos,target= event.cause,force = entity.force,speed = 0}
```
linaori: `event.cause` should be `cause` and `entity.force` should be `entity_force`

map_gen/maps/crash_site/entity_damaged_events.lua
linaori: I think he means `Task.set_timeout_in_ticks(4,spawn_bot_des, event)`

map_gen/maps/crash_site/entity_damaged_events.lua
```diff
@@ -0,0 +1,81 @@
...
+                local parm = {event = event}
+                for i = 1, repeat_cycle, 1 do
+                    entity.surface.create_entity{name = 'defender', position = entity_pos,target= cause,force = entity_force}
+                    Task.set_timeout_in_ticks(4,spawn_bot_des,parm)
+                    Task.set_timeout_in_ticks(6,spawn_bot_def,parm)
+                end
+            elseif entity.name == 'laser-turret' then
+                local parm = {event = event}
+                for i = 1, repeat_cycle, 1 do
+                    entity.surface.create_entity{name = 'defender', position = entity_pos,target= cause,force = entity_force}
```
linaori: Can you ensure that for all cases within loops, the create_entity is cached? This saves several expensive index lookups.
```lua
local create_entity = entity.surface.create_entity
for ... then
    create_entity{...}
end
```

linaori (changes requested):
Can you please make sure that there's a single space after commas and spaces around the equal signs? I'm having difficulties reading the code like this as it's rather inconsistent. 
```diff
-create_entity{name = 'defender', position = entity_pos,target= cause,force = entity_force}
+create_entity{name = 'defender', position = entity_pos, target = cause, force = entity_force}
```
I'm having issues reading

map_gen/maps/crash_site/entity_damaged_events.lua
Zahir-Khan: Thank you, im unaware of such intricacies both readability and performance is fixed. some of these things crept in with me trying out what works and doesn't.

grilledham:
I'm still worried about using the entity_damaged event.

I did some experimenting. I wrote some code that counts how many times the on_entity_damged / died event are called. Then used a flamthrower to destroy one gun turret in one of the mini outposts. The ratio is about 5000 : 1.
What causes most of the events appears to be the walls, even though they don't take damage from fire, they still raise the event.

You may be right that doing it the way I suggested with a surface find nearby entities might be more expensive. I actually don't know I haven't compared results. I not even sure how we could measure that in a meaningful way.

But I'm still worried about the potential performance impact of this approach. So I'm not really sure what to suggest.

Maybe instead of being on turret killed, it's on outpost captured, and we spawn a bunch of bots at a time. 

> I have noticed heavy lagg on crash site server when just a few players move around within the enemy arty range.

Also if you have evidence of that, it might be worth investigating. Maybe we can reduce the frequency of the searches, or schedule them so multiple searches don't happen on the same tick.

Zahir-Khan:
> I'm still worried about using the entity_damaged event.
> 
> I did some experimenting. I wrote some code that counts how many times the on_entity_damged / died event are called. Then used a flamthrower to destroy one gun turret in one of the mini outposts. The ratio is about 5000 : 1.
> What causes most of the events appears to be the walls, even though they don't take damage from fire, they still raise the event.
> 
> You may be right that doing it the way I suggested with a surface find nearby entities might be more expensive. I actually don't know I haven't compared results. I not even sure how we could measure that in a meaningful way.
> 
> But I'm still worried about the potential performance impact of this approach. So I'm not really sure what to suggest.
> 
> Maybe instead of being on turret killed, it's on outpost captured, and we spawn a bunch of bots at a time.
> 
> > I have noticed heavy lagg on crash site server when just a few players move around within the enemy arty range.
> 
> Also if you have evidence of that, it might be worth investigating. Maybe we can reduce the frequency of the searches, or schedule them so multiple searches don't happen on the same tick.

I'm all in for the performance perspective, we shouldn't include anything that may cause intermittent lag spikes. 

1. 5000:1 ratio is huge. I want to understand how was this metric calculated. Can you share the code?

2. I was afraid walls would be the biggest performance sink, didn't know they would still raise this event if they weren't damaged. Thinking a bit further with all the attacks happening simultaneously across the map you are right not a good idea to use it. If this event occurred once per tick per entity then it's being raised per tick  (no. of entities damaged x lookup x conditions checked) It does amount a lot more than what I was naively thinking. 

3. It is difficult to show you evidence for the performance issues during players walking about in the enemy arty area. It's something that was memorable for me I was new to modded factorio.  All I can say is this every time when few people(4-6) have braved to keep on fighting in the arty range the whole chat lights up about the lagg and someone once said not to keep loitering on the edge of the target boundary as the code keeps checking who to target. This happened once or twice when I first started playing crash site many months ago.

4. Wouldnt recommend spawning bots at the time when an outpost is captured - a lot of things are going on when the outpost is captured - market upgrade enable, toast, walls changing force already experience lagg during end game when outposts are captured - I play on a laptop so when i say lagg its probably not the server.

Options - Bots can be spawned relatively easy the problem is, a character entity reference is needed if they need to follow the player. If not would be just another static turret.
a. We could ask game makers to provide cause for turrets - on_entity_died event.
b. Is there a way to trigger one event form another? We could trigger the on_entity_damaged_event from on_entity_died_event all we need is the cause - it's late in the night don't judge me.

c. Probably our only option - Move this code to on_entity_died. Use surface.find_entities_filtered - But I think this feature will lose predictability because we would have to assign the bots randomly to a character an innocent bystander or a guy who is struggling to get away then suddenly out of nowhere bots chase and kill... Then he rage quits and makes profane servers. JK.

d. hehe we are back to the beginning - debate the performance impact of this feature. At this point, I say we either come up with some useful implementation or drop the feature. If you have a way to monitor/log redmew server performance I have a way to test how much the impact will be - load a save of crash site midgame log the base performance over a certain time. Now replace the files in this save file with bot feature and run the same thing again. Compare. (I have playtested this on redmew s9 and the file iv used is way more unoptimized than we have collaborated here didn't notice any performance issues even with damaging a whole bunch of walls and turrets with nades etc. It's still on s9 it's stopped though due to lack of player participation)

grilledham:
The code I used to test the event counts is quite simple.
```lua

global.event_count = {
    died = 0,
    damaged = 0
}

Event.add(
    defines.events.on_entity_died,
    function()
        global.event_count.died = global.event_count.died + 1
    end
)

Event.add(
    defines.events.on_entity_damaged,
    function()
        global.event_count.damaged = global.event_count.damaged + 1
    end
)
```

If you have `_DEBUG = true` in map_selection.lua, you can run /debug and inspect the value easily.
image

There is potentially a good way we could measure the performance of this feature.
If you run factorio from the command line you can benchmark a save for n ticks.
Something like 
`factorio.exe --benchmark FILE`
You can run `factorio.exe --help` to see all the options

Zahir-Khan:
I have made a request on the forum. If it doesnt get fulfilled would have to compromise on the functionality and target random players nearby (not a big fan of) or run the on_entity_damaged event for a few ticks after a turret dies which would make the spawn rate random.
https://forums.factorio.com/viewtopic.php?f=28&t=76653

Zahir-Khan:
The latest changes are working. May need balancing. For now bots spawn after 20% Evo and 2 squads spawn after 95% Evo.
One Squad
Gun Turrets: 2 defender 1 destroyer
Laser: 1 defender 2 destroyer
Flame: 3 Static distractor

grilledham:
The on_entity_died event was getting a bit long, so I refactored it into smaller functions.

I did a quick test and this seems good.

Zahir-Khan:
This works well.

Also, I was thinking to spawn about 24 bots (mix of defender and destroyer) for each enemy arty death. Right now Enemy arty is remotely destroyed by players without any repercussions. 
Destroying 4 arty (which is typically done) will spawn 100 bots that will race across the map towards the arty that destroyed them. In my testing (Max research), I used 2 layer gun turrets these bots will kill around 4 gun turrets. Take half health of the player arty. Will kill any player that is in the path of the bots. Will damage anything that is not protected.

This should be a nice late-game addition. I am not sure of the server performance impact.

grilledham:
I think we should merge this as is. Then if you want to spawn bots on artillery turrets death you can do that as a separate pull request.

Zahir-Khan:
Ok ready for merging.

PR #991, opened 2019-10-04
Concrete Jungle

SimonFlapse:
Suggested by Sangria Louie (https://discordapp.com/channels/432567222481846283/432570869039104012/629791144846426122)

Concrete Jungle is a new map where most entities are restricted from being placed on the ground unless the ground is covered in stone path, concrete or better.

Includes a minor change to the restriction module to fix upvalues in the callback and additions to it to prevent losing items from entities with inventories when they are upgraded (fast-replaced).
For chests only logistic-chests can be a different tier than the three standard chests. For all other entities with inventory it should spill the content to the ground

Spawn has a small circle of stone path to get started, and also a 3x3 concrete square to allow for the first assembly machine 2 to be placed, it's needed to produce more concrete.

You can only upgrade tiles, no downgrading or removing. (You can convert between hazard and regular concrete)

Also includes a change to the welcome message (Because I'm too lazy to create a new PR for it)


SimonFlapse:
Tier 0 (Ground):
image

Tier 1 (Stone-path):
Everything else

Tier 2 (Concrete or Hazard-concrete):
image
**iron-chest has been moved to tier 1**

Tier 3 (Refined-Concrete or Refined-Hazard-Concrete):
image
**steel-chest has been moved to tier 1**

SimonFlapse:
Feedback from discord by curiosity:

> I think, tiered assemblers will be interesting.
Maybe belts, furnaces.
If you want to be weird, miners :smile:
As for optimization, add all higher tier tiles to lower tiers, then you will only have to check one table.
Or do a single table with numbers instead of true. Would also make the tier system easily expandable.
And you will be able to have a single-condition check: (tier[name] or 0) <= (tile_tier[tile_name] or 0)
(of course, you'd want to cache most of that)

Source: https://discordapp.com/channels/432567222481846283/432570869039104012/629820033790509067

map_gen/maps/concrete_jungle.lua
```diff
@@ -0,0 +1,208 @@
...
+            name = entity.ghost_name
+        end
+
+        for x = floor(left_top.x), floor(right_bottom.x) do
+            for y = floor(left_top.y), floor(right_bottom.y) do
+                local tile_name = get_tile(x, y).name
+                local tier = tile_tiers[tile_name]
+
+                if not((entity_tiers[name] or 1) <= (tier or 0)) then
+                    return false
```
grilledham: This might read slightly better as:
```lua
local entity_tier = entity_tiers[name] or 1

for x = floor(left_top.x), floor(right_bottom.x) do
    for y = floor(left_top.y), floor(right_bottom.y) do
        local tile_name = get_tile(x, y).name
        local tile_tier = tile_tiers[tile_name] or 0

        if entity_tier > tile_tier then
            return false
```

map_gen/maps/concrete_jungle.lua
```diff
@@ -0,0 +1,208 @@
...
+end
+
+Event.add(RestrictEntities.events.on_pre_restricted_entity_destroyed, on_destroy)
+
+--Creating the starting circle
+local shape = b.circle(50)
+local stone_path = b.tile('stone-path')
+
+shape = b.invert(shape)
+local map = b.if_else(shape, stone_path)
```
grilledham: Just a suggestion, but this could be done like so:
```lua
local circle= b.circle(50)
local stone_circle = b.change_tile(shape, true, 'stone-path')

local map = b.if_else(stone_circle, b.full_shape)
```

SimonFlapse: Thanks, I was looking for that, but I just found another method of doing it

grilledham:
Overall this looks pretty good and I think it will be fun to try. I like the entity tiers idea.

Maybe I'm more cruel, but I'd be tempted to only have a 2x2 stone patch at spawn and require all entities to be at least tier 1.

SimonFlapse:
Should we implement some floating text for the restricted entities like they've done in 0.17:

image

grilledham:
If you can get that working, it seems like a nice addition.

SimonFlapse:
I'm having troubles with a lot of upvalues in the entity_placement_restricted module

grilledham (approved):

grilledham:
I only saw The 'Minor improvements' commit before I approved.
I'm still thinking about the 'Fixed upvalues' changes

SimonFlapse:
Should be good to go after some testing 🤞

grilledham:
Thinking about it I don't think tokenising the function for `set_keep_alive_callback` should be needed. We only need to do that if we want to store the function in global.

Can you provide a save where you think the problem was happen so I can debug?

R-Nukem:
Looking through the danger ores stuff in relation to a message on discord:
>SimonFlapse: TheOrangeAngle# We need to test all the danger ores maps again with #991. Because I had to change how the restrict entity module work

Everything seems fine. I copied the modified files over to the Omnimatter danger ores I've been working on and there is no change. Working as expected.

SimonFlapse:
> Looking through the danger ores stuff in relation to a message on discord:
> 
> > SimonFlapse: TheOrangeAngle# We need to test all the danger ores maps again with #991. Because I had to change how the restrict entity module work
> 
> Everything seems fine. I copied the modified files over to the Omnimatter danger ores I've been working on and there is no change. Working as expected.

Sound great, I've changed a bit more because of chests losing their content if upgraded without the necessary ground support. Now entities should spill their content to the ground before being destroyed

SimonFlapse:
**First playthrough (2019-10-05) feedback:**
- Content of chests can become lost when upgrading (Fixed in: https://github.com/Refactorio/RedMew/pull/991/commits/1f34779c8e5199b6ab670a8c8be7bd350057b513)
- Impossible to advance to concrete production since it requires Assembling Machine 2 which in turn requires concrete (Fixed in: https://github.com/Refactorio/RedMew/pull/991/commits/1f34779c8e5199b6ab670a8c8be7bd350057b513)

SimonFlapse:
**Minor typo**
I found that the welcome message tells the players to look for the question mark in the top left corner, but we changed it to an 'i'. I've included a fix for that here because I was lazy and didn't want to make a new PR.

**Refactoring**
I've refactored most of the additions to the restrict entity placement so that the overall logic should be the same for the danger ores maps. 

**Locales**
I've converted all strings from the Concrete Jungle to localised strings

map_gen/maps/concrete_jungle.lua
```diff
@@ -0,0 +1,490 @@
...
+
+--- Anti griefing
+
+local anti_grief_tries = 3 -- Change this number to change how many 'shots' a player has got
+local forgiveness_time = 432000 -- Time before previous offences are forgiven, in ticks (60 ticks/sec)
+local anti_grief_kick_tries = math.ceil(anti_grief_tries / 2) -- Tries before kick warning
+
+RestrictEntities.set_anti_grief_callback(
+    Token.register(
+        function(_, player)
```
grilledham, line 270: At some point in this function we should check the player's rank. We typically don't perform anti grief measures on auto-trusted or above ranks. Probably after the popup.

grilledham:
The spill behaviour for the entity restriction stuff is a hard one. In an idea world we would undo the action or prevent it from happening in the first place, so there would be no need to spill the items.
But that doesn't seem trivial to do.

SimonFlapse:
> The spill behaviour for the entity restriction stuff is a hard one. In an idea world we would undo the action or prevent it from happening in the first place, so there would be no need to spill the items.
> But that doesn't seem trivial to do.

I think it's worst for chests, the solution could be to set all chests to the same tier (Logistic chests could still be tiered higher, but replace with a steel chest)

Jayefuu:
This concept seems like an improved version of something Plague did last year. I can't for the life of me remember what he called it, nor can I find it. Perhaps he didn't commit it?

If it is found, it might be worth checking if there are any features or concepts worth carrying over.

EDIT: I remember now. In his the path/concrete expanded based on pollution and you could only build on parts you had produced enough pollution to use. I like this version better because it encourages more production of tiles and hence more factory buildings.

grilledham:
In plague's version you would take damage if you left the terraformed areas. Maybe we could do something similar?

SimonFlapse:
> > The spill behaviour for the entity restriction stuff is a hard one. In an idea world we would undo the action or prevent it from happening in the first place, so there would be no need to spill the items.
> > But that doesn't seem trivial to do.
> 
> I think it's worst for chests, the solution could be to set all chests to the same tier (Logistic chests could still be tiered higher, but replace with a steel chest)

In https://github.com/Refactorio/RedMew/pull/991/commits/a9836b31f2dca49896f4e0f80a2c0edfd8ba9122 I've moved all regular chests (`entity.type == 'containers'`) to tier 1.

I'm now fast replacing logistic containers with a steel-chest, preserving the chest content. To compensate for the lost logistic container I'm refunding the circuits required to make it.

SimonFlapse:
The new additions, since the first run, needs to be tested, after which it should be ready to merge

grilledham:
I think this is probably fine, but I still want to review this. Should be able to review at the weekend.

SimonFlapse:
@grilledham in case you forgot, you wanted to do another review.

We might also want to play test it again, before merging it @theorangeangle

map_gen/maps/concrete_jungle.lua
```diff
@@ -0,0 +1,496 @@
...
+            Popup.player(player, [[
+Look out!
+
+You are trying to replace entities which have items inside!
+This is causing the items to spill out on the ground.
+
+Please be careful!
+]], nil, nil, 'entity_placement_restriction_inventory_warning')
+
+            if not (Rank.equal_or_greater_than(player.name, Ranks.regular)) then
```
grilledham: In my opinion it should be the auto trusted rank, rather than regular. I think that would be more consistent with how we normally do this.

map_gen/shared/entity_placement_restriction.lua
```diff
@@ -109,6 +174,9 @@ local on_built_token =
 
         local name = entity.name
         if name == 'tile-ghost' then
+            if entity.ghost_name ~= 'landfill' and primitives.prevent_tile_bp and entity.valid then
```
grilledham: I don't think the `entity.valid` is needed here, we checked that a few lines up.

grilledham:
Looking at the entity_placement_restriction module, I can't help feel that it has become a mess and it would be easier to understand the code if we just didn't use it. 

I don't think it's your fault, I think the way it was implemented might have been flawed to begin with.

That said, the burden of rewriting it doesn't fall on you. So I might just approve this as is, with the intention that one day I'll rewrite the entity_placement_restriction module.

SimonFlapse:
Todo after second playthrough:
- [x] Blueprint books cause multiple flying texts (Unlike regular blueprints)
- [x] Entity ghost flying text doesn't display the entity sprite
- [x] Trains should always be allowed. Their bounding box makes it hard to place on rails
- [x] Typo: Security Measures 
- [ ] Upgrade planner might not play nicely with entities with inventories
- [x] Turrets with autofill cause players to get strikes for spilling items.

Other:
- [ ] Hard code a tag used for afk, on join message: {player_name} will be back soon | on leave message: {player_name} is with us again

theorangeangle:
Will be doing a beta test on tuesday 3/31 to see if there is additional feedback to be considered before merging @SimonFlapse

theorangeangle:
@SimonFlapse ^

PR #993, opened 2019-10-07
[WIP] Space Race

SimonFlapse:
A new map concept called **Space Race**

This is an Active/Passive PVP map where you'll have to race to launch a rocket, or conquer the other teams launch pad before they do.

You can choose to stay safe within your own zone, or go out and scout for oil, fight for uranium or just sabotage the other team for your own leisure.

This map shows the overall map layout along with some key informations.
image

I'd love to get some feedback.


SimonFlapse:
Changed die force in `Turret Active Delay` to `neutral`

https://github.com/Refactorio/RedMew/pull/993/commits/f0e147f39ecb878b3b3898dfe12b6bb02d13629d fails `/toggle-heavy-mode`

https://github.com/Refactorio/RedMew/pull/993/commits/d6842d8e4a8b6e474279b032519232e79d491c09 fixes desyncs by causes the oil to be in the same spot every reset.
I'd need some help with using `utils.random` and seeds

SimonFlapse:
I will need to get your review on my changes to `generate.lua` @grilledham 
You should be able to add surfaces to the `surfaces` table in `generate.lua` until `_LIFECYCLE > 7` without desyncs. 

Because shapes are functions we can't really use `utils.global` to allow changes after `_LIFECYCLE = 8` and the work around would involve `utils.token` which again needs to be registered before `_LIFECYCLE = 8` so it's better to just keep the current structure and only allow `surfaces` to be changed before `_LIFECYCLE = 8`

Problem is the `Public.get_surfaces()` in `generate.lua` returns a reference to the `surfaces` table which allows it to be changed during runtime, so either we should return a copy of it after `_LIFECYCLE = 8` or just don't allow the function to be called after `_LIFECYCLE = 8`

SimonFlapse:
Example on how to add a new surface to the `generate.lua`

```lua
-- Needs to be before or on: on_init or on_load (Before _LIFECYCLE = 8)
local Generate = require 'map_gen.shared.generate'
local shape = b.rectangle(45, 4)
shape = b.change_tile(shape, true, 'concrete')
Generate.get_surfaces()['snake'] = shape
```

The surface can be created at anytime here after, with the surface name `'snake'`
The `generate.lua` will only start running the shape function when the `on_chunk_generated` is triggered for the surface (Which can't happen before after the surface has been created)

map_gen/shared/generate.lua
```diff
@@ -282,7 +282,9 @@ end
 function Public.init(args)
     tiles_per_tick = args.tiles_per_tick or 32
     regen_decoratives = args.regen_decoratives or false
-    surfaces = args.surfaces or {}
+    for surface_name, shape in pairs(args.surfaces) do
```
grilledham: args.surfaces could be nil
```suggestion
    for surface_name, shape in pairs(args.surfaces or {}) do
```

map_gen/shared/generate.lua
```diff
@@ -312,6 +314,7 @@ function Public.register()
 end
 
 --- Returns the surfaces that the generate functions will act on
+-- Warning! Changing this table after on_init or on_load has run will cause desyncs!
```
grilledham, line 317: Let's throw an error if someone calls this at runtime

map_gen/maps/space_race/cliff_generator.lua
```diff
@@ -0,0 +1,177 @@
...
+
+local Public = {}
+
+local orientation = {
+    ['none-to-west'] = 'east-to-none',
+    ['west-to-none'] = 'none-to-east',
+    ['east-to-none'] = 'none-to-west',
+    ['none-to-east'] = 'west-to-none',
+    ['north-to-west'] = 'east-to-north',
+    ['west-to-north'] = 'north-to-easy',
```
grilledham: typo?
```suggestion
    ['west-to-north'] = 'north-to-east',
```

control.lua
```diff
@@ -151,7 +151,7 @@ if config.redmew_settings.enabled then
     require 'features.gui.redmew_settings'
 end
 
-require 'features.snake.control'
+--require 'features.snake.control'
```
grilledham, line 154: Why comment out?

map_gen/maps/space_race/gui/join_gui.lua
```diff
@@ -0,0 +1,126 @@
...
+
+Gui.on_click(
+    join_USA,
+    function(event)
+        local frame = Gui.get_data(event.element)
+        local player = event.player
+
+        if remote.call('space-race', 'join_usa', nil, player) then
+            Gui.remove_data_recursively(frame)
+            frame.destroy()
```
grilledham: These two lines can be done in one
```lua
Gui.destroy(frame)
```

map_gen/maps/space_race/gui/join_gui.lua
```diff
@@ -0,0 +1,126 @@
...
+    Gui.set_data(join_ussr_button, frame)
+end
+
+Gui.on_click(
+    join_USA,
+    function(event)
+        local frame = Gui.get_data(event.element)
+        local player = event.player
+
+        if remote.call('space-race', 'join_usa', nil, player) then
```
grilledham: `remote.call` is for communicating between mods. Using it internally seems odd to me. Is there any advantage to using it?

control.lua
SimonFlapse: It caused a bug where the snake gui button was being rendered for some players, even though I never used snake.control

map_gen/maps/space_race/gui/join_gui.lua
SimonFlapse: Preventing circular dependencies, Lua isn't happy when you require a file that requires that same file. So if I were to require the space-race/scenario.lua file in here i'd get a circular dependency

grilledham: I was worried you would say that. My general advice is to avoid designing your modules in such a way that there is a circular dependency. But if that proves hard to do, you basically end up using a mechanism similar to `remote.call` anyway.

control.lua
grilledham: Strange, might be worth investigating.

map_gen/maps/space_race/gui/join_gui.lua
SimonFlapse: I have quite a way to go, so I will look into splitting it up into more modules

SimonFlapse:
Introducing King of the Hill!
image

The earthlings have launched a supply mission to help you get back from this alien planet! Unfortunately it also crashed down! You can now race the other team for spare parts for your rocket. Capture and hold all three outposts in the middle of the map to salvage parts! 

(Takes minimum 2.8 hours to collect enough free resources for 1000 rocket parts)

SimonFlapse:
Custom ores in two shapes to signify the safe and danger zones. Mirrored for an equal map:
image _Subject to weight balancing_

theorangeangle:
@SimonFlapse since you're going to have limited availability for the next month, I'll be working with the people from 3RA to accomplish some goals:

v0.5 
- Team Colors
- /info command to open map info gui
- /SRversion command to open extra info gui 
- Better external documentation, graphics

v0.6
-Cutscene to explain KOTH that airs when PvP phase begins 
-auto restart

grilledham:
With regards to the auto restart feature it might be good to get something merged to develop. I can put the scenario on the webpanel, then when we push to develop the scenario will auto update. Having a scenario on the web panel would be the easiest way to do auto restarts.

In order to merge this to develop there is no requirement that the scenario is finished. If we waited for scenarios to be finished before we merged them, then nothing would get merged. We do however, need to check changes to common code is correct and doesn't break anything else. In future it might be best to separate changes to common code into it's own PR, just something to keep in mind.

theorangeangle:
Sounds good, I will merge with develop after 0.5 provided nothing major is broken and then work off separate PRs for other things.

Zr4g0n:
**Suggestion to help reduce extreme snowballing of one team:**
Whenever a team unlockes a new research, the opposing team gets it automatically after X minutes. I suggest somewhere between 5 and 10 minutes. This allows the power-gap to always be there, but never be too wide. This helps give the 'loosing' team a better chance of fighting back, without giving any negatives to the 'winning' team. The best team still wins, but the other team might just bite back if the winning team starts relaxing or getting sloppy. 

**Admin only spectator mode**
Fir recording and observing games, an admin only spectator team would be great, and better than /editor and choosing the neutral force. Partially to avoid inflating one team's player-count, and partially to be able to see the battle 'at night' at night. With editor mode, it's daylight only by default. 

**Toggle player lights**
Being able to turn of a player's own lights would be useful in ambush situations. Right now, sneaking up on a player in the PVP area at night is very difficult because you're literally a beacon of lights in the darkness.

theorangeangle:
@Zr4g0n do you plan to work on any of this sometime soon? otherwise i'm gonna merge and then throw the missing items into the high priority section of the roadmap (along with all the other things we need devs for)

theorangeangle:
> 
> 
> @SimonFlapse since you're going to have limited availability for the next month, I'll be working with the people from 3RA to accomplish some goals:
> 
> v0.5
> 
>     * Team Colors
> 
>     * /info command to open map info gui
> 
>     * /SRversion command to open extra info gui
> 
>     * Better external documentation, graphics
> 
> 
> v0.6
> -Cutscene to explain KOTH that airs when PvP phase begins
> -auto restart

To get QOL up further, we should implement the following three (critical) items:

1. Team Colors - editable in config file with RGB for now, fancy gui later
2. Bug Fix for map generation - if a rock spawns where the silo is then it causes the map generation script to crash, stopping game start
3. Spectator mode - for streamers or admins, places character on a non-team force (maybe united factory management to keep with the theme?) that can see both sides, with no inventory or ability to influence things on the map surface. also allows for playing snake :P and should be an admin-only command

PR #994, opened 2019-10-09
[Builder] Simple entity removal

### LINKED ISSUE #992: [Builder] Add ability to remove trees, enemies, ores etc. from a shape
Requesting new functions in the builder that allows to remove trees, enemies, ores etc. from a given shape function, without affecting the entire map generation (Disabling trees in map_gen)

See #979's no_trees function for how it's done today.

SimonFlapse:
Simple entity removal on types or on names

Performance should be improved or function only used sparsely 
(Radius 2 is needed otherwise it will miss entities on the edges of chunks)

closes #992 

**Example**
_remove_entities_by_type_
```lua

local map = b.full_shape
local entity_types = {'tree', 'resource', 'unit-spawner', 'turret', 'unit', 'cliff', 'simple-entity'}

map = b.apply_effect(map, b.remove_entities_by_type(entity_types))

return map
```
_remove_entities_by_name_
```lua

local map = b.full_shape
local entity_names = {'biter-spawner', 'spitter-spawner'}

map = b.apply_effect(map, b.remove_entities_by_names(entity_names))


return map
```

### DIFF
--- a/.luacheckrc
+++ b/.luacheckrc
@@ -169,14 +169,14 @@ stds.factorio = {
                 "add_shift", "merge", "premul_color", "encode", "decode", "insert_safe",
                 table = {
                     fields = {
-                        "compare", "deepcopy"
+                        "compare", "deepcopy", "shallow_copy"
                     },
                 },
             },
         },
         table = {
             fields = {
-                "compare", "deepcopy"
+                "compare", "deepcopy", "shallow_copy"
             },
         },
     },
--- a/map_gen/maps/HilbertSandTrap.lua
+++ b/map_gen/maps/HilbertSandTrap.lua
@@ -99,28 +99,7 @@ RS.set_map_gen_settings(
 		water_settings
     }
 )
---remove resources from sand
-local function no_resources(_, _, world, tile)
-    local entites =
-        world.surface.find_entities_filtered(
-        {type = 'resource', area = {{world.x, world.y}, {world.x + 1, world.y + 1}}}
-    )
-    for i = 1, #entites do
-        entites[i].destroy()
-    end
-    return tile
-end
---remove trees from sand
-local function no_trees(_, _, world, tile)
-    local entites =
-        world.surface.find_entities_filtered(
-        {type = 'tree', area = {{world.x, world.y}, {world.x + 1, world.y + 1}}}
-    )
-    for i = 1, #entites do
-        entites[i].destroy()
-    end
-    return tile
-end
+
 --start hilbert design. Note: The following code does contain bugs. Jayefuu and R.Nukem are aware of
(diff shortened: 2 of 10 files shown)


SimonFlapse:
https://lua-api.factorio.com/latest/LuaSurface.html#LuaSurface.find_enemy_units says that `find_enemy_units` are more efficient than `find_entities` so I've included a function that only looks at units in https://github.com/Refactorio/RedMew/pull/994/commits/3c41275624cda3240ce8ecb005bab5c067c817e7

SimonFlapse:
After testing the remove_units function had next to no effect since it only works on 'units' (biters and spitters, characters) and they would often not spawn before after the generate script has run.

grilledham:
I re-worked the remove_* builders that you added.
 - I changed them to not be effects so they are easier to use. example use:
```lua
shape = b.remove_map_gen_resources(shape)
```
To remove resources made by the factorio map gen for the shape.
 - I generalised the by type and by name to by filter which is the same filter facotrio uses.
 - I made some named ones for common cases, e.g. `remove_map_gen_resources`, `remove_map_gen_enemies`
 - I gave them the 'map_gen' name to fit the current naming convention.

I also added a `remove_entities_by_name` which acts on the entities that the shape makes. Personally I think it's kinda dumb as you would just not add entities you don't want, but it serves an example of how you could do that.

Also changed the current maps to use the remove_* builders where possible.

grilledham:
It appears I can't request you review this. Maybe because you are the author.
But could you review this please 😃

grilledham (approved):

SimonFlapse:
LGTM 😃

I understand your concern about the `remove_entities_by_name`, but it's nice to have the option to remove stuff. I'd just add a quick comment above it to warn people that it only removes entities that has been "placed" through the builders and not stuff on the surface.

I'll write up some documentation and then merge it, unless you have something else to add.

grilledham:
If you want to write some documentation that is cool.
I'm happy for you to merge this when you are ready.

Also in principle I'm happy for comments to be added to the builder function, if that is something we want to do.

SimonFlapse:
> If you want to write some documentation that is cool.
> I'm happy for you to merge this when you are ready.
> 
> Also in principle I'm happy for comments to be added to the builder function, if that is something we want to do.

I think I could do links for each function to the wiki. That way it's easier to keep the documentation updated, plus we can include examples

SimonFlapse:
Quick reference for the future:

The following maps have been changed to reflect the changes of this pull request:

- [ ] HilbertSandTrap

- [ ] connected_dots

- [ ] crash_site/scenario

- [ ] fractal_balls

- [ ] line_and_trees

- [ ] maltese_crossings

- [ ] tetris/shape

PR #995, opened 2019-10-10
[Snake] Added on_snake_player_died event

SimonFlapse:
I've added a event to the snake game for when a player looses

> on_snake_player_died
        Called when a player have died in a game of snake
        Contains
            name :: uint: Unique identifier of the event
            tick :: uint: Tick the event was generated.
            player :: LuaPlayer
            score :: uint: Score reached

Also included a demo map for playing Snake

### DIFF
--- a/features/snake/game.lua
+++ b/features/snake/game.lua
@@ -63,8 +63,11 @@ local function destroy_snake(index, snake)
         return
     end
 
+    player.set_controller{type = defines.controllers.spectator}
+
     local score = queue_size(snake.queue)
     game.print({'snake.snake_destroyed', player.name, score})
+
     script.raise_event(Public.events.on_snake_player_died, {
         player = player,
         score = score
--- a/features/snake/game.lua
+++ b/features/snake/game.lua
@@ -108,6 +108,7 @@ local function spawn_food()
             entity =
                 surface.create_entity({name = 'compilatron', position = pos, force = 'neutral', direction = random(7)})
             entity.active = false
+            entity.destructible = false
             food_count = food_count + 1
 
             break
@@ -221,6 +222,7 @@ local function tick_snake(index, snake)
         entity.character_running_speed_modifier = -1
         entity.color = player.color
         entity.active = false
+        entity.destructible = false
         push_to_end(snake_queue, {entity = entity, cord = tail_cord})
 
         board.food_count = board.food_count - 1
@@ -354,6 +356,7 @@ local function new_snake(player)
     player.create_character()
     character = player.character
     character.character_running_speed_modifier = -1
+    character.destructible = false
 
     local queue = queue_new()
     push(queue, {entity = character, cord = cord})
@@ -387,7 +390,7 @@ Public = {
         on_snake_player_died
         Called when a player have died in a game of snake
         Contains
-            name :: defines.events: Identifier of the event
+            name :: uint: Unique identifier of the event
             tick :: uint: Tick the event was generated.
             player :: LuaPlayer
             score :: uint: Score reached
--- a/features/snake/game.lua
+++ b/features/snake/game.lua
@@ -15,6 +15,8 @@ local queue_size = Queue.size
 local queue_pairs = Queue.pairs
 local pairs = pairs
 
+local Public = {}
+
 local snakes = {} -- player_index -> snake_data {is_marked_for_destroy:bool, queue :Queue of {entity, cord} }
 local board = {
     size = 0,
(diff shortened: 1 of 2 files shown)


grilledham:
While we have this PR open let's fix a bug I noticed last time I used this.

Line 223 In the tick_snake function:
When we spawn a new character we should make the entity as indestructible
Same for line 356 in the new_snake function.

features/snake/game.lua
```diff
@@ -374,7 +381,20 @@ local function new_game(surface, position, size, update_rate, max_food)
...
 end
 
-local Public = {}
+Public = {
+    events = {
+        --[[
+        on_snake_player_died
+        Called when a player have died in a game of snake
+        Contains
+            name :: defines.events: Identifier of the event
```
grilledham: If we don't store the event in `defines.event` I think this comment is misleading. 
Maybe instead it should just say `The unique id for the event.`

SimonFlapse: True, I was wondering about that too, I just copied the description from another custom event

SimonFlapse:
> While we have this PR open let's fix a bug I noticed last time I used this.
> 
> Line 223 In the tick_snake function:
> When we spawn a new character we should make the entity as indestructible
> Same for line 356 in the new_snake function.

I also did it on line 111 for the "food"

map_gen/maps/snake_demo.lua
```diff
@@ -0,0 +1,33 @@
...
+local config = global.config
+local size = 45
+
+config.market.enabled = false
+
+local snake_generate = Token.register(function()
+    local position = {x = -floor(size), y = 5}
+    local max_food = 8
+    local speed = 30
+    remote.call('snake', 'start_game', RS.get_surface(), position, size, speed, max_food)
```
grilledham: You don't have to use `remote.call`, you can instead use 
```lua
local Snake = require 'features.snake.game'
Snake.start_game(RS.get_surface(), position, size, speed, max_food)
```

map_gen/maps/snake_demo.lua
```diff
@@ -0,0 +1,33 @@
...
+
+Event.on_init(function()
+    Task.set_timeout_in_ticks(60, snake_generate)
+end)
+
+local snake_game = require 'features.snake.game'
+
+local function on_snake_player_died(event)
+    local player = event.player
+    player.set_controller{type = defines.controllers.spectator}
```
grilledham: I think it would make sense to add this set controller logic to the destroy_snake function

map_gen/maps/snake_demo.lua
SimonFlapse: Yeah, I just wanted to showcase the remote function since you registered it in the `control.lua` file

map_gen/maps/snake_demo.lua
SimonFlapse: Okay, I'll go add it

map_gen/maps/snake_demo.lua
```diff
@@ -17,6 +18,7 @@ local snake_generate = Token.register(function()
     local max_food = 8
     local speed = 30
     Snake.start_game(RS.get_surface(), position, size, speed, max_food)
+    Snake_Gui.show()
```
grilledham: I forgot about the snake gui, maybe it should use snake control instead so it's just one call.

map_gen/maps/snake_demo.lua
```diff
@@ -0,0 +1,35 @@
...
+    -- An alternative is to use: remote.call('snake', 'start_game', RS.get_surface(), position, size, speed, max_food)
+end)
+
+Event.on_init(function()
+    Task.set_timeout_in_ticks(60, snake_generate)
+end)
+
+local function on_snake_player_died(event)
+    local player = event.player
+    player.set_controller{type = defines.controllers.spectator}
```
grilledham: Now for the purpose of the demo this event needs to do something else, as it's confusing that it does something that is already done.
Or you just remove the event from the demo.

map_gen/maps/snake_demo.lua
```diff
@@ -0,0 +1,28 @@
+local Event = require 'utils.event'
+local Token = require 'utils.token'
+local Task = require 'utils.task'
+local b = require 'map_gen.shared.builders'
+local RS = require 'map_gen.shared.redmew_surface'
+local floor = math.floor
+local Snake = require 'features.snake.game'
```
SimonFlapse:
```suggestion
```
To fix Travis

grilledham (approved):

PR #997, opened 2019-10-13
Added Snakey Swamp map script.

Soggs:
Creates a ribbon works with a snakey path surrounded by water at the top and bottom.


map_gen/maps/snakey_swamp.lua
```diff
@@ -0,0 +1,97 @@
+-- Snakey Swamp by Soggs
+local b = require 'map_gen.shared.builders'
+local RS = require 'map_gen.shared.redmew_surface'
+local MGSP = require 'resources.map_gen_settings'
+local ceil = math.ceil
+local min = math.min
+
+-- Disable Landfill by default
+local Event = require 'utils.event'
+Event.add(
```
SimonFlapse: This could have been done with the Event.on_init(handler) this way you could enable landfill in game. Right now it will remove the landfill research everytime someone joins for the first time.

map_gen/maps/snakey_swamp.lua
```diff
@@ -0,0 +1,97 @@
...
+        world.surface.find_entities_filtered(
+            {type = 'resource', area = {{world.x, world.y}, {world.x+1, world.y+1}}}
+        )
+    ) do
+        e.destroy()
+    end
+    return t
+end
+
+ore_rectangle = b.apply_effect(ore_rectangle, no_resources)
```
SimonFlapse: With #994 merged you can now use the b.remove_map_gen_resources(shape) function instead of your no_resources function

SimonFlapse:
image
You would most likely need to increase the resource frequency for this map.

Almost no ores and no oil or uranium within 1024 tiles.

You can do this through the `RS.set_map_gen_settings`
You can look in 'resources.map_gen_settings' for presets.

I've also pushed two commits to this branch, so remember to pull before making any more commits

map_gen/maps/snakey_swamp.lua
SimonFlapse: Fixed in https://github.com/Refactorio/RedMew/pull/997/commits/4e4945f159845bad1687d2fcf7a7f7f65222c59b

map_gen/maps/snakey_swamp.lua
SimonFlapse: Fixed in https://github.com/Refactorio/RedMew/pull/997/commits/4e4945f159845bad1687d2fcf7a7f7f65222c59b

Soggs:
I know about the resources. Such is the fate of ribbon style worlds but I try to avoid overwriting map gen settings as much as possible. I know you typically do that differently but since I come from the modded angle having to go into the map script and change a lot about resource settings is quite annoying.

Soggs:
image

To illustrate how Snakier Swamps looks. Number of turns is variable, as is the water border width, path length and width, divider width and tile type.

Soggs:
Note for myself: Remove decoratives on the dividers and try shallow water 2 as well as green water.

SimonFlapse:
> Note for myself: Remove decoratives on the dividers and try shallow water 2 as well as green water.

I'm trying to introduce a function to the builders that allows you to do this quickly and only for a specific shape. See #998

map_gen/maps/snakier_swamp.lua
```diff
@@ -0,0 +1,113 @@
+-- Snakey Swamp by Soggs
```
grilledham: Shouldn't this be `-- Snakier Swamp by whoever`

map_gen/maps/snakey_swamp.lua
```diff
@@ -0,0 +1,89 @@
...
+local quarter_height = path_height / 4
+local max_height = 50
+local max_width = 50
+local ore_rectangle = b.rectangle(min(path_width - 2, max_height), min(quarter_height - 2, max_width) - 10) -- -10 to allow space for market and other buildings
+local ore_spacing = min(quarter_height, max_height + 2)
+local function amount(a)
+    return function(_, _)
+        return ceil(a / min(path_width - 2, max_height) / min(quarter_height - 2, max_width))
+    end
+end
```
grilledham, line 68: The value returned by the inner function is constant from the inner function's point of view. So this could be 
```lua
local function amount(a)
    local value = ceil(a / min(path_width - 2, max_height) / min(quarter_height - 2, max_width))
    return function()
        return value
    end
end
```
To save having to recalculate the same value at runtime.

Same again for snakier swamp

grilledham:
This looks good to me. Waiting for the merge conflict with #998 to be resolved before I approve.

map_gen/maps/snakey_swamp.lua
Soggs: I guess I got too used to compilers doing that for me. :D

map_gen/maps/snakey_swamp.lua
```diff
@@ -0,0 +1,89 @@
...
+local map = b.if_else(grid, b.full_shape)
+
+-- Starting resources
+local quarter_height = path_height / 4
+local max_height = 50
+local max_width = 50
+local ore_rectangle = b.rectangle(min(path_width - 2, max_height), min(quarter_height - 2, max_width) - 10) -- -10 to allow space for market and other buildings
+local ore_spacing = min(quarter_height, max_height + 2)
+local function amount(a)
+    return function(_, _)
```
SimonFlapse, line 65:
```suggestion
    return function()
```
Since you're not using any parameters for the return function you can just leave the parameters empty.

map_gen/maps/snakier_swamp.lua
```diff
@@ -0,0 +1,114 @@
...
+
+--Starting resources
+local max_height = 50
+local max_width = 50
+local ore_rectangle = b.rectangle(min(path_width-2, max_height), min(path_width-2, max_width))
+local ore_spacing = min(path_width, max_height+2)
+local function amount(a)
+    local value = ceil(a /min(path_width-2, max_height) /min(path_width-2, max_width))
+    return
+        function (_, _)
```
SimonFlapse, line 84:
```suggestion
        function ()
```
Same thing here.
You may also want to use the same function in both maps, but that is up to you.

SimonFlapse (approved):
LGTM, you may want to fix my other two comments for consistency.

grilledham:
LGTM, plus what Simon said.

grilledham (approved):

PR #998, opened 2019-10-14
Hilbert Water Trap

SimonFlapse:
I have adapted the Hilbert Sand Trap to use the new shallow water to completely remove the need for logic to prevent building on the sand

image

Because of the water, I had to change a bit in the `feature/market.lua` to allow us to delay the spawning of the market.

### DIFF
--- a/features/market.lua
+++ b/features/market.lua
@@ -326,6 +326,19 @@ Event.add(defines.events.on_entity_died, fish_drop_entity_died)
 Event.add(Retailer.events.on_market_purchase, market_item_purchased)
 Event.add(defines.events.on_player_crafted_item, fish_player_crafted_item)
 Event.add(defines.events.on_player_created, player_created)
-if global.config.market.create_standard_market then
-    Event.on_init(spawn_market)
+if market_config.create_standard_market then
+    local function delay_spawn_market()
+        spawn_market()
+    end
+    local spawn_market_token = Token.register(delay_spawn_market)
+    Event.on_init(
+        function()
+            local delay = market_config.delay
+            if delay then
+                Task.set_timeout_in_ticks(delay, spawn_market_token)
+            else
+                delay_spawn_market()
+            end
+        end
+    )
 end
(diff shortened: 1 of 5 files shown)


SimonFlapse:
@grilledham https://github.com/Refactorio/RedMew/pull/998/commits/1ddc2dc1f2e82f75680209bba9713209b198d461 I've added the ability to remove decoratives in the same style as the remove entities. I'd need your review on the changes

features/market.lua
```diff
@@ -326,6 +328,25 @@ Event.add(defines.events.on_entity_died, fish_drop_entity_died)
...
+            end
+        )
+    else
+        Event.on_init(
+            function()
+                delay_spawn_market()
+            end
+        )
+    end
 end
```
grilledham, line 342: I'm curious why the spawn market delay is needed and slightly worried that it might break assumptions other maps make about when the spawn market is available . But I don't feel like investigating now.

However couldn't this be simplified to:
```lua
if market_config.create_standard_market then
    local delay = market_config.delay
    if delay then
        local spawn_market_token = Token.register(spawn_market)
        Event.on_init(
            function()
                Task.set_timeout_in_ticks(delay, spawn_market_token)
            end
        )
    else
        Event.on_init(spawn_market)
    end
end
```

grilledham (approved):

features/market.lua
SimonFlapse: It's needed because the market is removed when shallow-water is spawned (Which happens slightly after first on_init) I'll simplify it, we don't need to change the logic of delay_spawn_market function to include anything else than spawn_market right now

map_gen/shared/builders.lua
```diff
@@ -821,21 +821,21 @@ function Builders.remove_entities_by_name(shape, names)
 end
 
 --- Docs: https://github.com/Refactorio/RedMew/wiki/Using-the-Builders/_edit#buildersremove_map_gen_decoratives_by_filter
-function Builders.remove_map_gen_decoratives_by_filter(shape, filter)
-    if filter then
-        filter = shallow_copy(filter)
+function Builders.remove_map_gen_decoratives_by_filter(shape, optional_filter)
```
grilledham: I did also suggest renaming the function to `remove_map_gen_decoratives` as the filter isn't always needed.

PR #1001, opened 2019-10-18
Server description generator

SimonFlapse:
Updated the rich text generation gui to support more types like technologies and achievements

Example:
image

Created a new Description Generator that can be used to generate the server name, server description and server tags, which then can be pasted into the relevant fields in the `/config` gui

Changed the info gui to help visual guide the maximum text width supported by the server browser

Hoping this request will be met: https://forums.factorio.com/viewtopic.php?f=28&t=76911 So that the server description process can be automated.

Example:
image

It also allows scenarios to register primary creator and secondary creator

### DIFF
--- a/config.lua
+++ b/config.lua
@@ -10,11 +10,15 @@ global.config = {
         -- The title of the map
         map_name_key = 'This Map has no name',
         -- The long description of the map, typically 1 paragraph
-        map_description_key = "This section is supposed to be filled out on a per map basis.\nIf you're seeing this message, ping the admin team to get a description\nadded for this map. 20 coins is rewarded to the first person that points this out.",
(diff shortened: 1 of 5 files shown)


SimonFlapse:
There seems to be some inconsistency between the server browsers description when handling word wrap. That is why I added the center text function in the first place. The center text function is very ineffective. I will try to get the server description flow to have `horizontal_align = 'center'`

Word-wrapping:
image

Without word-wrapping
image

https://forums.factorio.com/viewtopic.php?f=7&t=76980

Expected centering of wordwrap (First two lines)
image

Actual behavior:
image

It appears that it might just be an effect of the description being a multi line label, though this is possible:
image

It has been moved from Not a Bug to Ideas and Suggestions, so we most likely have to design our description around that fact that the description is left aligned

PR #1003, opened 2019-11-01
Crashsite: Change biters to spawn worms on death

grilledham:
I want to try this, mainly because I think it will be better for performance. 
If we don't like it we can revert it.


SimonFlapse:
Just note that worms often outrange any defenses if the biters where killed by a turret etc.

grilledham:
We had the same problem with spitters before. But overall I do think this change will make crashsite slightly harder.

Issue #1013, opened 2019-11-28
Thanksgiving/ "Turkey" map feedback


theorangeangle:
- [ ] Add a "funny/not funny" button after each joke for people to click
- [ ] Have the jokes in toasts instead of chat prints
(advanced route: do a cutscene)
- [ ] After 24 hours, halve all market prices and call it a "black friday bargain"
- [ ] Rename the pokes to all thanksgiving themed items
- [ ] make the "scarf" (refined concrete) non removable by players

- [x] make the vanilla ores disappear

theorangeangle:
#1014 fixes vanilla ores

PR #1014, opened 2019-11-28
Turkey.lua Improvements

Zahir-Khan:
Removed vanilla ore spawn.

### DIFF
--- a/map_gen/maps/turkey.lua
+++ b/map_gen/maps/turkey.lua
@@ -10,6 +10,7 @@ local turkey_message_random = require 'resources.turkey_messages'
 
 RS.set_map_gen_settings(
     {
+	MGSP.ore_oil_none,
         MGSP.cliff_none
     }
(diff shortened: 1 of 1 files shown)


grilledham:
Line 150 could use `Event.on_nth_tick`

Zahir-Khan:
> Line 150 could use `Event.on_nth_tick`

There was a mention of desync possibility, that is why I didn't use it and also I don't quite understand it fully.

grilledham:
We use it a lot, so there shouldn't be any desync risk.

Something like this should work:
```lua
Event.on_nth_tick(36000,
    function(event)
            local message = turkey_message_random[math.random(#turkey_message_random)]
            game.print('[color=yellow][font=compi]' .. message .. '[/font][/color]')
    end
)
```

Zahir-Khan:
This closes #1013

PR #1015, opened 2019-12-08
Minor crashsite cutscene improvements

Zahir-Khan:
1. Biters are now disabled on the cutscene surface.
2. Day only enabled on the cutscene surface.

### DIFF
--- a/map_gen/maps/crash_site/scenario.lua
+++ b/map_gen/maps/crash_site/scenario.lua
@@ -183,7 +183,8 @@ local function init()
     local outpost_builder = OutpostBuilder.new(outpost_random)
 
     if on_init then
-        game.create_surface('cutscene')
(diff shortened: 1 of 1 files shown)


Zahir-Khan:
Needs fixing
If a player dies during a cutscene or if a player initiates a cutscene while dead - the player is locked out and the cutscene never ends.
This issue is on the Redmew side as internally the game ends the cutscene but redmew is stuck in a code loop. Hint: see running_cutscene in cutscene_controller.lua

grilledham (approved):
I think we should merge this, then you work on the other issues in a new PR, if that's OK with you?

Zahir-Khan:
> I think we should merge this, then you work on the other issues in a new PR, if that's OK with you?

yes, that sounds better.

PR #1019, opened 2019-12-31
Crowdin sync 31-12-2019

SimonFlapse:
Also added new strings to crowdin from the english source. This means that some strings have been removed too.


grilledham (approved):

grilledham:
@SimonFlapse is this ready to be merged? If so I think this can go in now.

theorangeangle:
Looks good to me, although we might want to mark on Crowdin that the translations have been accepted.

PR #1025, opened 2020-02-04
Diggy update-create_particles

Claude477:
create_particles.lua fixed with changed create_entity=>create_particle to get working in 0.18.3


Claude477:
Making different commit so only 1 file updated each pull.

Claude477:

PR #1026, opened 2020-02-04
Diggy update-diggy_hole

Claude477:
diggy_hole.lua  fixed with changed create_entity=>create_particle to get working in 0.18.3


Claude477:

PR #1027, opened 2020-02-04
Diggy update-alien_spawner

Claude477:
alien_spawner.lua fixed with changed create_entity=>create_particle to get working in 0.18.3


Claude477:

PR #1033, opened 2020-02-06
Diggy patch for ore showing up on top of water

BlueRock2:
This PR is related to issue #552 and PR #983.    PR #983 was done a couple months ago to fix the issue #552 but a new bug was introduced where when in areas where an ore vein is present the ore would be placed back down on top of water.   This patch uses find_entities_filtered to only consider rocks that were affecting the ceiling stresses and not touch the ore.


SimonFlapse (commented):
To me it seems that this change would allow other entities than rocks (Like chests) to spawn on water, and not be moved.

BlueRock2:
I was looking mainly for ore on water when testing this, but I will temporarily bump up the water size and number of chests quite a bit and do some more testing today.

BlueRock2:
Ok, so I cranked up (temporarily) both the number of chests and water generation and took a couple screenshots of a cleared void area before and after this patch. 
before_water_patch
after_water_patch
I hadn't mentioned above, but I also did some prior testing to turn on the debugging showing the stress grid values in diggy and opened several large rooms in game to be sure that some rocks collapse into water as expected to support the ceiling and that there weren't any wide areas of water without any support when a room opens up.

BlueRock2:
@SimonFlapse  To address your specific concern.   Before the rocks are teleported, I believe it is only checking to see if a water tile is *about* to be placed, and that the actual placement occurs on line 84 (which is how the ore and chests get overwritten with water post patch).   After the water is set, only the rocks get teleported back to their original position (line 86).   Without this patch, everything gets placed back on top of the water when the teleport is reverted back.  Later, the rocks on top of the water get destroyed as the rooms open so that ceiling stresses get updated and some rocks fall into the water to hold up the ceiling.    Images above have cave collapses disabled which is why no rocks appear in the water, but I did separate tests to make sure everything was still fine there.

BlueRock2:
I originally invited Linaori to review this before I realized that she has been inactive for some time in regards to Factorio.   But since the change is pretty small and I believe I have tested it well, this should be ready to merge.

SimonFlapse (approved):
The behavior is as expected, so it looks good to me.

PR #1032, opened 2020-02-05
Diggy-map patch

Claude477:
Diggy-map patch by BlueRock

PR #1035, opened 2020-02-08
diggy hail_hydra patch

Claude477:
this fix some errors

### DIFF
--- a/map_gen/shared/hail_hydra.lua
+++ b/map_gen/shared/hail_hydra.lua
@@ -94,6 +94,7 @@ local on_died =
         local cause = event.cause
         local surface = entity.surface
         local create_entity = surface.create_entity
+        local create_particle = surface.create_particle
         local find_non_colliding_position = surface.find_non_colliding_position
 
         local command = create_attack_command(position, cause)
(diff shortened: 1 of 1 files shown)


SimonFlapse (commented):
Wouldn't it be better to merge this into #1032? Since it depends upon it.

Claude477:
Yes is true but #1032 has 3 critical files need patch frist then this hail_hydra file. this meybe can changed if not working corretly. plus this need tests out.

Claude477:
S9-server has diggy2020_patch3-deflated.zip if want checkout how is works out.

BlueRock2:
Did some solo testing by setting debug and cheat mode to true, set evolution to 100%, ran /c game.player.force.research_all_technologies()  so that I get the weapon damage upgrades.    Verified blood explosions occurring on alien death (a little larger than the normal blood spray when doing damage that doesn't cause death)  Also verified new aliens randomly spawn on death as expected with hail hydra.

BlueRock2:
After I tested this update, Claude477 indicated ok to merge this update in with PR 1032 which I did this morning.  This PR can be cancelled/closed.

PR #1038, opened 2020-02-16
Diggy - Prevent oil from spawning too close to other oil patches

BlueRock2:
Over time I have seen several people report that some oil patches that were too close to place adjacent pumpjacks in diggy.   Although it isn't game breaking (you can still choose one or the other to tap into) it keeps getting reported as buggy behavior.    Since vanilla oil spawn generation is too infrequent even on highest frequency for diggy, I created this update for scattered_resources.lua that will prevent oil patches from spawning less than a two tile gap from another oil patch.

### DIFF
--- a/map_gen/maps/diggy/feature/scattered_resources.lua
+++ b/map_gen/maps/diggy/feature/scattered_resources.lua
@@ -36,14 +36,9 @@ end
 
 --overcrowded_resource returns true if one or more entity <resource_name> is found within <min_gap> tiles from <position>
 local function overcrowded_resource(surface, position, resource_name, min_gap)
-    local d = min_gap
-    local bounding_box = {{x = position.x - d + 1, y = position.y - d + 1}, {x = position.x + d, y = position.y + d}}
+    local bounding_box = {{x = position.x - min_gap + 1, y = position.y - min_gap + 1}, {x = position.x + min_gap, y = position.y + min_gap}}
     local found_resources = surface.count_entities_filtered{area = bounding_box, name = resource_name, limit = 1}
-    if found_resources == 0 then
-        return false
-    else
-        return true
-    end
+    return found_resources ~= 0
 end
 
 --[[--
@@ -148,7 +143,7 @@ function ScatteredResources.register(config)
 
         local position = {x = x, y = y}
         if resource_name == 'crude-oil' then
-            local min_gap = 2  --must be at least this many tiles between crude-oil patches
+            local min_gap = 2  --default is 2.  Represents minimum tile gap between resources.
             if overcrowded_resource(surface, position, resource_name, min_gap) then
                 return false  --when overcrowded_resource is true, skip this resource spawn
             end
--- a/map_gen/maps/diggy/feature/scattered_resources.lua
+++ b/map_gen/maps/diggy/feature/scattered_resources.lua
@@ -34,6 +34,18 @@ local function get_name_by_weight(collection, sum)
     Debug.print('Current \'' .. current .. '\' should be higher or equal to random \'' .. target .. '\'')
 end
 
+--overcrowded_resource returns true if one or more entity <resource_name> is found within <min_gap> tiles from <position>
+local function overcrowded_resource(surface, position, resource_name, min_gap)
+    local d = min_gap
+    local bounding_box = {{x = position.x - d + 1, y = position.y - d + 1}, {x = position.x + d, y = position.y + d}}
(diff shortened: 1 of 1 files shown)


BlueRock2:
In testing I temporarily increased oil spawns 3x and marked locations that an oil spawn was prevented using a red chest in the screenshot below.
image

map_gen/maps/diggy/feature/scattered_resources.lua
```diff
@@ -34,6 +34,18 @@ local function get_name_by_weight(collection, sum)
     Debug.print('Current \'' .. current .. '\' should be higher or equal to random \'' .. target .. '\'')
 end
 
+--overcrowded_resource returns true if one or more entity <resource_name> is found within <min_gap> tiles from <position>
+local function overcrowded_resource(surface, position, resource_name, min_gap)
+    local d = min_gap
+    local bounding_box = {{x = position.x - d + 1, y = position.y - d + 1}, {x = position.x + d, y = position.y + d}}
+    local found_resources = surface.find_entities_filtered{area = bounding_box, name = resource_name}
```
grilledham: May I suggest <https://lua-api.factorio.com/latest/LuaSurface.html#LuaSurface.count_entities_filtered> instead.
Also with `limit = 1`.

grilledham (approved):
Looks good to me.

BlueRock2:
Updated per grilledham's suggestions.  Re-tested again with a temporary high oil spawn rate to confirm update works as expected.   Unless anyone else needs to review or approve, this should be ready to merge.

map_gen/maps/diggy/feature/scattered_resources.lua
```diff
@@ -34,6 +34,18 @@ local function get_name_by_weight(collection, sum)
...
+--overcrowded_resource returns true if one or more entity <resource_name> is found within <min_gap> tiles from <position>
+local function overcrowded_resource(surface, position, resource_name, min_gap)
+    local d = min_gap
+    local bounding_box = {{x = position.x - d + 1, y = position.y - d + 1}, {x = position.x + d, y = position.y + d}}
+    local found_resources = surface.count_entities_filtered{area = bounding_box, name = resource_name, limit = 1}
+    if found_resources == 0 then
+        return false
+    else
+        return true
+    end
```
SimonFlapse: Not to nitpick but you could write this in a cleaner way
I'd think either of these would work `return found_resources ~= 0` or `return not (found_resources == 0)`

However if you feel the conditional statement could be expanded with more logic, just ignore my comment.

map_gen/maps/diggy/feature/scattered_resources.lua
```diff
@@ -34,6 +34,18 @@ local function get_name_by_weight(collection, sum)
     Debug.print('Current \'' .. current .. '\' should be higher or equal to random \'' .. target .. '\'')
 end
 
+--overcrowded_resource returns true if one or more entity <resource_name> is found within <min_gap> tiles from <position>
+local function overcrowded_resource(surface, position, resource_name, min_gap)
+    local d = min_gap
```
SimonFlapse: It seems you are "renaming" the `min_gap` variable to `d`. I can't see the purpose of this assignment apart from keeping the following line 6 characters shorter for each reference to the variable.

SimonFlapse (approved):
LGTM. I do have some minor suggestion that I think would improve the readability.

map_gen/maps/diggy/feature/scattered_resources.lua
```diff
@@ -134,14 +146,22 @@ function ScatteredResources.register(config)
             return false
         end
 
+        local position = {x = x, y = y}
+        if resource_name == 'crude-oil' then
+            local min_gap = 2  --must be at least this many tiles between crude-oil patches
```
SimonFlapse: Sometimes we like to write the default or recommended value for variables. Take the case that if we were to enforce a larger minimum gap for diggy, but later on choose to lower the value. In this case having a comment stating the recommended or lowest value possible would be nice to have in the future.

Just my two cents.

map_gen/maps/diggy/feature/scattered_resources.lua
BlueRock2: Thanks for the feedback.  Updated with your suggestion.

map_gen/maps/diggy/feature/scattered_resources.lua
BlueRock2: Yeah, was to make the one line shorter, but it wasn't all that long so took your suggestion.

PR #1039, opened 2020-02-18
Diggy mask debug surface bug

### LINKED ISSUE #1037: Diggy debug grid print
Currently, if you enable the debug grid in 0.18.6 it will systematically produce a fatal error.

SimonFlapse:
Prevents the print_grid_value function from producing an error when surface is nil or invalid.

The Mask debugging setting may, or may not work. I'm simply not sure what it is supposed to do.

closes #1037

Ideally we would have to make sure the mask debugging works, before merging this PR


BlueRock2 (commented):
I tried this change, but received an error that there was an attempt to index surface (nil value).  No error if I check surface first then if not nil, check surface.valid.   However, I think the code that displays the mask only executes once early on.  So if the surface isn't ready yet, it never displays?  I think the call to display the mask might still need to be relocated, but not sure where.  In a temp fix I did, I displayed right after the first mined rock message for user pops up,  but is there a way to call the function when the surface is ready and user first gains control?

SimonFlapse:
> I tried this change, but received an error that there was an attempt to index surface (nil value). No error if I check surface first then if not nil, check surface.valid. However, I think the code that displays the mask only executes once early on. So if the surface isn't ready yet, it never displays? I think the call to display the mask might still need to be relocated, but not sure where. In a temp fix I did, I displayed right after the first mined rock message for user pops up, but is there a way to call the function when the surface is ready and user first gains control?

Weird. The new code snippet should return if the surface is nil, and only try to index the .valid field if the surface isn’t nil.

BlueRock2:
Still concerned that even with the surface check the code gets blocked the one time it tries to run.   Curious Simon what you think about this:   Within diggy_cave_collapse.lua  the code that prints the mask is:
```
if (config.enable_mask_debug) then
        mask_disc_blur(
            0,
            0,
            10,
            function(x, y, fraction)
                Debug.print_grid_value(fraction, surface, {x = x, y = y})
            end
        )
    end
```
 It is currently located in a function that does config registration and sets up events but think it is trying run before the diggy surface is setup.   However, I also noticed that diggy_cave_collapse does have some code to check for the event, on_surface_created and a handler for it.
`  Event.add(defines.events.on_surface_created, on_surface_created)`
and `on_surface_created = function(event)`     Would it make sense to move mask debug code block within this function?     If so, then I think before running it we also check to be sure index == 1.

Issue #244, opened 2018-10-19
Diggy Scenario Idea - Biter caves:


naquadah:
**Biter caves while mining rocks**
Mining rocks can **rarely** cause a cave opening with "lots(40-60?)" of biters and spitters to spawn.

Trigger options when mining:
- Completely random 
- Based on distance(the longer radius you get from spawn, the bigger chance for Biter Caves. (danger per distance))

Adds more combat diversity and challenge to map.
Distance/danger factor can prevent people from digging miles, especially early.

iltar:
Currently biters are spawned by chance, we could alter this chance by increasing spawn rates for certain noise values, matching the noise map from the room generator. This would make all rooms spawn it though, so we would probably have to use a second noise map or distance calculator to control this.

PR #1041, opened 2020-03-09
Add distant large biter caves to Diggy

BlueRock2:
Description:   This pull request is related to Diggy feedback/todo collection thread #926 and specifically, Diggy Scenario Idea - Biter caves: #244.   
The implementation below creates two different noise patterns.   One compound noise pattern is used near the spawn area and creates small openings near the spawn similar in size to current Diggy implementation.   A second compound noise pattern is used when over a certain distance (set in config) from the origin which has the potential to generate cave-like openings that can be approximately 2-10 times as large in size.     Effort was made to use the same seeded_noise function in scattered_resources.lua to move closer towards another wishlist item "Replace room and terrain generation by the same system that generates ores". 
Additional small change in this PR:  Allow for the use of the new landfill tile introduced in 0.17.10 which looks great in Diggy.  (it used to be grass-1 and Diggy overwrote it with random dirt tiles). 
Would like to open to beta test play and gather some feedback prior to merging.

### DIFF
--- a/map_gen/maps/diggy/config.lua
+++ b/map_gen/maps/diggy/config.lua
@@ -169,15 +169,35 @@ local Config = {
         -- automatically opens areas
         simple_room_generator = {
             enabled = true,
-            -- value between 0 and 1, higher value means stronger variance between coordinates
-            noise_variance = 0.066,
+
+            --distance in tiles that the room noise pattern changes to enable large room (cave) generation
+            --if large caves are not desired, this number can be set to a large distance, i.e. 100000
+            large_room_minimum_distance = 128,
+
+            noise_settings = {
+                -- higher numbers for variance actually have less variance  i.e. larger room features.
+                -- Similar system used ores in scattered_resources.lua.  Here we are using two noise patterns for rooms (caves)
+                -- starting_sources is noise pattern used near spawn to generate small rooms similar to traditional Diggy
+                -- distant_sources is noise pattern used after large_room_minimum_distance to sometimes generate large caves
+                starting_sources = {
+                    {variance = 60, weight = 0, offset = 000, type = "perlin"},
+                    {variance = 20, weight = 1, offset = 150, type = "perlin"},
+                    {variance = 8, weight = 0.2, offset = 300, type = "perlin"}
+                },
+                distant_sources = {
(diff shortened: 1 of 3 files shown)


grilledham:
LGTM.

Also, I'm a bit out of the loop, was this tested in multiplayer? If so any Feedback?

BlueRock2:
TL;DR  We had a fun beta test and there was some actual teamwork fighting in Diggy.   No obvious issues.   Just a few things to watch and potentially improve on in the future.

I opened this to a beta test on on Thurs March 12 and I think we had five players try it out the first day and a few the next.    It was played up until blue science/advanced oil.   Some thoughts:
Pros:  
   -Large caves added some added excitement as there could often be a rather large number of biters inside   This also added a new element of teamwork to Diggy, as some caves needed three or more players to be able to clear out and more help with corpse recovery.
  -Appears it could reduce some of the drudgery of mining late game.   I have seen some diggy games end because ore dried up too fast late game no-one wanted to mine new ore.   A good size cave can uncover a good amount of ore. 
  -Enabling the landfill that was introduced in Factorio 17.10 looks good.
Cons?:   
  -Maybe a little more spaghetti or more twists with rails going long distance because of larger bodies of water if landfill cost stays the same.   IMHO this is a positive.
  -Possible late game might be too hard?  We played up to 60% evo.   But from what I have seen when opening a large cave there is usually a narrow passage and most of the the biters stay confined to the cave, so I think it will be ok.  

Some things that might need to be watched for possible adjustment after a few games (they seem ok for now):
  -May have to lower experience penalty on death if players aren't able to progress through market levels sufficiently due to higher death rate.  
  -Amount of increase in evo for void cleared (large caves clear void pretty quick)
  -Landfill cost

Other potential future improvements: 
  -Spawn fish in treasure chests?  This would be nice to help a little for the longer battles in large caves.
  -A way to add some alien spawners to the caves?   (may be too much though in conjunction with hail-hydra
  -Diggy only leaves rock-large behind inside cave when it self collapses.   Would like to add other rock types as well.

BlueRock2:
FYI.  For anyone that asks about single player, I would recommend in the config changing the line:
large_room_minimum_distance = 128  to some unreachable distance such as 10000

PR #1043, opened 2020-03-24
Fix poll

BlueRock2:
Fix poll in shared RedMew GUI.   One of the recent Factorio updates appears to have done away with the sprite 'utility/remove', so replacing it with 'utility/trash'.


SimonFlapse (approved):
LGTM

theorangeangle:
@SimonFlapse i think travis broke again

PR #1047, opened 2020-04-18
Added maze_with_rooms map

Ogg-Technologies:
Added a new map which makes everything void except a random maze
Randomly sized rooms are placed around the maze

### DIFF
--- a/map_gen/maps/maze_with_rooms.lua
+++ b/map_gen/maps/maze_with_rooms.lua
@@ -1,3 +1,7 @@
+local Global = require 'utils.global'
+local RS = require 'map_gen.shared.redmew_surface'
+local b = require 'map_gen.shared.builders'
+
 -- Height and width should always be odd numbers
 local maze_width = 53
 local maze_height = 53
@@ -268,8 +272,8 @@ end
 
 -- Goes through the map and finds dead ends (tiles with 3 walls around them) and fills them in
 -- does this recursivly until every tile has at least 2 ground neighbours
-local function remove_dead_ends(map)
-    function get_neighbours_with_ground(x, y)
+local function remove_all_dead_ends(map)
+    local function get_neighbours_with_ground(x, y)
         local neighbours_with_ground = {}
         for _, dir in pairs(dirs) do
             local xx = x + dir.x / 2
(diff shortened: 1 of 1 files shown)


map_gen/maps/maze_with_rooms.lua
```diff
@@ -0,0 +1,354 @@
...
+        local connector = connectors[random(#connectors)]
+        local pos = connector.pos
+        map[pos.y][pos.x] = true
+    end
+end
+
+-- Goes through the map and finds dead ends (tiles with 3 walls around them) and fills them in
+-- does this recursivly until every tile has at least 2 ground neighbours
+local function remove_dead_ends(map)
+    function get_neighbours_with_ground(x, y)
```
grilledham: missing local

map_gen/maps/maze_with_rooms.lua
```diff
@@ -0,0 +1,354 @@
...
+        for _, dir in pairs(dirs) do
+            local xx = x + dir.x / 2
+            local yy = y + dir.y / 2
+            if map[yy] and map[yy][xx] then
+                neighbours_with_ground[#neighbours_with_ground + 1] = {x = xx, y = yy}
+            end
+        end
+        return neighbours_with_ground
+    end
+    function remove_dead_end(x, y)
```
grilledham: missing local.
Also maybe this function should have a different name to the containing function, maybe remove_dead_end_inner, naming is hard.

map_gen/maps/maze_with_rooms.lua
```diff
@@ -0,0 +1,354 @@
...
+create_map()
+print_map(map)
+]]
+--[[
+    Uncomment below to run in factorio
+]]
+local Global = require 'utils.global'
+local RS = require 'map_gen.shared.redmew_surface'
+local b = require 'map_gen.shared.builders'
+local Event = require 'utils.event'
```
grilledham: We normally put all our requires at the top of the file, so we can easily see the dependencies.

grilledham (approved):
This looks good to me.
If you are interested in micro optimisations I can leave a few more suggestions.

map_gen/maps/maze_with_rooms.lua
Ogg-Technologies: Good catch

map_gen/maps/maze_with_rooms.lua
Ogg-Technologies: I don't really like adding inner after it. The outer function (remove_dead_ends) loops through all tiles that could be dead ends and then calls remove_dead_end on them. But I agree that having such similar names is confusing. Maybe the outer could be remove_all_dead_ends and inner could be fill_dead_end

map_gen/maps/maze_with_rooms.lua
Ogg-Technologies: Yeah, the reason I didn't was that when I tested it I wanted to run the file separatly and print out the map. So I put all factorio related stuff at the same place so I could comment it out when testing. But I can move it now since it works.

grilledham:
What if any are the changes to this map since we last played it on the Redmew server?

Ogg-Technologies:
I haven't made any changes since we played it on Redmew. This should be the same code. Although I sent the code to Jayefuu who set up the map and may have tweaked some constants but from what I know this is the same code.

grilledham:
Ok, we should merge this the. @Ogg-Technologies is that ok with you?

Ogg-Technologies:
Yeah, merge it

grilledham:
Cool, thanks for your contribution. 👍

PR #1049, opened 2020-04-26
Artillery kills cause bots to spawn and shoot across the map.

Zahir-Khan:
- If player Arty kills enemy arty about 60 bots spawn (30 defenders + 30 destroyers)
Typically players kill 4 enemy arty so about 240 bots will spawn.

- If player Arty kills any enemy turret spawn about 16 bots (8 defenders + 8 destroyers) 
Typically arty automatically kills 6 enemy turrets so 96 bots will be spawned that 

Tested on a typical end game crash site map haven't noticed any performance issues. (I don't have a good PC)

I feel it is a good addition to the late game. Players will have to fortify around the arty turrets and make sure the damage is repaired while before the arty's were just placed and fed ammo.

### DIFF
--- a/map_gen/maps/crash_site/entity_died_events.lua
+++ b/map_gen/maps/crash_site/entity_died_events.lua
@@ -283,7 +283,14 @@ end
 local bot_spawn_whitelist = {
     ['gun-turret'] = true,
     ['laser-turret'] = true,
-    ['flamethrower-turret'] = true
+    ['flamethrower-turret'] = true,
+    ['artillery-turret'] = true
+}
+
+local bot_cause_whitelist = {
+    ['character'] = true,
+    ['artillery-turret'] = true,
+    ['artillery-wagon'] = true,
 }
 
 local function do_bot_spawn(entity_name, entity, event)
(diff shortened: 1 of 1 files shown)


grilledham:
I like the idea that when artillery kills a turret the bots target the artillery.

I'm not sure if I agree that how the turret is killed should change the number of bots that spawn. That seems odd to me. I'd prefer the bot spawn rate to be determined by evolution, the outpost the turret is in or distance from spawn.

Zahir-Khan:
I had to make balancing decisions based on these points

- Combat bot health is too low and constant. 
- Their attack params are already linked to evolution.
- Typically Arty starts being used after or near 100% EVO. There is no mass use of arty until the player defenses can handle the influx of large attack waves.
- Player combat is different from arty combat. Arty tend to be protected within bases makes them easy to set up and forget.

If we don't change spawn rates based on the way a turret is killed then there will be a huge discrepancy between player vs arty combat. One of them will be easy or too difficult.

Distance-based spawn is already baked into the game more of a symptom really. There are more defenses in a farther away outpost than a closer one. If arty is left alone they will destroy more defenses more rapidly.  Iv tested all of this it scales beautifully during the end game as well as during early arty stages of the game.

grilledham:
Those are some pretty good points. I still think it's weird that how you kill a turret changes the bot spawn, but maybe it's just me 😄 .

Either way we should test this. You can either make a save file for next crashsite, or we can merge this and let it happen next restart. Your choice.

grilledham (approved):

PR #1050, opened 2020-04-29
Minor fix in Crashsite entity_died_events.lua

Zahir-Khan:
event parameter added to do_spawn_entity

Not sure why hydra is not working. Needs testing.


grilledham:
Look good to me.

Guess I forgot to test spawn on player buildings death 😄

PR #1052, opened 2020-05-08
Create manhattan.lua

Claude477:
Manhattan-map with crashsite.

### DIFF
--- a/map_gen/maps/crash_site/presets/manhattan.lua
+++ b/map_gen/maps/crash_site/presets/manhattan.lua
@@ -1,0 +1,75 @@
+local b = require 'map_gen.shared.builders'
+local MGSP = require 'resources.map_gen_settings'
+
+local type = type
+local water_tiles = b.water_tiles
+local path_tiles = b.path_tiles
+
+local pic = require 'map_gen.data.presets.manhattan'
+local world_map = b.picture(pic)
+local degrees = require "utils.math".degrees
+
+local x_offset, y_offset = 200, 1337
+world_map = b.translate(world_map, x_offset, y_offset)
+world_map = b.rotate(world_map, degrees(-270))
+
+local scale = 2
+local height = 8000 * scale
+local width = 8000 * scale
+
+world_map = b.scale(world_map, scale)
+
+
+local bounds = b.rectangle(width, height)
+bounds = b.translate(bounds, x_offset * scale, y_offset * scale)
+
+local config = {
+    scenario_name = 'crashsite-manhattan',
+    map_gen_settings = {
+        MGSP.starting_area_very_low,
+        MGSP.ore_oil_none,
+        MGSP.enemy_none,
+        MGSP.cliff_none
+    },
+    grid_number_of_blocks = 15,
+    mini_grid_number_of_blocks = 29,
+    bounds_shape = bounds
+}
+
+local Scenario = require 'map_gen.maps.crash_site.scenario'
+local crashsite = Scenario.init(config)
+
+local function get_tile_name(tile)
+    if type(tile) == 'table' then
+        return tile.tile
+    else
+        return tile
+    end
+end
+
+local function map(x, y, world)
+    local tile = world_map(x, y, world)
+    if not tile then
+        return tile
+    end
+
+    local world_tile_name = get_tile_name(tile)
+    if not world_tile_name or water_tiles[world_tile_name] then
+        return tile
+    end
+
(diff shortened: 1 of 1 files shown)


map_gen/maps/crash_site/presets/manhattan.lua
```diff
@@ -0,0 +1,75 @@
...
+local world_map = b.picture(pic)
+local degrees = require "utils.math".degrees
+
+local x_offset, y_offset = 200, 1337
+world_map = b.translate(world_map, x_offset, y_offset)
+world_map = b.rotate(world_map, degrees(-270))
+
+local scale = 2
+local height = 8000 * scale
+local width = 8000 * scale
```
grilledham, line 18: Why 8000 for the height and width?
In the world preset  these numbers come from the develop/map_gen/data/presets/world-map height and width values.
Same question for the other two pull requests.

map_gen/maps/crash_site/presets/manhattan.lua
```diff
@@ -0,0 +1,75 @@
...
+local config = {
+    scenario_name = 'crashsite-manhattan',
+    map_gen_settings = {
+        MGSP.starting_area_very_low,
+        MGSP.ore_oil_none,
+        MGSP.enemy_none,
+        MGSP.cliff_none
+    },
+    grid_number_of_blocks = 15,
+    mini_grid_number_of_blocks = 29,
```
grilledham, line 35: We might be able to use smaller numbers for grid_number_of_blocks and mini_grid_number_of_blocks depending on map size.
If I remember correctly a good value for grid_number_of_blocks is floor(map_size/180)
and for mini_grid_number_of_blocks floor(map_size/96)
Where map size is the larger of width or height after scaling.

Same for the other two pull requests

grilledham (commented):
Some suggestions for change, but they would need to be tested in cases I'm mistaken.

map_gen/maps/crash_site/presets/manhattan.lua
Claude477: normal map meybe, too small so i make it bigger

map_gen/maps/crash_site/presets/manhattan.lua
Claude477: ok that one i dont know about it i just copyed in crashsite default-configs

map_gen/maps/crash_site/presets/manhattan.lua
Claude477: height = 2979,
width = 898 was original settings so yeah is can make put 2x larger size..

Claude477: height=5958
width = 1796 with scale 2

grilledham: Scaling the map is fine, but these values and used to make the bounds_shape which determines where outposts can be made, it shouldn't need to be bigger than the map itself.

Claude477: local x_offset, y_offset = 200, 1337 if change it need reset default reswap location ohter place..

Claude477: ahh i see now 5958x1796(map) and bound size(16000x16000) yeah bound size can make as map size sound good..

Claude477: local scale = 2
local height = 3000 * scale
local width = 1000 * scale sound good solution ?

Claude477: as bound size ?

Claude477: image
 hmm need still fix it T_T

Claude477: need make x location 0->8000 bound size..

Claude477: ok trying fix bound size..

map_gen/maps/crash_site/presets/manhattan.lua
Claude477: ok found out in crashsite no need that blocks

map_gen/maps/crash_site/presets/manhattan.lua
Claude477: image
found out that 8000x8000 is best in that map for fill all map with bitters tryed 3k x 5958 geted by image

PR #1053, opened 2020-05-08
Create venice.lua

Claude477:
Crashsite with venice.lua map

### DIFF
--- a/map_gen/maps/crash_site/presets/venice.lua
+++ b/map_gen/maps/crash_site/presets/venice.lua
@@ -1,0 +1,75 @@
+local b = require 'map_gen.shared.builders'
+local MGSP = require 'resources.map_gen_settings'
+
+local type = type
+local water_tiles = b.water_tiles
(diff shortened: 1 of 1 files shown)


Claude477:
local scale = 1
local height = 4000 * scale
local width = 4000 * scale

sizes was original
height = 2903,
width = 3833,

theorangeangle:
@grilledham should we make a subfolder for all the crash site variants?

grilledham:
Assuming you are talking about the files directly in the maps folder*, probably not. I think the intention was to have a flat list so that map_selection is simpler.

(* The core crashsite files already are in a separate folder.)

theorangeangle:
Okay - are we good to merge then?
  

  
  
  

  
  
>

grilledham:
I think I still need to check this.
I might also make a change to crashsite scenario that makes it easier to make these variants without inadvertently adding hard to spot errors.

PR #1058, opened 2020-05-10
Update redmew.cfg

theorangeangle:
added a description that people see when loading the scenario in SP


SimonFlapse (approved):
I'm assuming this is in the correct location to be found and used by factorio.

PR #1061, opened 2020-05-27
Diggy updates for Factorio .18.27

BlueRock2:
Proposed fixes for Diggy to work with Factorio .18.27.    No issues that I could tell in single player testing.   Comments/feedback appreciated.

### DIFF
--- a/map_gen/maps/diggy/feature/alien_spawner.lua
+++ b/map_gen/maps/diggy/feature/alien_spawner.lua
@@ -76,7 +76,7 @@ local do_alien_mining = Token.register(function(params)
         for rock_index = rock_count, 1, -1 do
             local rock = rocks[rock_index]
             destroy_rock(create_particle, particle_count, rock.position)
-            rock.die(force)
+            rock.die("enemy")
         end
     end
 
--- a/map_gen/maps/diggy/feature/alien_spawner.lua
+++ b/map_gen/maps/diggy/feature/alien_spawner.lua
@@ -22,7 +22,6 @@ local get_aliens = AlienEvolutionProgress.get_aliens
 local create_spawner_request = AlienEvolutionProgress.create_spawner_request
 local set_timeout_in_ticks = Task.set_timeout_in_ticks
 local destroy_rock = CreateParticles.destroy_rock
-local is_diggy_rock = Template.is_diggy_rock
 
 -- this
 local AlienSpawner = {}
@@ -77,7 +76,8 @@ local do_alien_mining = Token.register(function(params)
         for rock_index = rock_count, 1, -1 do
             local rock = rocks[rock_index]
             destroy_rock(create_particle, particle_count, rock.position)
-            rock.die("enemy")
+            -- diggy_hole.lua will handle removing loot from entities that died with force set to "remove_loot"
+            rock.die("remove_loot")
         end
     end
 
@@ -169,12 +169,6 @@ function AlienSpawner.register(cfg)
         global.config.hail_hydra.enabled = true
         global.config.hail_hydra.hydras = hail_hydra
     end
-
-    Event.add(defines.events.on_entity_died, function (event)
-        if is_diggy_rock(event.entity.name) and event.force.name == "enemy" and event.loot then
-            event.loot.clear()
-        end
-    end)
 
     Event.add(Template.events.on_void_removed, function (event)
         local force = game.forces.enemy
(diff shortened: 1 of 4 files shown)


map_gen/maps/diggy/feature/diggy_hole.lua
```diff
@@ -181,9 +180,7 @@ function DiggyHole.register(cfg)
         if not is_diggy_rock(name) then
             return
         end
-
-        raise_event(defines.events.on_entity_died, {entity = entity, cause = event.cause, force = event.force})
-        entity.destroy()
+        entity.die(force)
```
Soggs: Should be `entity.die(event.force, event.cause)` here.

map_gen/maps/diggy/feature/simple_room_generator.lua
```diff
@@ -35,8 +34,7 @@ local do_mine = Token.register(function(params)
 
     for i = rock_count, 1, -1 do
         local rock = rocks[i]
-        raise_event(on_entity_died, {entity = rock})
-        rock.destroy()
+        rock.die(force)
```
Soggs: Here you might want to hard code the player or the neutral force into it. Which would be either `rock.die("player")` or `rock.die("neutral")`.

map_gen/maps/diggy/feature/alien_spawner.lua
```diff
@@ -76,9 +75,8 @@ local do_alien_mining = Token.register(function(params)
 
         for rock_index = rock_count, 1, -1 do
             local rock = rocks[rock_index]
-            raise_event(defines.events.on_entity_died, {entity = rock})
             destroy_rock(create_particle, particle_count, rock.position)
-            rock.destroy()
+            rock.die(force)
```
Soggs: The appropriate force here would be enemy. So `rock.die("enemy")`.

Soggs (commented):
https://lua-api.factorio.com/latest/LuaEntity.html#LuaEntity.die

The `die` call requires a force definition. In the three cases it looks like the field `force` is just nil. Which could initially work but could probably error on us later, so better be safe than sorry. :D

map_gen/maps/diggy/feature/simple_room_generator.lua
```diff
@@ -69,6 +67,12 @@ function SimpleRoomGenerator.register(config)
         return perlin_noise(x * noise_variance, y * noise_variance, seed)
     end
 
+    Event.add(defines.events.on_entity_died, function (event)
+        if (event.loot and event.loot.valid) then
+            event.loot.clear()
+        end
+    end, {{filter = "type", type = "simple-entity"}})
+
```
SimonFlapse: IIRC, filters work on a per mod basis (The entire redmew framework would be a single mod), this means that we can't use filters in our event handler because it would prevent other event handlers on the same event from triggering.

My point is that 
1. I don't believe you can give the filter as a parameter for our Event.add method
2. Even if you could it would end up conflicting with other event handlers for the same event, that rely on another filter.
3. This event handler removes all loot from every entity, whether it's a simple-entity or not.

I could be wrong.

SimonFlapse (commented):
From discord:
I'm not sure the loot should be cleared only when it is the neutral edit:(or enemy) force that did the killing. I think that it should apply for all cases where we previously used destroy()
~~You might be able to clear the loot after the .die() call~~

For the robot mining, you could create a custom force just for the die method so you know the cause is a robot mining a diggy rock
Otherwise I like you thinking with preserving functionality even if some of the features are turned off

SimonFlarup (approved):

SimonFlapse (approved):
Assuming the changes have been tested it LGTM.

I'd recommend starting a fresh Diggy server with these changes and see how it reacts, merge if no issues shows within a couple of hours

BlueRock2:
Updated so that the diggy rocks that die will not drop loot in any case (due to reactor explosion generating lots of stone).    I also re-introduced entity.destroy() in a few cases where it makes sense like opening rooms (as die was creating lots of rock death animations and slowing things down).     Did this this for rocks reaching 0 health by damage as well as it signifigantly improved performance in case a reactor blows up (and is the way it was in .18.26).     I just had to generate and catch destroy events in addition in diggy_hole and cave_collapse.    Lastly if a rock is set to die (like in bot mining currently) ensured scripted 'death particles' that were intended for entity destroyed don't run since die creates its own death particles.

BlueRock2:
Tested latest commits on S9 today (3 June) for more than four hours with no issues found.   Both this PR #1061 as well as the the bot-mining update  #1064.

SimonFlapse (approved):
LGTM, testing indicates we are ready to merge

PR #1063, opened 2020-05-30
Remove evolution_scale_formula.

grilledham:
0.18.28 doesn't allow functions in global so we are removing the evolution_scale_formula function from being stored in global.


SimonFlapse (approved):
I'd assume this works since the formula never changes, so it shouldn't be a desync risk.

PR #1064, opened 2020-05-31
Recommended Diggy bot mining update for Factorio .18.27

BlueRock2:
This update is not strictly required for Factorio .18.27
However, .18.27 adds a bot deconstruction sound (no sound previously) and with the way that Diggy currently handles bot 'mining', each bot is triggering the sound 60 times a second.    This creates a very loud and annoying buzzing sound.
The proposed update changes bot behavior to only mark rocks again for deconstruction every x ticks instead of every tick.    This results in a softer, more pleasing pulsed mining sound.
This update also seems to improve bot mining predictability a little as well. 
If one has equal or more bots than the number of rocks selected bots will focus on their own rocks.  
Without this update even if one has more bots than rocks selected, some bots will jump around between different rocks (not exactly sure why).    Initially marking this as work in progress to get feedback and tweak frequency and damage for early and late game balance.

Special note:   This update would need to be merged after PR #1061 since it contains changes in diggy_hole.lua that are part of that PR

### DIFF
--- a/map_gen/maps/diggy/config.lua
+++ b/map_gen/maps/diggy/config.lua
@@ -56,8 +56,10 @@ local Config = {
         -- core feature
         diggy_hole = {
             enabled = true,
-            -- initial damage per tick it damages a rock to mine, can be enhanced by robot_damage_per_mining_prod_level
-            robot_initial_mining_damage = 4,
+            -- delay in ticks between robot mining rock and rock being marked again for deconstruction
+            robot_mining_delay = 10,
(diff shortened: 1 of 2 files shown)


BlueRock2:
Tested latest commits on S9 today (3 June) for more than four hours with no issues found. Both PR #1061 (for required 0.18.27 changes) as well as this bot-mining update #1064 update together.

SimonFlapse (approved):
Nice work. LGTM

Issue #878, opened 2019-03-17
Error in hail_hydra


mheguy:
(Server time: 2019-03-02 05:12:36)
Time of error: 9 hours 35 minutes
RedMew version: Time of map launch: 2019-03-01 19:33:13 UTC
Target must be entity with force.
stack traceback:
	...o/3/temp/currently-playing/map_gen/shared/hail_hydra.lua:147: in function <...o/3/temp/currently-playing/map_gen/shared/hail_hydra.lua:73>
	[C]: in function 'pcall'
	/factorio/3/temp/currently-playing/utils/event_core.lua:32: in function 'call_handlers'
	/factorio/3/temp/currently-playing/utils/event_core.lua:43: in function </factorio/3/temp/currently-playing/utils/event_core.lua:41>

grilledham:
I don't see any mention of this in the logs for the diggy server, so I'm going to assume this was fixed.

PR #1072, opened 2020-08-08
Various Diggy enhancements and updates

BlueRock2:
*Corrects linter error in Diggy_cave collapse line 298 (caused by bug in Diggy update for .18.27)

*Added reactors to the support_beam_entities list to allow for more nuclear plant build options in Diggy (received 7 thumbs up and only one thumbs down in suggestions channel)

*Provide example of proper use of built in Diggy debugging command "diggy-clear-void" within comments

*Changed landfill tile in Diggy to the landfill graphic introduced in Factorio 0.17.10.   Up until now Diggy has suppressed the landfill texture because it used to be the same as grass-1 which didn't look good in Diggy.   The new graphic looks great in Diggy.

*New Feature:   Tank tunelling/buff   When tanks advance against rocks or damage rocks with non-fire based weapons the rocks will be destroyed when under 1500 health.   This makes tanks a viable (but not OP) option prior to bots for clearing rocks.   Tanks will be damaged by and not instantly destroyed by ceiling collapses.

*Minor ore generation tweaks based on player feedback over several months to help balance ore rations and densities from early to late game.
  -Changed iron and copper vein purity from 86% to 90%. Helps lessen the huge glut of coal and stone late game.
  -Start with a little less copper density near the origin but grow in copper richness a little faster with distance. 
  -Slightly increase coal density near origin, but increase coal density less with distance
  -Slightly increase stone density near origin, but increase stone density less with distance
  -Increase fragmented ore vein ore density a little near origin, but significantly lessen density growth with distance
  -Increase minimum distance of uranium from 86 to 128 tiles from origin
  -Kept the number of oil patches the same (so they are still easy to find in single player) but reduced the yield of individual oil patches by about 30%.  The number of oil patches was increased about 3x in May last year to make oil easier to find in single player.   However this also made the total yield 3x as well and too easy overall.

Have been slowly adding and testing the changes above to ensure no problems or issues over the past four weeks.


map_gen/maps/diggy/feature/diggy_hole.lua
```diff
@@ -166,7 +166,7 @@ Command.add('diggy-clear-void', {
         end
     end
 
-    Template.insert(game.surfaces[arguments.surface_index], tiles, {})
+    Template.insert(game.surfaces[tonumber(arguments.surface_index)], tiles, {})
```
grilledham: I see this is a debug command so maybe this isn't a concern.

Before if surface_index was a the name of a surface this would've worked, but now it only takes numbers. Maybe that's fine it is called surface_index after all.

tonumber could return nill, which would mean game.surfaces[nil] return nil. Maybe there should be a check that the surface was found?

grilledham (approved):

map_gen/shared/hail_hydra.lua
```diff
@@ -143,8 +143,10 @@ local on_died =
                 local spawned = create_entity({name = hydra_spawn, force = force, position = position})
                 if spawned and spawned.type == 'unit' then
                     spawned.set_command(command)
-                elseif spawned and cause and cause.valid and cause.force then
-                    spawned.shooting_target = cause
+                elseif spawned and cause and cause.valid and cause.force and cause.force.valid then
+                    if cause.name ~= 'tank' then
```
SimonFlapse: Is it necessary to exclude tanks? Since this is a shared module, I'd like to make sure this is absolutely necessary, otherwise it should be configurable.

SimonFlapse (approved):
Looks good to me, we may want to have the new tank feature be optional/configurable.

Please have a look at my question regarding the hail hydra changes.

map_gen/maps/diggy/feature/diggy_hole.lua
BlueRock2: I have cancelled this particular update due to my own user error.   I found that with proper use, I can specify 'redmew' as the surface_index argument when using the command instead of '1' and converting it into a number for the surface index.   I have added a comment to the code with a usage example that may help others in the future.

map_gen/shared/hail_hydra.lua
BlueRock2: After more research, found that entity type 'car' (which includes both cars and tanks) needs to be excluded due to a base game limitation.    This goes all the way back to hail_hydra issue #878 that was thought to have been resolved over a year ago, but is actually a bug that happens in a very specific case.     An enemy needs to have been killed using a tank or a car, and a new alien is spawned as part of hail_hydra, and the enemy unit is of turret type.   In this case an attempt is made to have the enemy unit type of 'turret', target an entity type of 'car' which is not allowed by the base game and throws an error.  I verified this with Bilka in the official Factorio discord 'mod-making' channel.   Made a new commit to update the code.

grilledham:
LGTM 👍

PR #1078, opened 2020-09-02
Container dump

grilledham:


grilledham:
Tested on Terraforming danger ores in multiplayer, seems fine.

PR #1088, opened 2020-09-23
Updated crash site /restart command so that guests and regulars can call it when map is cleared

Jayefuu:
Updated crash site /restart command so that guests and regulars can call it when map is cleared
- Only admin can abort, once the map is cleared.
- Guests and Regular can both call /restart when all enemy units and structures are dead
- Tested on S10 by running this command:
/c local surface=game.player.surface
for key, entity in pairs(surface.find_entities_filtered({force="enemy"})) do
	entity.destroy()
end
- Need to merge into develop to test on live server in case conditions at the end of the game are different
- Thought about adding rocket launched condition but decided against it.

### DIFF
--- a/map_gen/maps/crash_site/commands.lua
+++ b/map_gen/maps/crash_site/commands.lua
@@ -59,10 +59,10 @@ local function restart(args, player)
 
     if Rank.less_than(player.name, Ranks.admin) then
         -- Check enemy count
-        local enemy_count = game.player.surface.count_entities_filtered{force= "enemy"}
+        local enemy_count = game.player.surface.count_entities_filtered{force= "enemy", limit=1}
         
         if (enemy_count ~= 0) then
-            game.player.print('All enemy spawners, worms, buildings, biters and spitters must be cleared for non-admin restart. Current enemy count: '.. enemy_count)
+            game.player.print('All enemy spawners, worms, buildings, biters and spitters must be cleared for non-admin restart.')
             return
         end 
     end
@@ -136,7 +136,7 @@ function Public.control(config)
             description = {'command_description.crash_site_restart'},
             arguments = {'scenario_name'},
             default_values = {scenario_name = default_name},
-            required_rank = Ranks.guest,
+            required_rank = Ranks.auto_trusted,
             allowed_by_server = true
         },
         restart
--- a/map_gen/maps/crash_site/commands.lua
+++ b/map_gen/maps/crash_site/commands.lua
@@ -1,10 +1,12 @@
 local Command = require 'utils.command'
+local Rank = require 'features.rank_system'
 local Task = require 'utils.task'
 local Token = require 'utils.token'
 local Server = require 'features.server'
 local Popup = require 'features.gui.popup'
 local Global = require 'utils.global'
 local Ranks = require 'resources.ranks'
+local Utils = require 'utils.core'
 
 local server_player = {name = '<server>', print = print}
 
@@ -55,6 +57,16 @@ local function restart(args, player)
         return
     end
 
+    if Rank.less_than(player.name, Ranks.admin) then
+        -- Check enemy count
+        local enemy_count = game.player.surface.count_entities_filtered{force= "enemy"}
+        
+        if (enemy_count ~= 0) then
+            game.player.print('All enemy spawners, worms, buildings, biters and spitters must be cleared for non-admin restart. Current enemy count: '.. enemy_count)
+            return
+        end 
+    end
+    
     global_data.restarting = true
 
     double_print('#################-Attention-#################')
@@ -67,7 +79,7 @@ local function restart(args, player)
         end
     end
     print('Abort restart with /abort')
-
+    
     Task.set_timeout_in_ticks(60, callback, {name = player.name, scenario_name = args.scenario_name, state = 10})
 end
 
@@ -124,7 +136,7 @@ function Public.control(config)
             description = {'command_description.crash_site_restart'},
             arguments = {'scenario_name'},
             default_values = {scenario_name = default_name},
-            required_rank = Ranks.admin,
+            required_rank = Ranks.guest,
             allowed_by_server = true
         },
         restart
--- a/map_gen/maps/crash_site/commands.lua
+++ b/map_gen/maps/crash_site/commands.lua
@@ -6,7 +6,6 @@ local Server = require 'features.server'
 local Popup = require 'features.gui.popup'
 local Global = require 'utils.global'
 local Ranks = require 'resources.ranks'
-local Utils = require 'utils.core'
 
 local Public = {}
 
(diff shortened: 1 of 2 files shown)


map_gen/maps/crash_site/commands.lua
```diff
@@ -55,6 +57,16 @@ local function restart(args, player)
         return
     end
 
+    if Rank.less_than(player.name, Ranks.admin) then
+        -- Check enemy count
+        local enemy_count = game.player.surface.count_entities_filtered{force= "enemy"}
+        
+        if (enemy_count ~= 0) then
+            game.player.print('All enemy spawners, worms, buildings, biters and spitters must be cleared for non-admin restart. Current enemy count: '.. enemy_count)
```
grilledham: I wonder if we should drop the `enemy_count` part. People might use this as a way to count remainging biters, which I'm not sure we want. Also if we drop the count we can put `limit = 1` in the `count_entities_filtered` which makes it quicker.

grilledham: Is this meant to be `player.print` instead of `game.player.print`?

map_gen/maps/crash_site/commands.lua
```diff
@@ -55,6 +57,16 @@ local function restart(args, player)
         return
     end
 
+    if Rank.less_than(player.name, Ranks.admin) then
+        -- Check enemy count
+        local enemy_count = game.player.surface.count_entities_filtered{force= "enemy"}
```
grilledham: This will include walls, which might be undesirable, not all walls are capturable. But maybe this isn't a problem?

map_gen/maps/crash_site/commands.lua
```diff
@@ -67,7 +79,7 @@ local function restart(args, player)
         end
     end
     print('Abort restart with /abort')
-
+    
     Task.set_timeout_in_ticks(60, callback, {name = player.name, scenario_name = args.scenario_name, state = 10})
```
grilledham: `scenario_name` could be any scenario name not just crashsite maps. This wasn't a problem when only admins were running the command, but could be now. We don't want troll players starting a diggy map instead.

map_gen/maps/crash_site/commands.lua
```diff
@@ -124,7 +136,7 @@ function Public.control(config)
             description = {'command_description.crash_site_restart'},
             arguments = {'scenario_name'},
             default_values = {scenario_name = default_name},
-            required_rank = Ranks.admin,
+            required_rank = Ranks.guest,
```
grilledham: This should probably be `auto_trusted`. Easily allows us to remove the ability from players who might be miss using it.

grilledham (changes requested):
Non admin restarts cannot be aborted by non admins. Is this intentional? If we do allow non admins to abort, then we have to think about non admins aborting when an admin orders the restart.

Have you done a test run on the server, just as a sanity check?

This is a good start, thanks for taking the time to do this. However the nature of this command where it allows any player to perform an action on the server means we have to be careful with it. So there are still some issues we might need to think about.

grilledham:
I wonder if it's worth putting a map time check (>5 minutes?) in for non admin users, probably not needed?

I bet once we add this command we will get requests to allow restarts on lose conditions, or some kind of vote restart system.

map_gen/maps/crash_site/commands.lua
Jayefuu: Isn't it the same thing?

Jayefuu: Ok. I had it there for debug, I thought it would be useful. But if there's a problem I need to debug I can always print the count with a console command. Will remove the count print.

map_gen/maps/crash_site/commands.lua
Jayefuu: I'm not sure. I did some play testing and used the destroy all command to allow guests and regulars to do the restart. This worked. But we didn't play through to a natural completion, just an artificial one. We should merge it with develop and if needs be debug the enemy_count limit later. I'm sure there are some things that might change. If the /restart command doesn't work for them one of us can hop on and see what entity remains that's stopping the restart.

map_gen/maps/crash_site/commands.lua
Jayefuu: I intentionally left it so that regulars/guests can't abort so that I didn't have to handle it. I can't imagine it will matter if the biters are cleared that someone calls restart. If others have the ability to abort a restart then they might fight over it.

grilledham: `game.player` is the player that ran the command, this is `nil` when run from the server console. Which why we have the `player = player or server_player` above.

Now that I think about this some more, `game.player.surface` wont work from the server console. You will need to do something like:
```lua
...
RS.get_surface()
```

grilledham: I think that is fine then.

map_gen/maps/crash_site/commands.lua
grilledham: I guess we can wait to see if this turns into an actual issue.

map_gen/maps/crash_site/commands.lua
grilledham: Maybe for non admins we set `args.scenario_name` to `config.scenario_name` below. Would mean moving this function into the Public.control function.

Jayefuu:
@grilledham I have just managed to catch the end of a crash site game. There are 544 enemy entities left, mainly walls and power poles. 

I moved it here in case we want to put it on S10 for debugging:
https://redmew.com/saves?directory=public/final&file=2020-09-28_crash_site_finished.zip

In your opinion should we:
a) Make the players clear all this crap, which could be laborious, to get an enemy count of 0 before they're allowed to reset
b) Filter the count_entities_filtered() to only be units, spawners and turrets, which might have performance impacts?
c) Set a limit so they can restart when less than 500 enemy entities on the map. I checked and immediately upon restarting there are already 1000 entities, so this won't allow them to restart straight away if they don't like the mapgen. And that will grow very quickly. So this could be a way to allow them to restart without clearing ALL walls and power poles without the performance hit of b?

Which option do you prefer?

grilledham:
@Jayefuu That is good information to have.

I would do a variation of b).

```lua
local entities_to_check = {
    'biter-spawner',
    ...
    --spawners, biters, spitters, worms, turrets, drones
}

local function map_cleared(force)
    local get_entity_count = force.get_entity_count
    for i = 1, #entities_to_check do
        local name = entities_to_check[i]
        if get_entity_count(name) > 0 then
            return false
        end
    end

    return true
end
```

map_gen/maps/crash_site/commands.lua
```diff
@@ -47,14 +53,46 @@ callback =
     end
 )
 
+local entities_to_check = {
+    'spitter-spawner','biter-spawner',
+    'small-worm-turret', 'medium-worm-turret','big-worm-turret', 'behemoth-worm-turret', 
+    'small-spitter', 'medium-spitter', 'big-spitter', 'behemoth-spitter', 
+    'small-biter', 'medium-biter', 'big-biter', 'behemoth-biter'
```
grilledham: Missing outpost turrets: 'gun-turret', 'laser-turret', 'artillery-turret', 'flamethrower-turret'.
Also maybe combat-robot: 'defender', 'destroyer', 'distractor'

map_gen/maps/crash_site/commands.lua
```diff
@@ -47,14 +53,46 @@ callback =
...
 local function restart(args, player)
     player = player or server_player
+    local sanitised_scenario = args.scenario_name
 
     if global_data.restarting then
         player.print('Restart already in progress')
         return
     end
 
+    if Rank.less_than(player.name, Ranks.admin) then
```
grilledham: I think this should be something like.
```suggestion
    if player ~= server_player and Rank.less_than(player.name, Ranks.admin) then
```
For the case of running this from the server console.

map_gen/maps/crash_site/commands.lua
```diff
@@ -47,14 +53,46 @@ callback =
...
+        -- Check enemy count      
+        if not map_cleared() then
+            game.player.print('All enemy spawners, worms, buildings, biters and spitters must be cleared for non-admin restart.')
```
grilledham: Could be `player` instead of `game.player` for consistency with above.

grilledham (changes requested):
A few minor changes and I think this is ready. 😃

grilledham:
You will also have to fix the luacheck warnings.

grilledham (approved):

PR #1097, opened 2020-10-22
Added Discord notification for "@crash site" when map restarts

Jayefuu:
Added discord notification for the @crash site role when crash site is restarted using the /restart command. Pings users with @Crash Site role and prints the map time to add to competitiveness.

### DIFF
--- a/map_gen/maps/crash_site/commands.lua
+++ b/map_gen/maps/crash_site/commands.lua
@@ -6,6 +6,7 @@ local Server = require 'features.server'
 local Popup = require 'features.gui.popup'
 local Global = require 'utils.global'
 local Ranks = require 'resources.ranks'
+local Core = require 'utils.core'
 
 local Public = {}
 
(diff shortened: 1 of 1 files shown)


map_gen/maps/crash_site/commands.lua
```diff
@@ -42,6 +42,10 @@ callback =
             global_data.restarting = nil
             return
         elseif state == 1 then
+            local game_seconds = math.floor(game.ticks_played / 60) -- should probably account for game speed changes, but we rarely (if ever?) change crash site performance scale. This is good enough?
+            local game_hours = math.floor(game_seconds / 3600)
+            local game_min_remainder = math.floor((game_seconds - (game_hours * 3600)) / 60)
```
grilledham: Does this do what you want <https://github.com/Refactorio/RedMew/blob/develop/utils/core.lua#L136>?

map_gen/maps/crash_site/commands.lua
```diff
@@ -42,6 +42,10 @@ callback =
...
+            Server.to_discord_raw('<@&762441731194748958> Crash Site has just restarted!! Previous map lasted: ' .. game_hours .. ' hours, ' .. game_min_remainder .. ' minutes')
```
grilledham: Add markdown to make it **bold**?

```suggestion
            Server.to_discord_raw('<@&762441731194748958> **Crash Site has just restarted!! Previous map lasted: ' .. game_hours .. ' hours, ' .. game_min_remainder .. ' minutes**')
```

grilledham (approved):
Some suggestions, but this looks fine to me.

map_gen/maps/crash_site/commands.lua
grilledham: It might also be woth having a local variable for `local crashsite_role ='<@&762441731194748958>'` just to explain what that number is.

Jayefuu: All good suggestions, thank you. I'll make the changes and test tonight.

Jayefuu: I think I'll also try work out how to put the notification in #map-update not #s7 so that all the map pings stay in one place and people don't have to scroll through the server channel.

map_gen/maps/crash_site/commands.lua
```diff
@@ -123,6 +128,7 @@ local function restart(args, player)
     end
     print('Abort restart with /abort')
 
+    --Server.to_discord_raw('<@&762441731194748958> Crash site is restarting!')
```
grilledham: After our discssion about not being able to send the pings to the map update channel (yet), I assume you will uncomment this.

grilledham (approved):

map_gen/maps/crash_site/commands.lua
Jayefuu: No, good catch. The first time I wrote this I had the ping go out when the restart command was called, but later moved it to the same place as the popup so that players didn't get a notification if /restart then /abort was used. I've now deleted that commented line since the to_discord_raw is now on line 49.

PR #1098, opened 2020-10-29
Begin logging crash site game time to scenario data

Jayefuu:
Added lines to store the unix time the restart happens as well as how many ticks the game lasted. Starting storing these now so I have some interesting data to use in a week or two.

### DIFF
--- a/map_gen/maps/crash_site/commands.lua
+++ b/map_gen/maps/crash_site/commands.lua
@@ -47,11 +47,11 @@ callback =
             local discord_crashsite_role = '<@&762441731194748958>' -- @crash_site
             --local discord_crashsite_role = '<@&593534612051984431>' -- @test
             Server.to_discord_raw(discord_crashsite_role .. ' **Crash Site has just restarted! Previous map lasted: ' .. time_string .. '!**')
-                
+
             local unix_time = Server.get_current_time()
             local game_time = game.ticks_played
             Server.set_data('crash_site_data', tostring(unix_time),  game_time) -- Store the server unix time as key and total game ticks in the Scenario Data
-                
+
             Popup.all('\nServer restarting!\nInitiated by ' .. data.name .. '\n')
         end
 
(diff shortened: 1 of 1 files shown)


map_gen/maps/crash_site/commands.lua
```diff
@@ -47,6 +47,11 @@ callback =
             local discord_crashsite_role = '<@&762441731194748958>' -- @crash_site
             --local discord_crashsite_role = '<@&593534612051984431>' -- @test
             Server.to_discord_raw(discord_crashsite_role .. ' **Crash Site has just restarted! Previous map lasted: ' .. time_string .. '!**')
+
+            local unix_time = Server.get_current_time()
+            local game_time = game.ticks_played
+            Server.set_data('crash_site_data', tostring(unix_time),  game_time) -- Store the server unix time as key and total game ticks in the Scenario Data
```
grilledham, line 53: 'crashsite_data' would be more consistent with the scenario name, but this is minor so it's ok.

grilledham, line 53: Merely a suggestion, but <https://github.com/Refactorio/RedMew/blob/develop/utils/timestamp.lua#L147> might make the unix time easier to read. Up to you.

grilledham (approved):

map_gen/maps/crash_site/commands.lua
Jayefuu: Is it crashsite or crash site? It changes depending on where you look, I had settled on always calling it crash site. Hence the _.

Jayefuu: I'll keep it as a long int. Easier for me to do math on later I think. I'm ok with it not being human readable if it makes it easier later.

grilledham: Yeah you are right, I'm not very consistent with myself 😄. As players don't see this it doens't really matter.

grilledham: That's a good enough reason.

PR #1099, opened 2020-11-04
Crash Site: Enemy combat robots now explode when they timeout

Jayefuu:
Players were using remotely located artillery wagons in lakes to cheesily clear the map with no repercussions. Swarms of destroyer capsules would hover over the artillery but not target it. They would timeout and  the player could then refill the wagon with ammo.

I've changed the entity_died_events file to events.lua since it now contains the events on_entity_died, on_player_died and on_combat_robot_expired.

I have added the on_combat_robot_expired event so that when the robots timeout they explode, damaging nearby entities and stopping players from not defending their artillery wagons.

### DIFF
--- a/map_gen/maps/crash_site/events.lua
+++ b/map_gen/maps/crash_site/events.lua
@@ -365,6 +365,7 @@ local function do_bot_spawn(entity_name, entity, event)
     end
 end
 
+-- Drops coins when biter/spitter spawners and worms are killed
 local function do_coin_drop(entity_name, entity)
     local position = entity.position
     local bounds = entity_drop_amount[entity_name]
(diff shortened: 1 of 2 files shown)


map_gen/maps/crash_site/events.lua
```diff
@@ -478,3 +479,17 @@ Event.add(
...
 )
+
+Event.add(
+    defines.events.on_combat_robot_expired,
+    function(event)
+
+        local entity = event.robot
+        local position = entity.position
+        if entity.force.name == 'enemy' then
+            entity.surface.create_entity{name = "cluster-grenade", position=position, target=position, speed=1}
```
grilledham, line 490: I wonder if a cluster grenade is a bit much, but I'll leave that for you to decide.

grilledham (approved):

map_gen/maps/crash_site/events.lua
Jayefuu: I decided on that because the blast radius is larger and sometimes the bots get bored and drift away a little. Using a projectile with a larger blast radius seemed like a better solution than searching for a nearby entity to set as the projectile target. It should be fine since with this change players will be incentivised to defend their artillery wagons properly.

Please merge if this is ok.

PR #1100, opened 2020-11-04
Spidertron artillery collision fix

Jayefuu:
Adapted code from Damageable spider leg mod and added it to events.lua. 

Spidertrons now take damage from enemy artillery since before they were immune.

### DIFF
--- a/map_gen/maps/crash_site/events.lua
+++ b/map_gen/maps/crash_site/events.lua
@@ -450,6 +450,42 @@ local function do_spawn_entity(entity_name, entity, event)
     end
 end
 
+-- damageable-spider-leg lua module
+-- adapted from mod: https://mods.factorio.com/mod/damageable-spider-leg
+-- by x2605
+
+local damageable_spider_leg = {}
+local debugging = false
+local damage_modifier = 1/6
+
+local register_spider =  function(entity)
+  if not entity then return end
+  if not entity.valid then return end
+  if entity.type ~= 'spider-vehicle' then return end
+  if not global.damageable_spider_leg then
+    global.damageable_spider_leg = {}
+  end
+  local data = {}
+  local spider = entity
+  local surface = spider.surface
+  local legs = surface.find_entities_filtered{
+    position = spider.position,
+    radius = 10,
+    type = 'spider-leg'
+  }
+  for _, leg in pairs(legs) do
+    if not leg.destructible then
+      leg.destructible = true
+      data[#data + 1] = leg
+    end
+  end
+  global.damageable_spider_leg[tostring(spider.unit_number)] = data
+  script.register_on_entity_destroyed(spider)
+  if debugging then
+    game.print{"",'legs=',#data,' ',spider.name,spider.unit_number}
+  end
+end
+
 Event.add(
     defines.events.on_entity_died,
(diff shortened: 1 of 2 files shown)


Jayefuu:
Could do with discussing the failed checks before I fix them @grilledham.

grilledham (commented):
As discussed I think we will need to use a different solution rather than using the `on_entity_damaged` event.

With regards to the luacheck:
 - `accessing undefined field register_on_entity_destroyed of global script`
We would need to add the event name to <https://github.com/Refactorio/RedMew/blob/develop/.luacheckrc#L1004>
 -  `value assigned to variable unit_number is unused`
Seems like a code style choice we have made to not explicitly assign to null if that value isn't used (In lua values are by default nil anyway).
 - `unused variable damageable_spider_leg` 
The local value `damageable_spider_leg` line 457 doesn't seem to be used, maybe you didn't mean to use `global.damageable_spider_leg` everywhere. Adding `damageable_spider_leg` to `Global.register` on line 22 would probably make more sense.

Jayefuu:
The problem is that spidertrons don't have a collision mask for their bodies and so take no damage from artillery which hits the surface not the spidertron, I think.

I also spent some time looking for an event for when turrets fire or something, if I could hook into that I could remove the projectile when targetting spidertrons and replace it with a laser or something. Or directly apply damage to the spidertron when it's targetted.

I could maybe write a function that creates a laser projectile from an empty artillery turret to target entities within a range. If it just ran every 5 seconds and only targetted spidertrons it could be an interesting "new weapon" without having to use mods.

Jayefuu:
Another idea I had was to make the spidertron non-mineable. At the moment, since it's invulnerable to artillery players just ignore the arty while in the spidertron. But if their spider goes back to base the artillery continues to fire on them even though they should be out of range. This has lead to a meta where if you return to base with your spidertron you HAVE to pick it up otherwise the arty destroys the base. By making it non-mineable players will have to park their spider somewhere it won't damage the base and somewhat offset the fact they're invulnterable to arty.,

grilledham:
Wait, artillery targets spidertrons? I thought artillery didn't target non structure types. Has this been changed, i.e. in vanilla factorio does artillery target cars/ tanks, players, etc?

Crashsite has some special code to fake artillery targeting players and vehicles, maybe we could hook into that and do something for spidertrons 
<https://github.com/Refactorio/RedMew/blob/develop/map_gen/maps/crash_site/outpost_builder.lua#L1112>

Jayefuu:
It definitely targets spidertrons. I'll take a look at that code, cheers.

Jayefuu:
@grilledham I was told today that we're missing a trick with this. Apparently if we use event filters we can stop the event triggering and make it more performant than entering the event every time an entity is damaged then returning if not spidertron leg.

Please can you advise me how to add this? https://lua-api.factorio.com/latest/Event-Filters.html
And how I can test the performance.

I don't understand how it will interact with the changes you made to events within the scenario.
`script.set_event_filter(defines.events.on_entity_damaged, {{filter = "name", name = "spidertron-leg"}})`

Jayefuu:
Closing in favour of a more performant solution, PR #1113

PR #1101, opened 2020-11-04
Added a /spy command to allow players to reveal map

Jayefuu:
Might need some extra checks to make sure I've parsed the arguments safely.

### DIFF
--- a/locale/en/redmew_command_text.cfg
+++ b/locale/en/redmew_command_text.cfg
@@ -7,6 +7,9 @@ diggy_clear_void=Clears the void in a given area but still triggers all events D
 crash_site_restart=Restarts the crashsite scenario.
 crash_site_restart_abort=Aborts the restart of the crashsite scenario.
 crash_site_spy=Spend coins to send a fish to spy on the enemy for a short time.
+crash_site_spy_invalid=You need to add a valid location to send a spy fish. e.g /spy [gps=-110,-17,redmew]
+crash_site_spy_funds=Training these spy fish ain't cheap! You need more coins!
+crash_site_spy_success=Jayefuu used the /spy command and spent 1000 coins to train a fish to spy on the enemy [gps=__1__,__2__,redmew]
 dataset_copy=Copies a dataset
 dataset_move=Moves a dataset
 dataset_delete=Deletes a dataset
--- a/locale/en/redmew_command_text.cfg
+++ b/locale/en/redmew_command_text.cfg
@@ -6,6 +6,7 @@ reward=Gives a reward to a target player (removes if quantity is negative)
 diggy_clear_void=Clears the void in a given area but still triggers all events Diggy would when clearing void.
 crash_site_restart=Restarts the crashsite scenario.
 crash_site_restart_abort=Aborts the restart of the crashsite scenario.
+crash_site_spy=Spend coins to send a fish to spy on the enemy for a short time.
 dataset_copy=Copies a dataset
 dataset_move=Moves a dataset
 dataset_delete=Deletes a dataset
--- a/locale/en/redmew_command_text.cfg
+++ b/locale/en/redmew_command_text.cfg
@@ -9,7 +9,7 @@ crash_site_restart_abort=Aborts the restart of the crashsite scenario.
 crash_site_spy=Spend coins to send a fish to spy on the enemy for a short time.
 crash_site_spy_invalid=You need to add a valid location to send a spy fish. e.g /spy [gps=-110,-17,redmew]
 crash_site_spy_funds=Training these spy fish ain't cheap! You need more coins!
-crash_site_spy_success=Jayefuu used the /spy command and spent 1000 coins to train a fish to spy on the enemy [gps=__1__,__2__,redmew]
+crash_site_spy_success=__1__ used the /spy command and spent 1000 coins to train a fish to spy on the enemy [gps=__2__,__3__,redmew]
 dataset_copy=Copies a dataset
 dataset_move=Moves a dataset
 dataset_delete=Deletes a dataset
--- a/map_gen/maps/crash_site/commands.lua
+++ b/map_gen/maps/crash_site/commands.lua
@@ -156,7 +156,7 @@ local function spy(args, player)
     -- {location = "[gps=-110,-17,redmew]"}
     local location_string = args.location
     local coords = {}
- 
+
     for m in string.gmatch( location_string, "%-?%d+" ) do
         table.insert(coords, tonumber(m))
     end
@@ -174,7 +174,6 @@ local function spy(args, player)
         local rad=16
         player.force.chart(player.surface, {{xpos-rad, ypos-rad}, {xpos+rad, ypos+rad}})
         game.print({'command_description.crash_site_spy_success', player_name, xpos,ypos}, Color.success)
-        
         inv.remove({name = "coin", count = 1000})
     end
 end
--- a/map_gen/maps/crash_site/commands.lua
+++ b/map_gen/maps/crash_site/commands.lua
@@ -7,6 +7,7 @@ local Popup = require 'features.gui.popup'
 local Global = require 'utils.global'
 local Ranks = require 'resources.ranks'
 local Core = require 'utils.core'
+local Color = require 'resources.color_presets'
 
 local Public = {}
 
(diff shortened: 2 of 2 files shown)


map_gen/maps/crash_site/commands.lua
```diff
@@ -146,6 +146,39 @@ local function abort(_, player)
...
+    local coords = {}
+    local coord_count = 0
+    for m in string.gmatch( location_string, "%-?%d+" ) do
+        table.insert(coords, tonumber(m))
+        coord_count = coord_count + 1
+    end
+
+    -- Do some checks then reveal the pinged map and remove 1000 coins
+    if coord_count ~= 2 then
+        game.players[player_name].print("[color=red]You need to add a valid location to send a spy fish. e.g /spy [gps=-110,-17,redmew][/color]")
```
grilledham: all instances of `game.player` can be replaced with just `player`

grilledham: We should add all the printed messages to a locale file.

map_gen/maps/crash_site/commands.lua
```diff
@@ -146,6 +146,39 @@ local function abort(_, player)
...
+        return
+    elseif coin_count < 1000 then
+        game.players[player_name].print("[color=red]Training these spy fish ain't cheap! You need more coins![/color]")
+        return
+    else
+        local xpos=coords[1]
+        local ypos=coords[2]
+        local rad=16
+        game.player.force.chart(game.player.surface, {{xpos-rad, ypos-rad}, {xpos+rad, ypos+rad}})
```
grilledham: Only whole chunks can be charted, so does this reveal 1 or 4 (or is it 6 or 9?) chunks depending on where the target coords are?
Maybe dividing the coord by 32 and flooring would be a better strategy (possible + 1 afterwards).

map_gen/maps/crash_site/commands.lua
```diff
@@ -167,30 +200,41 @@ Command.add(
...
+    },
+    restart
+)
+
+Command.add(
+    'spy',
+    {
+        description = {'command_description.crash_site_spy'},
+        arguments = {'location'},
+        required_rank = Ranks.auto_trusted,
```
grilledham: Why do they need to be trusted, is it to prevent a player wasting coins? There is no trusted check for buying from market / upgrading outposts.

map_gen/maps/crash_site/commands.lua
```diff
@@ -167,30 +200,41 @@ Command.add(
...
+        allowed_by_server = true
+    },
+    restart
+)
+
+Command.add(
+    'spy',
+    {
+        description = {'command_description.crash_site_spy'},
+        arguments = {'location'},
```
grilledham, line 230: I would add the `capture_excess_arguments = true` option. If I remember correctly without it spaces are used as separators between arguments, so `32, 64` wouldn't be considered a valid coord.

map_gen/maps/crash_site/commands.lua
```diff
@@ -146,6 +146,39 @@ local function abort(_, player)
...
+    local location_string = args.location
+    local coords = {}
+    local coord_count = 0
+    for m in string.gmatch( location_string, "%-?%d+" ) do
+        table.insert(coords, tonumber(m))
+        coord_count = coord_count + 1
+    end
+
+    -- Do some checks then reveal the pinged map and remove 1000 coins
+    if coord_count ~= 2 then
```
grilledham: This would fail if the surface name happened to be a number. Not sure if that is a problem we care about, or what we would do about that.

map_gen/maps/crash_site/commands.lua
```diff
@@ -146,6 +146,39 @@ local function abort(_, player)
...
+        game.players[player_name].print("[color=red]Training these spy fish ain't cheap! You need more coins![/color]")
+        return
+    else
+        local xpos=coords[1]
+        local ypos=coords[2]
+        local rad=16
+        game.player.force.chart(game.player.surface, {{xpos-rad, ypos-rad}, {xpos+rad, ypos+rad}})
+        game.print("[color=red]Jayefuu used the /spy command and spent 1000 coins to train a fish to spy on the enemy [gps=" .. xpos .. "," .. ypos .. ",redmew][/color]")
+        local inv = game.player.get_inventory(1)
+        inv.remove("coin")
```
grilledham: Could do 
```suggestion
        inv.remove({name = "coin", count = 1000})
```
Instead of remove one stack and add all coins minus 1000, which I think wouldn't work corretly if a player had > 100k coins.

map_gen/maps/crash_site/commands.lua
```diff
@@ -146,6 +146,39 @@ local function abort(_, player)
...
+    elseif coin_count < 1000 then
+        game.players[player_name].print("[color=red]Training these spy fish ain't cheap! You need more coins![/color]")
+        return
+    else
+        local xpos=coords[1]
+        local ypos=coords[2]
+        local rad=16
+        game.player.force.chart(game.player.surface, {{xpos-rad, ypos-rad}, {xpos+rad, ypos+rad}})
+        game.print("[color=red]Jayefuu used the /spy command and spent 1000 coins to train a fish to spy on the enemy [gps=" .. xpos .. "," .. ypos .. ",redmew][/color]")
+        local inv = game.player.get_inventory(1)
```
grilledham: I assume 1 is `defines.inventory.character_main` either add a comment, use `defines.inventory.character_main` directly or make a variable that uses  `defines.inventory.character_main` either locally or at the top of the file.

map_gen/maps/crash_site/commands.lua
```diff
@@ -146,6 +146,39 @@ local function abort(_, player)
     end
 end
 
+local function spy(args, player)
+    local player_name = player.name
+    local coin_count = player.get_item_count("coin")
```
grilledham: I would have to check to be sure, but I think this counts all coins in the player, not just the character_main inventory. Which might be a problem if the player held their coins in the cursor stack, the count would pass, but the coins wouldn't actually get removed. I suggest only counting coins in the inventory you are going to remove them from.

map_gen/maps/crash_site/commands.lua
```diff
@@ -146,6 +146,39 @@ local function abort(_, player)
...
+
+    -- Parse the values from the location string
+    -- {location = "[gps=-110,-17,redmew]"}
+    local location_string = args.location
+    local coords = {}
+    local coord_count = 0
+    for m in string.gmatch( location_string, "%-?%d+" ) do
+        table.insert(coords, tonumber(m))
+        coord_count = coord_count + 1
```
grilledham: Could use `#coords` instead of keeping count, but this is fine.

grilledham (commented):
Cool idea. Being able to use map pings is also pretty neat.

In the style of crashsite it might make more sense for it to be a button you have to click in the market, but I have no idea how that would work, so the command will do 😄.

locale/en/redmew_command_text.cfg
```diff
@@ -6,6 +6,10 @@ reward=Gives a reward to a target player (removes if quantity is negative)
 diggy_clear_void=Clears the void in a given area but still triggers all events Diggy would when clearing void.
 crash_site_restart=Restarts the crashsite scenario.
 crash_site_restart_abort=Aborts the restart of the crashsite scenario.
+crash_site_spy=Spend coins to send a fish to spy on the enemy for a short time.
+crash_site_spy_invalid=You need to add a valid location to send a spy fish. e.g /spy [gps=-110,-17,redmew]
+crash_site_spy_funds=Training these spy fish ain't cheap! You need more coins!
+crash_site_spy_success=Jayefuu used the /spy command and spent 1000 coins to train a fish to spy on the enemy [gps=__1__,__2__,redmew]
```
grilledham: Hardcoding your name into the message 😄.

grilledham (commented):
There are still some comments from the previous review that I don't think have been resolved. Let me know if you disagree with them, or you explain them better.

map_gen/maps/crash_site/commands.lua
Jayefuu: Resolved.

map_gen/maps/crash_site/commands.lua
Jayefuu: I'm not sure, I think it charts the chunk that the ping is in. Do I need to be more specific to select the centre of the chunk?

map_gen/maps/crash_site/commands.lua
Jayefuu: I believe I've resolved this now. I moved inv to the top and then always use inv. So if the coins are in a different inventory then they won't be found and it will fail early.

locale/en/redmew_command_text.cfg
Jayefuu: Fixed in new version

map_gen/maps/crash_site/commands.lua
grilledham: There is a `game.player` on line 152.
Also there are `game.players[player_name]` on line 166 and 169 that could be replaced with `player`.

map_gen/maps/crash_site/commands.lua
grilledham: I think you are confusing this with a different comment. My point here was that when reading the code it’s not easy to know what `get_inventory(1)` does, as opposed to `get_inventory(defines.inventory.character_main)`.

map_gen/maps/crash_site/commands.lua
grilledham: It reveals all chunks that are in the area you provide. So my point was that depending on where you ping within a chunk, potentially a different number of chunks could be revealed. Say for example the area crosses chunk boundaries. 

If the intention was to only reveal one chunk then you could probably just use `{{xpos, ypos}, {xpos, ypos}})` (untested).

grilledham (commented):
This looks good. There are still two unresolved conversations though.

map_gen/maps/crash_site/commands.lua
Jayefuu: Since it's a crash site specific command, I can't see us ever not using redmew or nauvis as the surface. 

However, it would be easier to make the check `if coord_count < 2 then` it wouldn't error if the surface name had a number in it.

map_gen/maps/crash_site/commands.lua
Jayefuu: Fixed in latest commits. Thank you.

Jayefuu:
> 
> 
> This looks good. There are still two unresolved conversations though.

I think I got them all now? :)

grilledham (approved):

PR #1105, opened 2020-11-08
Ping corpse location on player death

grilledham:


Jayefuu:
This is a great addition! Being able to look up in the chat to find  a marker to a body will be really useful.

Do you think it might be better to use player.print instead of player.force.print? The ping noise every time someone dies could be annoying and it will be easier to find your corpse ping if it messages only one player not all players upon death so there aren't lots of other pings to choose from.

features/server.lua
```diff
@@ -696,59 +677,59 @@ Event.add(Public.events.on_server_started, set_scenario_version)
...
-            message[#message + 1] = ' was killed by '
+    local cause = event.cause
 
-            local name = cause.name
-            if name == 'character' then
-                name = cause.player.name
-            end
+    local message = {discord_bold_tag, player.name}
+    if cause and cause.valid then
+        message[#message + 1] = ' was killed by '
```
Jayefuu, line 710: `message[#message + 1] = ' was killed by '` should be `message[#message + 1] = ' was killed by a'` so that the message is "grilledham was killed by a train" or "grilledham was killed by a fish".

features/server.lua
```diff
@@ -696,59 +677,59 @@ Event.add(Public.events.on_server_started, set_scenario_version)
...
+    message[#message + 1] = string.format('%.1f', position.x)
+    message[#message + 1] = ','
+    message[#message + 1] = string.format('%.1f', position.y)
+    message[#message + 1] = ','
+    message[#message + 1] = player.surface.name
+    message[#message + 1] = ']'
+
+    message = concat(message)
+    raw_print(message)
+end)
```
Jayefuu, line 735: Is  there a good reason to build the message as an array then concatenate it rather than doing:
`print(player.name .. "was killed by a " .. cause.name .. ". [gps='" .. string.format('%.1f', position.x) .. "," .. string.format('%.1f', position.y) .. "," .. player.surface.name .. "]")` ?

Jayefuu (changes requested):
Looks good. Just a minor change needed and a question.

features/server.lua
grilledham: I don't disagree with you that the grammar is wrong, but it's consistent with the death messages that factorio outputs i.e. factorio does this `"grilledham was killed by Fish"`. 
We would also have to handle the case of being killed by a player differently i.e. `"grilledham was killed by a Jayefuu"` wouldn't be right.
Do you think it's worth making the change?

features/server.lua
grilledham: That wouldn't handle the case of `cause` being nil`.

But I assume your question is about `table.concat` vs the concat operator `..`. `..` is inefficient because it allocates a new string for each concat, whereas the `table.concat` function can be more efficient in how it contacts strings. I’m not sure what the actual implementation is. So when doing multiple concats I’ve gotten into the habit of using `table.concat`.

features/server.lua
grilledham: I've added an 'a' when not killed by a player.

Jayefuu: I've just retested this. I have two issues that I'm not sure are new or just un-noticed previously:
1) The cause printed to discord is event.cause.name which prints:
` Jayefuu was killed by a medium-worm-turret. [gps=0.0,0.0,redmew]`
event.cause.name needs changing to a human readable name instead, event.cause.localised_name.
2) However the in-game message is still `Jayefuu was killed by Medium worm`. Do we have control over this death message? If so this should also be changed to add "a" if it's not a player that killed the player.

Jayefuu (changes requested):

features/server.lua
grilledham: 1. We can't print localised names to discord, I think we just have to live with that.
2. We can't control the default message that Factorio prints in game.

Jayefuu (approved):

PR #1106, opened 2020-11-08
Update events.lua to fix robot follower count bug

Jayefuu:
Checks that there is an artillery wagon near upon robot follower timeout before creating a cluster grenade projectile.


map_gen/maps/crash_site/events.lua
```diff
@@ -487,9 +487,12 @@ Event.add(
         local entity = event.robot
         local position = entity.position
         if entity.force.name == 'enemy' then
-            entity.surface.create_entity{name = "cluster-grenade", position=position, target=position, speed=1}
+            -- only create a grenade entity if there is an artillery wagon near, otherwise the maximum robot follower count will spoil your fun
```
grilledham: Can you get this from `event.owner`? <https://lua-api.factorio.com/latest/events.html#on_combat_robot_expired>
Failing that count_entities_filtered <https://lua-api.factorio.com/latest/LuaSurface.html#LuaSurface.count_entities_filtered> might be a better choice.

map_gen/maps/crash_site/events.lua
```diff
@@ -483,13 +483,11 @@ Event.add(
...
     function(event)
-
         local entity = event.robot
         local position = entity.position
         if entity.force.name == 'enemy' then
-            -- only create a grenade entity if there is an artillery wagon near, otherwise the maximum robot follower count will spoil your fun
-            local wagon_entities = entity.surface.find_entities_filtered {name="artillery-wagon", position=position, radius=15} 
-            if #wagon_entities > 0 then
+            -- only create a grenade entity if an artillery wagon (the event.owner) killed the target that spawned the combabt robot
+            if event.owner.name == "artillery-wagon"
```
grilledham: owner could be nil or invalid so we will have to check that.

map_gen/maps/crash_site/events.lua
```diff
@@ -483,13 +483,17 @@ Event.add(
...
         local entity = event.robot
         local position = entity.position
+        local owner = event.owner
+        if owner == nil or not owner.valid then
+            return
+        end
         if entity.force.name == 'enemy' then
-            entity.surface.create_entity{name = "cluster-grenade", position=position, target=position, speed=1}
+            -- only create a grenade entity if an artillery wagon (the event owner) killed the target that spawned the combabt robot
+            if owner.name == "artillery-wagon" then
```
grilledham: You could merge this with the previous `if`, but that is just a matter of style.

grilledham (approved):

map_gen/maps/crash_site/events.lua
Jayefuu: You're right it's neater, it was left over from some tests. I've changed it as suggested.

Issue #1102, opened 2020-11-05
Jailing players while they're in a spidertron causes issues


Jayefuu:
I don't have any more details for now. Reported by Tigress88 on Crash Site. Leaving this not here for me to look into it.

Ferefang:
cript @/external/factorio/7/temp/currently-playing/utils/command.lua:261: Error while running 'jail' with arguments {player = "MeggalBozale"}: Entity is not car or unit or projectile or artillery-projectile

Jayefuu:
Thanks @Ferefang, I think I can fix this so will have a go.

Jayefuu:
This is now fixed and merged.

PR #1108, opened 2020-11-13
Spidertron jail fix

Jayefuu:
Issue #1102 describes errors when jailing a player while they're inside a spidertron.

This PR fixes that by checking if they're in a spidertron and setting the autopilot destination rather than the speed like for trains/cars/tanks.

### DIFF
--- a/features/report.lua
+++ b/features/report.lua
@@ -195,7 +195,6 @@ function Module.jail(target_player, player)
 
     -- Add player to jail group
     permission_group.add_player(target_player)
-
     -- If in vehicle, kick them out and set the speed to 0.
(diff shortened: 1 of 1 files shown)


features/report.lua
```diff
@@ -195,14 +195,18 @@ function Module.jail(target_player, player)
 
     -- Add player to jail group
     permission_group.add_player(target_player)
-
```
grilledham: Why remove this line break?

grilledham (approved):

features/report.lua
Jayefuu: I didn't mean to. I've now fixed it.

PR #1110, opened 2020-11-15
Fix unjail invulnerable characters

### LINKED ISSUE #1103: /jail bug: Invulnerable character upon leaving after jail
If a jailed player leaves, then they're unjailed by an admin before they rejoin, they are invulnerable when they join. You cannot shoot them (you get a hover text saying "you cannot damage this") and the biters do not detect or get angry.

I've tested I can reproduce this. Was found on crash site. Tested on S10 with the develop branch as we were playing a copy with /spy in it on S7when it was found.

Here are some steps that created the bug and what I tried unsuccessfully to fix it:
/jail nerdinski27
-player leaves-
/unjail nerdinski27
-player joins- (and is invulnerable)
/jail nerdinski27
/dc game.players["nerdinski27"].character.destroy()
/c game.players["nerdinski27"].create_character()
/unjail nerdinski27
-player is invulnerable-(the above doesn't fix the bug)

After these the player is still invulnerable.
/jail nerdinski27
-player leaves-
/unjail nerdinski27
-player joins- (and is invulnerable)
/jail nerdinski27
/dc game.players["nerdinski27"].character.destroy()
/c game.players["nerdinski27"].create_character()
Can now shoot him and he will die, but is still jailed and will respawn invulnerable.

Once the problem occurs this is the sequence to fix it:
/jail nerdinski27
-player leaves-
/unjail nerdinski27
<player joins> (and is invulnerable)
/jail nerdinski27
/dc game.players["nerdinski27"].character.destroy()
/unjail nerdinski27
/c game.players["nerdinski27"].create_character()
<player is fixed and can be shot, agros biters etc>

grilledham:
- If a player didn't have a character when they were jailed we would record their previous destructible state as `nil` which was treated as `false`. Now we treat `nil` as no previous character info and set the previous destructible state on repsawn/join.
- If a player was unjailed when offline or dead, we wouldn't restore the player character's destructible state, now we wait for them to rejoin/respawn to do that.

jail_test-deflated.zip on s10 to test.

fixes #1103

### DIFF
--- a/features/report.lua
+++ b/features/report.lua
@@ -5,6 +5,7 @@ local Global = require 'utils.global'
 local Command = require 'utils.command'
 local Popup = require 'features.gui.popup'
 local Color = require 'resources.color_presets'
+local Event = require 'utils.event'
 
 local format = string.format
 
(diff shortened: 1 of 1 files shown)


Jayefuu (approved):
Looks good! Tested it on server 10 and seems to have fixed both previous problems.

PR #1111, opened 2020-11-15
Fix report showing outdated time.

### LINKED ISSUE #1107: Timer on /report does not work
When a player uses the report feature to report another player, the pop up when an admin logs into the game is supposed 
display the game time that report was received and how long ago that was. It does not display how long ago it was received properly. It always states 0 seconds ago.

The report feature should state game time that report occurred AND how long ago that was.

1. Player uses /report while no admin is online
2. Admin logs into map and report pops up.

timer

I had a player create a /report before I logged in. In screenshot you can see the timer that says 0 seconds ago. And in the chat, you can see the game time showing that it was actually reported 11 minutes ago.

grilledham:
When an admin rejoins, if they have the reports open, reopen it again to update the time on the reports.

Fixes #1107

### DIFF
--- a/features/report.lua
+++ b/features/report.lua
@@ -441,7 +441,7 @@ Gui.on_click(reporting_submit_button_name, function(event)
     print(prefix_e)
 end)
 
-local function restore_non_character_unjailed_players(event)
+local function restore_player_jailed_state(event)
     local player_index = event.player_index
     local player = game.get_player(player_index)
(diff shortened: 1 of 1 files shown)


Jayefuu (approved):
I can't see any problems with the code. If you've tested and are confident it works as intended then go ahead and merge. If you need help testing it on S10 grab me on discord and I'll help.

PR #1114, opened 2020-11-17
WIP - airstrike command

Jayefuu:
- Type `/strike [gps=x,y,surface] to send a large cloud of poison into the enemy
- Searches a chest near spawn for poison capsules. Current cost for command is 100 capsules but could be balanced.
- Needs market items adding for technology research, bought with coins.
- Variables strikeCount and strikeRadius start low. As the market upgrades are made (start at 5k each?) the size of teh strike and amount of capsules increases.
- Should be balanced to allow use from mid-game onwards. Don't allow early game use as manual throwing is fine for early outposts. This should be to speed up the boring end-game clearing of worms.
- Need to consider if referencing a chest at -0.5 -0.35 will cause problems. On one hand I like it because you can replace it with a provider. It's also so specific that it doesn't pick up other chests near it. But it might be confusing if someone removes it. I've added checks but we should discuss if this is the best method of collecting resources from the players.
- Probably needs refactoring to take it out of commands.lua and put it in a new features\airstrike.lua

### DIFF
--- a/locale/en/redmew_command_text.cfg
+++ b/locale/en/redmew_command_text.cfg
@@ -16,7 +16,6 @@ crash_site_airstrike_radius=Upgrade the airstrike radius to level __1__\n\nTo us
 crash_site_airstrike_radius_name_label=Airstrike Radius __1__
 crash_site_airstrike_count_name_label=Airstrike Damage __1__
 crash_site_airstrike_not_researched=You have not researched airstrike yet. Visit the market at the spawn [gps=-3,-3,redmew]
-crash_site_airstrike_chest_removed_error=Someone removed the chest. Replace it here: [gps=-0.5,-3.5,redmew]
 crash_site_airstrike_insufficient_currency_error=To send an air strike, load __1__ more poison capsules into the payment chest [gps=-0.5,-3.5,redmew]
 crash_site_airstrike_friendly_fire_error=You don't want to do that, no enemies found in the target area.
 crash_site_airstrike_damage_upgrade_success=Airstrike damage has been upgraded to level __1__
--- a/locale/en/redmew_command_text.cfg
+++ b/locale/en/redmew_command_text.cfg
@@ -9,9 +9,19 @@ crash_site_restart_abort=Aborts the restart of the crashsite scenario.
 crash_site_spy=Spend coins to send a fish to spy on the enemy for a short time.
 crash_site_spy_invalid=You need to add a valid location to send a spy fish. e.g /spy [gps=-110,-17,redmew]
 crash_site_spy_funds=Training these spy fish ain't cheap! You need more coins!
+crash_site_airstrike_invalid=Invalid co-ordinates.
 crash_site_spy_success=__1__ used the /spy command and spent 1000 coins to train a fish to spy on the enemy [gps=__2__,__3__,redmew]
-crash_site_airstrike_count=Upgrade the airstrike damage to  to level __1__\n\nTo use airstrike place poison capsules in the spawn chest then type /strike with a gps position\n\nDamage upgrades increase the number of poison capsules launched by airstrike.\n\nCurrent level: __2__\nCurrent shell count: __3__\nCurrent cost: __4__ poison capsules"
+crash_site_airstrike_count=Upgrade the airstrike damage to  to level __1__\n\nTo use airstrike place poison capsules in the spawn chest then type /strike with a gps position\n\nDamage upgrades increase the number of poison capsules launched by airstrike.\n\nCurrent level: __2__\nCurrent shell count: __3__\nCurrent cost: __4__
 crash_site_airstrike_radius=Upgrade the airstrike radius to level __1__\n\nTo use airstrike place poison capsules in the spawn chest then type /strike with a gps position\n\nCurrent level: __2__\nCurrent radius: __3__
+crash_site_airstrike_radius_name_label=Airstrike Radius __1__
+crash_site_airstrike_count_name_label=Airstrike Damage __1__
+crash_site_airstrike_not_researched=You have not researched airstrike yet. Visit the market at the spawn [gps=-3,-3,redmew]
+crash_site_airstrike_chest_removed_error=Someone removed the chest. Replace it here: [gps=-0.5,-3.5,redmew]
+crash_site_airstrike_insufficient_currency_error=To send an air strike, load __1__ more poison capsules into the payment chest [gps=-0.5,-3.5,redmew]
+crash_site_airstrike_friendly_fire_error=You don't want to do that, no enemies found in the target area.
+crash_site_airstrike_damage_upgrade_success=Airstrike damage has been upgraded to level __1__
+crash_site_airstrike_radius_upgrade_success=Airstrike radius has been upgraded to level __1__
+crash_site_airstrike_success=__1__ called an air strike [gps=__2__,__3__,redmew]
 dataset_copy=Copies a dataset
 dataset_move=Moves a dataset
 dataset_delete=Deletes a dataset
--- a/locale/en/redmew_command_text.cfg
+++ b/locale/en/redmew_command_text.cfg
@@ -10,6 +10,8 @@ crash_site_spy=Spend coins to send a fish to spy on the enemy for a short time.
 crash_site_spy_invalid=You need to add a valid location to send a spy fish. e.g /spy [gps=-110,-17,redmew]
 crash_site_spy_funds=Training these spy fish ain't cheap! You need more coins!
 crash_site_spy_success=__1__ used the /spy command and spent 1000 coins to train a fish to spy on the enemy [gps=__2__,__3__,redmew]
+crash_site_airstrike_count=Upgrade the airstrike damage to  to level __1__\n\nTo use airstrike place poison capsules in the spawn chest then type /strike with a gps position\n\nDamage upgrades increase the number of poison capsules launched by airstrike.\n\nCurrent level: __2__\nCurrent shell count: __3__\nCurrent cost: __4__ poison capsules"
+crash_site_airstrike_radius=Upgrade the airstrike radius to level __1__\n\nTo use airstrike place poison capsules in the spawn chest then type /strike with a gps position\n\nCurrent level: __2__\nCurrent radius: __3__
 dataset_copy=Copies a dataset
 dataset_move=Moves a dataset
 dataset_delete=Deletes a dataset
--- a/map_gen/maps/crash_site/commands.lua
+++ b/map_gen/maps/crash_site/commands.lua
@@ -11,6 +11,7 @@ local Ranks = require 'resources.ranks'
 local Core = require 'utils.core'
 local Color = require 'resources.color_presets'
 local Toast = require 'features.gui.toast'
+local Utils = require 'utils.core'
 local set_timeout_in_ticks = Task.set_timeout_in_ticks
 
 local Public = {}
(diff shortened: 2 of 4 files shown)


map_gen/maps/crash_site/scenario.lua
```diff
@@ -812,7 +828,7 @@ local function init(config)
         [1] = {
             market = market,
             [15] = {entity = {name = 'market', force = 'neutral', callback = 'market'}},
-            [18] = {entity = {name = 'wooden-chest', force = 'player'}}
+            [18] = {entity = {name = 'wooden-chest', force = 'player', destructible=false, minable=false}}
```
Jayefuu: I tried adding `destructible=false, minable=false` but it didn't work when I tested it.

grilledham: I've pushed changes to make this work.

Jayefuu:
@grilledham pending style/bug fixes, I think I've taken this as far as I can without a live play test.

map_gen/maps/crash_site/commands.lua
```diff
@@ -171,12 +187,139 @@ local function spy(args, player)
...
+    local count_level = airstrike_data.count_level      -- the number of poison capsules launched at the enemy
+    local radius = 5+(radius_level*3)
+    local count = (count_level-1)*5+3
+    local strikeCost = count * 4
+
+    local name = item.name
+    if name == 'airstrike_damage' then
+        airstrike_data.count_level = airstrike_data.count_level + 1
+
+        Toast.toast_all_players(2, "Airstrike damage has been upgraded to level " .. count_level)
```
grilledham: Duration of 2 seems a bit short we use 15 for outpost upgrades.

map_gen/maps/crash_site/scenario.lua
```diff
@@ -771,6 +771,22 @@ local function init(config)
             upgrade_rate = 0.5,
             upgrade_base_cost = 500,
             upgrade_cost_base = 2,
+            {
+                price = 1000,
+                type= 'airstrike',
+                name = 'airstrike_damage',
+                name_label = 'Airstrike Damage 1',
```
grilledham: Use a Localised string for name_label for airstrike_damage and airstrike_radius

map_gen/maps/crash_site/scenario.lua
```diff
@@ -771,6 +771,22 @@ local function init(config)
...
+                sprite = 'item-group/production',
+                description = 'Upgrade the airstrike damage to  to level 1. Damage upgrades increase the number of poison capsules launched by airstrike. Use /strike with a gps position'
```
grilledham: Use the Localised string that you already made here too. (And for airstrike_radius).

locale/en/redmew_command_text.cfg
```diff
@@ -10,6 +10,8 @@ crash_site_spy=Spend coins to send a fish to spy on the enemy for a short time.
 crash_site_spy_invalid=You need to add a valid location to send a spy fish. e.g /spy [gps=-110,-17,redmew]
 crash_site_spy_funds=Training these spy fish ain't cheap! You need more coins!
 crash_site_spy_success=__1__ used the /spy command and spent 1000 coins to train a fish to spy on the enemy [gps=__2__,__3__,redmew]
+crash_site_airstrike_count=Upgrade the airstrike damage to  to level __1__\n\nTo use airstrike place poison capsules in the spawn chest then type /strike with a gps position\n\nDamage upgrades increase the number of poison capsules launched by airstrike.\n\nCurrent level: __2__\nCurrent shell count: __3__\nCurrent cost: __4__ poison capsules"
```
grilledham: Remove `"` at end of line.

```suggestion
crash_site_airstrike_count=Upgrade the airstrike damage to  to level __1__\n\nTo use airstrike place poison capsules in the spawn chest then type /strike with a gps position\n\nDamage upgrades increase the number of poison capsules launched by airstrike.\n\nCurrent level: __2__\nCurrent shell count: __3__\nCurrent cost: __4__ poison capsules
```

map_gen/maps/crash_site/commands.lua
```diff
@@ -171,12 +187,139 @@ local function spy(args, player)
...
+    if item.type ~= 'airstrike' then
+        return
+    end
+
+    -- airstrike stuff
+    local radius_level = airstrike_data.radius_level    -- max radius of the strike area
+    local count_level = airstrike_data.count_level      -- the number of poison capsules launched at the enemy
+    local radius = 5+(radius_level*3)
+    local count = (count_level-1)*5+3
+    local strikeCost = count * 4
```
grilledham, line 299: `strikeCost` is 4 times count here, but in the command it is 3 time count?

map_gen/maps/crash_site/commands.lua
```diff
@@ -171,12 +187,139 @@ local function spy(args, player)
...
+    local capCount = inv.get_item_count("poison-capsule")
+
+    if capCount < strikeCost then
+        player.print("To send an air strike, load " .. strikeCost - capCount .. " more poison capsules into the payment chest [gps=-0.5,-3.5,redmew]")
+        return
+    end
+
+    -- Do a simple check to make sure the player isn't trying to grief the base
+    --local enemyEntities = s.find_entities_filtered {position = {xpos, ypos}, radius = radius, force = "enemy"}
+    local enemyEntities = game.player.surface.find_entities_filtered {area={{100,-100},{-100,100}}, force = "enemy"}
```
grilledham: If we only care if the count is 0 or not, we could use `count_entities_filtered` and set the limit to 1.

map_gen/maps/crash_site/commands.lua
```diff
@@ -171,12 +187,139 @@ local function spy(args, player)
...
+    game.print(#enemyEntities)
```
grilledham: I assume this was debugging code that you forgot to remove.

map_gen/maps/crash_site/commands.lua
```diff
@@ -171,12 +187,139 @@ local function spy(args, player)
...
+    if #enemyEntities < 1 then
+        player.print("You don't want to do that, no enemies found in the target area.")
+        for _, p in pairs(game.players) do
```
grilledham: Could use <https://github.com/Refactorio/RedMew/blob/develop/utils/core.lua#L53>

map_gen/maps/crash_site/commands.lua
```diff
@@ -171,12 +187,139 @@ local function spy(args, player)
...
+        return
+    end
+    local xpos=coords[1]
+    local ypos=coords[2]
+
+    -- Check the chest is still there. Better than making destructible and mineable false because then players can upgrade the chest
+    local entities = s.find_entities_filtered {position = {-0.5, -3.5}, type = 'container', limit=1}
+    local dropbox = entities[1]
+    if dropbox == nil then
+        player.print("Someone removed the chest. Replace it here: [gps=-0.5,-3.5,redmew] ")
```
grilledham: Make a localised string.

map_gen/maps/crash_site/commands.lua
```diff
@@ -171,12 +187,139 @@ local function spy(args, player)
...
+        return
+    end
+
+    -- Check the contents of the chest by spawn for enough poison capsules to use as payment
+    local inv = dropbox.get_inventory(defines.inventory.chest)
+    local capCount = inv.get_item_count("poison-capsule")
+
+    if capCount < strikeCost then
+        player.print("To send an air strike, load " .. strikeCost - capCount .. " more poison capsules into the payment chest [gps=-0.5,-3.5,redmew]")
```
grilledham: Make a localised string.

map_gen/maps/crash_site/commands.lua
```diff
@@ -171,12 +187,139 @@ local function spy(args, player)
...
+        for _, p in pairs(game.players) do
+            if p.admin then
+                p.print(player.name .. " tried to airstrike the base here: [gps=" .. xpos .. "," .. ypos ..",redmew]")
+            end
+        end
+        return
+    end
+
+    inv.remove({name = "poison-capsule", count = strikeCost})
+    game.print(player.name .. " called an air strike [gps=" .. xpos .. "," .. ypos ..",redmew]")
```
grilledham: Make a localised string.

grilledham (commented):
I think a fixed chest at spawn is easier to understand as a player, we could maybe change it to a requester chest.

+1 to moving the code to a different file.

I'd also probably change the market item descriptions to be more like the outpost upgrades.
For example for airstrike damage I'd do something like:
 - Change item name to `Upgrade Airstrike Damage to __level__`
 - Remove current level, it's obviously one less than the upgrade level.
 - Change Current shell count to be : Shell count 3 -> 8.
 - Change Current Cost to : Cost 12 -> 32 [item=poison-capsule]

map_gen/maps/crash_site/commands.lua
Jayefuu: Updated in upcoming commit.

map_gen/maps/crash_site/scenario.lua
Jayefuu: Updated in upcoming commit.

locale/en/redmew_command_text.cfg
Jayefuu: Updated

map_gen/maps/crash_site/scenario.lua
Jayefuu: Updated

map_gen/maps/crash_site/commands.lua
Jayefuu: Updated 3 to 4. Needed cost increase after balance checks.

map_gen/maps/crash_site/commands.lua
Jayefuu: Done.

map_gen/maps/crash_site/commands.lua
Jayefuu: Done

map_gen/maps/crash_site/commands.lua
Jayefuu: Done.

map_gen/maps/crash_site/commands.lua
Jayefuu: Ah ha! I was looking for something like that.

map_gen/maps/crash_site/commands.lua
```diff
@@ -171,12 +188,134 @@ local function spy(args, player)
...
+
+    -- Do some checks on the coordinates passed in the argument
+    if #coords < 2 then
+        player.print({'command_description.crash_site_airstrike_invalid'}, Color.fail)
+        return
+    end
+    local xpos=coords[1]
+    local ypos=coords[2]
+
+    -- Check the chest is still there. Better than making destructible and mineable false because then players can upgrade the chest
```
grilledham: The code to make the chest indestructible and unmineable is still in <https://github.com/Refactorio/RedMew/pull/1114/files#diff-c951ee557a72cf92bec77bc9e8f8cbf30e973c5557fafe84a908f914377f6f4dR836>.
Is that intentional?

map_gen/maps/crash_site/commands.lua
```diff
@@ -171,12 +188,134 @@ local function spy(args, player)
...
+end)
+
+local function strike(args, player)
+    local s = player.surface
+    local location_string = args.location
+    local coords = {}
+
+    local radius_level = airstrike_data.radius_level    -- max radius of the strike area
+    local count_level = airstrike_data.count_level      -- the number of poison capsules launched at the enemy
+    if count_level == 1 then
```
grilledham, line 216: So you only to upgrade `count_level` before you can use strike? A bit confusing as in the market radius is shown as n/a before you upgrade it.

map_gen/maps/crash_site/commands.lua
grilledham: I think you should make this change.

Jayefuu:
I have changed the radius description, having it as n/a was a mistake. The intended behaviour is that it has a starting radius at level 0 of 5 tiles. But you can't fire an airstrike until you get the first level of damage.

I've removed the superfluous code for a missing chest.

map_gen/maps/crash_site/commands.lua
grilledham: You still need to check that the chest was found and otherwise print some error message something like can not run command, spawn chest not found. I'd recommend checking that the chest is not nil or invalid. 

There is also the question of what happens if you run /strike during the cutscene. But I think if you bring back the check that will be fine.

My original comment was that the code comment was incorrect, which yo removed then brought back?

Jayefuu:
I replaced the code to check the chest location.

grilledham (approved):

PR #1116, opened 2020-11-29
Fixed artillery pickup exploit

Jayefuu:
- Current behaviour is that when destroyers/defenders are created their target position is set to the entity that killed them.
- If a player picks up the artillery after the capsules spawn then they lose their cause.target and stop, not reaching the cause of the turret death. If the player picks up the artillery after it spawns the killing projectile but before it hits the target, then the event has no cause and no bots spawn!
- I've changed this behaviour so that if the target.cause is artillery then it spawns a defender/destroyer projectile NOT an entity. The projectile target can therefore be a position not an entity. So the capsules won't stop if the artillery is picked up.
- If the cause is not artillery turret or wagon then a destroyer/defender capsule entity is spawned instead of a projectile because a projectile cannot track the player. This will mean we'll still get the nice behaviour where the capsule target updates as the player/tank/car/spidertron moves and they can kite them.
- If the player picks up the artillery turret before the projectile that kills a turret arrives then the event cause.name will be missing and no bots spawn at all. I have changed the behaviour so that "if not cause" then we assume they're doing something fucky. We won't have the entity cause name or position, so let's just spawn double the normal amount of destroyers and just send them at 0, 0 since we know nothing else to do with them.

### DIFF
--- a/map_gen/maps/crash_site/events.lua
+++ b/map_gen/maps/crash_site/events.lua
@@ -295,48 +295,78 @@ local bot_cause_whitelist = {
 }
 
 local function do_bot_spawn(entity_name, entity, event)
-    if not bot_spawn_whitelist[entity_name] then
-        return
-    end
 
     local cause = event.cause
-    if not cause or not bot_cause_whitelist[cause.name] then
-        return
-    end
-
     local entity_force = entity.force
     local ef = entity_force.evolution_factor
+    local create_entity = entity.surface.create_entity
 
     if ef <= 0.2 then
         return
     end
-
-    local create_entity = entity.surface.create_entity
-    local repeat_cycle = 1 -- The number of times a squad of robots are spawned default must be 1
-    if ef > .95 then
-        repeat_cycle = 2
-    end
-
+    
     local spawn_entity = {
         position = entity.position,
         target = cause,
         force = entity_force
     }
 
+    if not cause then
+        -- If we reach here then the player might have picked up the artillery turret before the projectile hit the entity and killed it.
+        -- We therefore won't know the location of the artillery turret/wagon that killed the turret, so let's punish them and send even more bots, straight at spawn?
+        for i = 1, 60 do
+            spawn_entity.name = 'destroyer-capsule'
+            spawn_entity.speed = 0.2
+            spawn_entity.target = {0,0}
+            create_entity(spawn_entity)
+        end
+        return
+    end
+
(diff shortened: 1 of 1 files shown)


map_gen/maps/crash_site/events.lua
```diff
@@ -295,48 +295,78 @@ local bot_cause_whitelist = {
 }
 
 local function do_bot_spawn(entity_name, entity, event)
-    if not bot_spawn_whitelist[entity_name] then
```
grilledham, line 298: Why is this check no longer at the top? Does this mean if you kill enemy walls with artillery and remove the artillery before the wall dies that lots of enemy bots are sent to spawn?

Jayefuu: Woops! Thanks. I've fixed it so that it returns if not whitelist and not nil cause.

map_gen/maps/crash_site/events.lua
```diff
@@ -295,48 +295,78 @@ local bot_cause_whitelist = {
...
         position = entity.position,
         target = cause,
         force = entity_force
     }
 
+    if not bot_spawn_whitelist[entity_name] then
+        if cause then
+            return
+        end
+    end
```
grilledham: This still isn't correct. In the case that the entity is not on the whitelist, e.g. walls and the artillery turret has been removed, this would still spawn bots.

Also I think the check should be at the `if not bot_spawn_whitelist[entity_name]` should be at the very top of the function, so we don't do any unnecessarily work if we can avoid it.

grilledham (changes requested):

map_gen/maps/crash_site/events.lua
```diff
@@ -295,48 +295,79 @@ local bot_cause_whitelist = {
...
+
+    local repeat_cycle = 1 -- The number of times a squad of robots are spawned default must be 1
+    if ef > .95 then
+        repeat_cycle = 2
+    end
+
     if cause.name ~= 'character' then
-        if entity_name == 'artillery-turret' then
+        if (entity_name == 'artillery-turret') or (entity_name == 'artillery-wagon') then
```
grilledham: No action required here unless you want to change it, but why are we checking for 'artillery-wagon' there are no enemy artillery-wagons on crashsite?

Jayefuu: I got confused between entity_name and cause.name I think. Will update.

grilledham (approved):

PR #1117, opened 2020-12-08
Fixed /strike command range issue

Jayefuu:
Previously the airstrike was not reaching the corners of the map. There was a default max_range value of 1000 that limited how far the poison capsules could travel. Have set it to 1400 so that it can reach the corners of the map.

~~THIS NEEDS PLAY TESTING. Please don't merge until I've confirmed I've play tested it.~~

### DIFF
--- a/map_gen/maps/crash_site/commands.lua
+++ b/map_gen/maps/crash_site/commands.lua
@@ -203,7 +203,7 @@ end
 
 local spawn_poison_callback = Token.register(function(data)
     local r = data.r
-    data.s.create_entity{name = "poison-capsule", position={0,0}, target={data.xpos + math.random(-r,r), data.ypos + math.random(-r,r)}, speed=10}
+    data.s.create_entity{name = "poison-capsule", position={0,0}, target={data.xpos + math.random(-r,r), data.ypos + math.random(-r,r)}, speed=10, max_range=1400}
 end)
 
 local function strike(args, player)
--- a/map_gen/maps/crash_site/commands.lua
+++ b/map_gen/maps/crash_site/commands.lua
@@ -203,7 +203,7 @@ end
 
 local spawn_poison_callback = Token.register(function(data)
     local r = data.r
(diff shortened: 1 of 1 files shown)


Jayefuu:
I've now tested that this works. Poison capsules now reach the corners as intended.

map_gen/maps/crash_site/commands.lua
```diff
@@ -203,7 +203,7 @@ end
 
 local spawn_poison_callback = Token.register(function(data)
     local r = data.r
-    data.s.create_entity{name = "poison-capsule", position={0,0}, target={data.xpos + math.random(-r,r), data.ypos + math.random(-r,r)}, speed=10}
+    data.s.create_entity{name = "poison-capsule", position={0,0}, target={data.xpos + math.random(-r,r), data.ypos + math.random(-r,r)}, speed=10, max_range=1400}
```
grilledham: Can we set the position to data.xpos and data.ypos instead? It would be better to not hard code the range. I'm thinking that not all crashsite maps are the same size.

Jayefuu: I really like that you can see the poison capsules appear at spawn and sail over the map. Your solution then they'd just appear from nowhere. I could set the max_range very, very high. According to Klonan there never used to be a range limit but people were crashing the game by sending the projectile into infinity. We're unlikely to ever have one bigger than 10000 so we could set the max_range to that.

Another option to solve both problems would be to add a third market purchase so that to reach further they need to upgrade the artillery strike range. But I think this might feel a bit grindy for coins and make them use it less.

grilledham: Ok that makes sense.

Can we make the range a really high number like 2,000,000 so there is no chance we have to worry about the range being too short in any future crashsite version.

Jayefuu: Yes. I'll set it to 100k. That's insanely big as it is.

grilledham (approved):

PR #1119, opened 2020-12-14
Loader GUI buttons

Jayefuu:
Adding this to give myself a to-do list:
- [X] Mouse buttons and shift control quantity crafted
- [X] Logistics-1 research enables loader
- [x] logistics 2 and 3 enable fast and express loaders
- [x] Change GUI behaviour so loader buttons are stacked vertical not horizontal
- [x] Add fast and express loader buttons to assembly machines
- [x] Make clicking the player gui loader 2 and 3 buttons do something. Find a good way to do it without repeating the code block for Gui.on_click(player_craft_loader_1, function(event)
- [x] Add tooltips
- [x] How does it handle offline players? We update the gui for all players when research is done and we update the gui for new players when they are created, but what if research happens when someone is offline and then they join? Need to update for each player as they join? Needs testing.


Jayefuu:
@grilledham I think this now has all of the features we discussed. I've tested it on S10 with Anubi5 and Tigress and it handles new players and offline players fine, the loaders unlock as research is finished and it handles left, right and shift+left clicks.

How do I fix the failed checks? I don't understand them.
a) We use defines in other places, why doesn't it like them here?
b) Why doesn't it like count? It's used on line 423.....

features/redmew_qol.lua
```diff
@@ -339,11 +310,172 @@ if config.save_bots then
...
+    local player = game.get_player(event.player_index)
+    local player_force = player.force
+    if not player_force.technologies['logistics'].researched then
+        return
+    end
+    local panel = player.gui.relative
+    draw_loader_frame_for_player(panel)
+end)
+
+Event.add(defines.events.on_research_finished, function(event)
```
grilledham: Can we merge this will the on_research_finished event above. (Move `enable_loaders` function down here).
Also only register the event when config.loaders is truthy.

features/redmew_qol.lua
```diff
@@ -339,11 +310,172 @@ if config.save_bots then
...
+    local anchor = {
+        gui = defines.relative_gui_type.assembling_machine_select_recipe_gui,
+        position = defines.relative_gui_position.right
+    }
+    frame = parent.add {type = 'frame', name = loader_crafter_frame_for_assembly_machine_name, anchor = anchor, direction='vertical'}
+
+    local button = frame.add {type = 'choose-elem-button', name = machine_craft_loader_1, elem_type='recipe', recipe = 'loader'}
+    button.locked = true
+    Gui.set_data(button, entity)
+    if player_force.technologies['logistics-2'].researched then
```
grilledham: We should check the recipe is enabled, not the technology is research, that way we can disable the recipe for modded games when we forget to turn off the config.

(This also applies to other places where you check for the research).

features/redmew_qol.lua
```diff
@@ -339,11 +310,172 @@ if config.save_bots then
...
+    if not player_force.technologies['logistics'].researched then
+        return
+    end
+    local panel = player.gui.relative
+    draw_loader_frame_for_player(panel)
+end)
+
+Event.add(defines.events.on_research_finished, function(event)
+    local research = event.research.name
+    if (research == "logistics") or (research == "logistics-2") or (research == "logistics-3") then
```
grilledham: check name is in the `loaders_technology_map` tabel declared on line 85.

features/redmew_qol.lua
```diff
@@ -339,11 +310,172 @@ if config.save_bots then
...
+
+Gui.on_click(player_craft_loader_2, function(event)
+    player_craft_loaders(event, "fast-loader")
+end)
+
+Gui.on_click(player_craft_loader_3, function(event)
+    player_craft_loaders(event, "express-loader")
+end)
+
+Gui.on_click(machine_craft_loader_1, function(event)
```
grilledham: Much like the `player_craft_loaders` function, you could make a similar function for the assembly machine click events to avoid repeating the same logic.

grilledham (changes requested):
Good stuff, some comments that need addressing.
I also pushed some code changes.
1. when `Gui.set_data` is used the data needs to be cleared with `Gui.destroy` when the gui element is removed, otherwise there is memory leak. This only applied to the assembly machine crafting frame as that is the only place `Gui.set_data` was used. As this was a problem with my original prototype, I felt I should fix it 😄.
2. Hopefully fix the luacheck issues. We need to add the defines that were added in 1.1 to the list luacheck knows about.

Jayefuu:
@grilledham updated as per your suggestions. I think I ticked them all. Retested it works.

features/redmew_qol.lua
```diff
@@ -330,20 +292,192 @@ if config.set_alt_on_create then
...
+end)
+
+Gui.on_click(player_craft_loader_3, function(event)
+    player_craft_loaders(event, "express-loader")
+end)
+
+local function set_assembly_machine_recipe(event, loader_name)
+    local entity = Gui.get_data(event.element)
+    entity.set_recipe(loader_name)
+    Task.set_timeout_in_ticks(1, close_gui_token, {player = event.player, entity = entity})
```
grilledham: `close_gui_token` does not need the entity. Either remove from the table or make a shared table that is passed to both `close_gui_token` and `open_gui_token`

features/redmew_qol.lua
```diff
@@ -330,20 +292,192 @@ if config.set_alt_on_create then
...
+
+if config.loaders then
+    Event.add(defines.events.on_player_created, function(event)
+        local player = game.get_player(event.player_index)
+        local player_force = player.force
+        if not player_force.recipes['loader'].enabled then
+            return
+        end
+        local panel = player.gui.relative
+        draw_loader_frame_for_player(panel)
```
grilledham: We might need to do this when the player opens the crafting gui rather than on player created. Thinking about the rare case where the loader recipes are enabled /disabled with a command.

In such a case the player's may still have access to the loader buttons when they shouldn't. Given that it is a rare event, I think it would be ok for the gui to not update until the next time the player opens the crafting gui. We could also put a check in the button click events that the recipe is enabled.

If we don't draw the frame on player create, the update to the gui that happens from researches could check to see if the frame is open before redrawing.

grilledham (commented):
Some late night thoughts. I haven't play tested this yet, but I should have time tomorrow.

features/redmew_qol.lua
Jayefuu: I tried to get it to work that way the first time I tried your example, but I couldn't get it to trigger when the player opens their inventory. 

I'm also getting a load of errors in my logs that I'm not sure what I changed to cause, or if I caused them. Any ideas?

`1109.503 Checksum for script C:/Users/James/AppData/Roaming/Factorio/temp/currently-playing/control.lua: 3361046532
1109.769 Script @C:/Users/James/AppData/Roaming/Factorio/temp/currently-playing/utils/event_core.lua:35: .../Factorio/temp/currently-playing/features/redmew_qol.lua:205: attempt to index field 'forces' (a nil value)
1109.769 Script @C:/Users/James/AppData/Roaming/Factorio/temp/currently-playing/utils/event_core.lua:35: .../Factorio/temp/currently-playing/features/redmew_qol.lua:302: attempt to index field 'forces' (a nil value)
1109.799 Script @C:/Users/James/AppData/Roaming/Factorio/temp/currently-playing/utils/event_core.lua:35: .../Factorio/temp/currently-playing/features/redmew_qol.lua:79: attempt to call field 'get_player' (a nil value)
1110.422 Script @C:/Users/James/AppData/Roaming/Factorio/temp/currently-playing/utils/event_core.lua:35: .../Factorio/temp/currently-playing/features/redmew_qol.lua:107: attempt to call field 'get_player' (a nil value)
1110.487 Script @C:/Users/James/AppData/Roaming/Factorio/temp/currently-playing/utils/event_core.lua:35: .../Factorio/temp/currently-playing/features/redmew_qol.lua:107: attempt to call field 'get_player' (a nil value)
1111.913 Script @C:/Users/James/AppData/Roaming/Factorio/temp/currently-playing/utils/event_core.lua:35: .../Factorio/temp/currently-playing/features/redmew_qol.lua:390: attempt to index field 'forces' (a nil value)
1113.804 Script @C:/Users/James/AppData/Roaming/Factorio/temp/currently-playing/utils/event_core.lua:35: .../Factorio/temp/currently-playing/features/redmew_qol.lua:390: attempt to index field 'forces' (a nil value)
1114.665 Script @C:/Users/James/AppData/Roaming/Factorio/temp/currently-playing/utils/event_core.lua:35: .../Factorio/temp/currently-playing/features/redmew_qol.lua:488: attempt to call field 'get_player' (a nil value)
1118.286 Script @C:/Users/James/AppData/Roaming/Factorio/temp/currently-playing/utils/event_core.lua:35: .../Factorio/temp/currently-playing/features/redmew_qol.lua:390: attempt to index field 'forces' (a nil value)
1119.566 Script @C:/Users/James/AppData/Roaming/Factorio/temp/currently-playing/utils/event_core.lua:35: .../Factorio/temp/currently-playing/features/redmew_qol.lua:488: attempt to call field 'get_player' (a nil value)
1120.306 Script @C:/Users/James/AppData/Roaming/Factorio/temp/currently-playing/utils/event_core.lua:35: .../Factorio/temp/currently-playing/features/redmew_qol.lua:107: attempt to call field 'get_player' (a nil value)
1121.546 Script @C:/Users/James/AppData/Roaming/Factorio/temp/currently-playing/utils/event_core.lua:35: .../Factorio/temp/currently-playing/features/redmew_qol.lua:390: attempt to index field 'forces' (a nil value)
1121.567 Script @C:/Users/James/AppData/Roaming/Factorio/temp/currently-playing/utils/event_core.lua:35: .../Factorio/temp/currently-playing/features/redmew_qol.lua:107: attempt to call field 'get_player' (a nil value)
1123.687 Script @C:/Users/James/AppData/Roaming/Factorio/temp/currently-playing/utils/event_core.lua:35: .../Factorio/temp/currently-playing/features/redmew_qol.lua:488: attempt to call field 'get_player' (a nil value)
1123.707 Script @C:/Users/James/AppData/Roaming/Factorio/temp/currently-playing/utils/event_core.lua:35: .../Factorio/temp/currently-playing/features/redmew_qol.lua:107: attempt to call field 'get_player' (a nil value)
1124.027 Script @C:/Users/James/AppData/Roaming/Factorio/temp/currently-playing/utils/event_core.lua:35: .../Factorio/temp/currently-playing/features/redmew_qol.lua:390: attempt to index field 'forces' (a nil value)
1124.047 Script @C:/Users/James/AppData/Roaming/Factorio/temp/currently-playing/utils/event_core.lua:35: .../Factorio/temp/currently-playing/features/redmew_qol.lua:107: attempt to call field 'get_player' (a nil value)
1189.704 Script @C:/Users/James/AppData/Roaming/Factorio/temp/currently-playing/utils/event_core.lua:35: .../Factorio/temp/currently-playing/features/redmew_qol.lua:488: attempt to call field 'get_player' (a nil value)
1189.884 Script @C:/Users/James/AppData/Roaming/Factorio/temp/currently-playing/utils/event_core.lua:35: .../Factorio/temp/currently-playing/features/redmew_qol.lua:107: attempt to call field 'get_player' (a nil value)
1189.984 Script @C:/Users/James/AppData/Roaming/Factorio/temp/currently-playing/utils/event_core.lua:35: .../Factorio/temp/currently-playing/features/redmew_qol.lua:107: attempt to call field 'get_player' (a nil value)
1190.044 Script @C:/Users/James/AppData/Roaming/Factorio/temp/currently-playing/utils/event_core.lua:35: .../Factorio/temp/currently-playing/features/redmew_qol.lua:107: attempt to call field 'get_player' (a nil value`

grilledham:
@Jayefuu I pushed some changes, so I guess it's your turn to review now.

features/redmew_qol.lua
grilledham: I'm not sure about the errors, I don't get them. If you still get them after my changes, let me know and I'll investigate.
Also check they aren't old errors, factorio doesn't always clear out the logs.

grilledham (approved):

PR #1124, opened 2020-12-23
Ping map-promotion channel on crashsite restart.

grilledham:


Jayefuu:
I see no problems with this, great addition, been looking forward to this! 👍

Jayefuu (approved):

PR #1125, opened 2020-12-23
Reports ping channel

grilledham:
Need to make the discord channel and link it.
Based on https://github.com/Refactorio/RedMew/pull/1124 to avoid merge conflicts in resources/discord_channel_names.lua.

### DIFF
--- a/resources/discord_channel_names.lua
+++ b/resources/discord_channel_names.lua
@@ -1,5 +1,0 @@
---- The names of the discord channels that can be referenced by name.
--- See features.server.to_discord_named
-return {
-    map_promotion = 'map-promotion'
-}
--- a/features/report.lua
+++ b/features/report.lua
@@ -6,6 +6,12 @@ local Command = require 'utils.command'
 local Popup = require 'features.gui.popup'
 local Color = require 'resources.color_presets'
 local Event = require 'utils.event'
+local Server = require 'features.server'
+local Discord = require 'resources.discord'
+
+local helpdesk_channel = Discord.channel_names.helpdesk
+local moderation_log_channel = Discord.channel_names.moderation_log
+local moderator_role_mention = Discord.role_mentions.moderator
 
 local format = string.format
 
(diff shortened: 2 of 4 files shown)


resources/discord_channel_names.lua
```diff
@@ -0,0 +1,7 @@
+--- The names of the discord channels that can be referenced by name.
+-- See features.server.to_discord_named
+return {
+    bot_playground ='bot-playground',
```
Jayefuu: For consistency, should these return vars have dashes instead of underscores?

Jayefuu (commented):
Would it make sense to clone the messages to a public channel so that players can see that we're seeing them and responding to them? We could put a copy of the report message, without the role ping in helpdesk? Or ping in helpdesk and put a copy without the ping in moderation log?

resources/discord_channel_names.lua
grilledham: `bot_playground` has an underscore because it's easier to use in lua that way. 'bot-playground' uses a dash to match the name shown in discord.

grilledham:
> Would it make sense to clone the messages to a public channel so that players can see that we're seeing them and responding to them? We could put a copy of the report message, without the role ping in helpdesk? Or ping in helpdesk and put a copy without the ping in moderation log?

I worry about sending unfiltered player messages to a public channel.

Edit:
That was my gut reaction. But now I think about it we do that already with server chat.

> Or ping in helpdesk and put a copy without the ping in moderation log?
That probably makes the most sense.

If we put the messages in helpdesk then there is the question of whether it should be the moderator roles that gets pinged.

Jayefuu:
> If we put the messages in helpdesk then there is the question of whether it should be the moderator roles that gets pinged.

Let's do that. It is what the role is for after all. It should have the ping in helpdesk, then a copy of the message automatically put in moderation log. Or is the fact that when it's deleted it goes into omnilog good enough?

grilledham:
> > If we put the messages in helpdesk then there is the question of whether it should be the moderator roles that gets pinged.
> 
> Let's do that. It is what the role is for after all. It should have the ping in helpdesk, then a copy of the message automatically put in moderation log. Or is the fact that when it's deleted it goes into omnilog good enough?

Changed to ping moderator.
I send a copy without the ping to moderation log.

Jayefuu (approved):

grilledham:
Rebased to fix merge conflicts.

PR #1127, opened 2020-12-24
Added additional statistics to Crash Site server scenario data

Jayefuu:
Previously we couldn't tell if a restart was due to game win or fail. Statistics are now saved that include:
- the start and end epoch
- the game time in ticks (so we can gauge activity)
- how many biters, worms and nests were killed
- how many enemy_entities remain (so we can tell if the map was cleared)
- how many players participated in total.

### DIFF
--- a/map_gen/maps/crash_site/commands.lua
+++ b/map_gen/maps/crash_site/commands.lua
@@ -21,8 +21,7 @@ local Public = {}
 function Public.control(config)
 
     local server_player = {name = '<server>', print = print}
-    local start_epoch = nil
-    local global_data = {restarting = nil, start_epoch = start_epoch}
+    local global_data = {restarting = nil, start_epoch = nil}
     local airstrike_data = {radius_level = 1, count_level = 1}
 
     Global.register({global_data = global_data, airstrike_data = airstrike_data}, function(tbl)
@@ -94,7 +93,7 @@ function Public.control(config)
 
             local end_epoch = Server.get_current_time()
             if end_epoch == nil then
-                end_epoch = -1 -- for if a save was made locally so it doesn't crash on restart
+                end_epoch = -1 -- end_epoch is nil if the restart command is used locally rather than on the server
             end
 
             local statistics = {
@@ -314,7 +313,6 @@ function Public.control(config)
     Event.add(Server.events.on_server_started, function()
         if not global_data.start_epoch then
             global_data.start_epoch = Server.get_current_time()
-            game.print("Start epoch: " .. global_data.start_epoch)
         end
     end)
 
--- a/map_gen/maps/crash_site/commands.lua
+++ b/map_gen/maps/crash_site/commands.lua
@@ -20,59 +20,61 @@ local Public = {}
 
 function Public.control(config)
 
-local server_player = {name = '<server>', print = print}
-local start_epoch = Server.get_current_time()
-if start_epoch == nil then
-    start_epoch = 0      -- for if a save was made locally so it doesn't crash
-end
+    local server_player = {name = '<server>', print = print}
+    local start_epoch = nil
+    local global_data = {restarting = nil, start_epoch = start_epoch}
+    local airstrike_data = {radius_level = 1, count_level = 1}
 
-local global_data = {restarting = nil, start_epoch = start_epoch}
-local airstrike_data = {radius_level = 1, count_level = 1}
-
-Global.register({
-    global_data = global_data,
-    airstrike_data = airstrike_data
-},
-    function(tbl)
+    Global.register({global_data = global_data, airstrike_data = airstrike_data}, function(tbl)
         global_data = tbl.global_data
         airstrike_data = tbl.airstrike_data
+    end)
+
+    local function double_print(str)
+        game.print(str)
+        print(str)
     end
-)
 
-local function double_print(str)
-    game.print(str)
-    print(str)
-end
+    local static_entities_to_check = {
+        'spitter-spawner',
+        'biter-spawner',
+        'small-worm-turret',
+        'medium-worm-turret',
+        'big-worm-turret',
+        'behemoth-worm-turret',
+        'gun-turret',
+        'laser-turret',
+        'artillery-turret',
+        'flamethrower-turret'
+    }
 
-local static_entities_to_check = {
-    'spitter-spawner','biter-spawner',
-    'small-worm-turret', 'medium-worm-turret','big-worm-turret', 'behemoth-worm-turret',
-    'gun-turret', 'laser-turret', 'artillery-turret', 'flamethrower-turret'
-}
+    local biter_entities_to_check = {
+        'small-spitter',
+        'medium-spitter',
+        'big-spitter',
+        'behemoth-spitter',
+        'small-biter',
+        'medium-biter',
+        'big-biter',
+        'behemoth-biter'
+    }
 
-local biter_entities_to_check = {
-    'small-spitter', 'medium-spitter', 'big-spitter', 'behemoth-spitter',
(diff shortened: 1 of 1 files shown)


map_gen/maps/crash_site/commands.lua
```diff
@@ -39,6 +43,31 @@ local function double_print(str)
...
+    'small-worm-turret', 'medium-worm-turret','big-worm-turret', 'behemoth-worm-turret',
+    'gun-turret', 'laser-turret', 'artillery-turret', 'flamethrower-turret'
+}
+
+local biter_entities_to_check = {
+    'small-spitter', 'medium-spitter', 'big-spitter', 'behemoth-spitter',
+    'small-biter', 'medium-biter', 'big-biter', 'behemoth-biter'
+}
+
+local function count_enemy_entities()
```
Jayefuu: I thought about making this function take the static_entities_to_check and biter_entities_to_check  as an argument so that when used later I can just call it on biter entities for the restart check. Is there a good way to allow it to take both tables as an argument or do I need to concatenate the tables first?

map_gen/maps/crash_site/commands.lua
```diff
@@ -20,8 +20,12 @@ local Public = {}
 function Public.control(config)
 
 local server_player = {name = '<server>', print = print}
+local start_epoch = Server.get_current_time()
```
Jayefuu: I haven't been able to test this works as I didn't know how to start a new scenario from the server interface. The value is saved correctly to global though.

map_gen/maps/crash_site/commands.lua
```diff
@@ -59,10 +88,20 @@ callback =
...
-            Server.set_data('crash_site_data', tostring(unix_time),  game_time) -- Store the server unix time as key and total game ticks in the Scenario Data
-
+            local end_epoch = Server.get_current_time()
+            if end_epoch == nil then
+                end_epoch = -1       -- for if a save was made locally so it doesn't crash on restart
+            end
+
+            local statistics = {
+                start_epoch = start_epoch,
+                end_epoch = end_epoch,  -- stored as key already, useful to have it as part of same structure
```
Jayefuu: Unsure if I should include end_epoch in the table. I think it could be useful later if it's in the same structure of the rest of the data, but we do already use the end_epoch as the unique key.

map_gen/maps/crash_site/commands.lua
grilledham: Sounds like a good idea.
Make the function take a table as a single argument.
Make a new variable, something like this
```lua
local all_entities_to_check = {}
all_entities_to_check = table.add_all(all_entities_to_check, static_entities_to_check)
all_entities_to_check = table.add_all(all_entities_to_check, biter_entities_to_check)
```

map_gen/maps/crash_site/commands.lua
grilledham: You want to do something different here.
`Server.get_current_time()` won't work if called during the init stage.

You need to use the `Event.add(Server.events.on_server_started, function() --[[ do stuff ]] end)` to get the start time.
Note that it may be called multiple times per save, if the server is stopped and started again, so only set the start time when it is nil. This should also work if you make a save locally and test on the server.

map_gen/maps/crash_site/commands.lua
```diff
@@ -14,14 +14,19 @@ local Toast = require 'features.gui.toast'
...
+local ScoreTracker = require 'utils.score_tracker'
 
 local Public = {}
 
 function Public.control(config)
 
 local server_player = {name = '<server>', print = print}
+local start_epoch = Server.get_current_time()
+if start_epoch == nil then
+    start_epoch = 0      -- for if a save was made locally so it doesn't crash
```
grilledham: `nil` shouldn't cause a crash for the data, only for the key.

map_gen/maps/crash_site/commands.lua
```diff
@@ -59,10 +89,20 @@ callback =
             --local discord_crashsite_role = '<@&593534612051984431>' -- @test
             Server.to_discord_named_raw(DiscordChannelNames.map_promotion, discord_crashsite_role .. ' **Crash Site has just restarted! Previous map lasted: ' .. time_string .. '!**')
 
-            local unix_time = Server.get_current_time()
-            local game_time = game.ticks_played
-            Server.set_data('crash_site_data', tostring(unix_time),  game_time) -- Store the server unix time as key and total game ticks in the Scenario Data
-
+            local end_epoch = Server.get_current_time()
+            if end_epoch == nil then
+                end_epoch = -1       -- for if a save was made locally so it doesn't crash on restart
```
grilledham: If `end_epoch` was `nil` that *probably* means restart was used locally, it which case we don't need to call `Server.set_data`.

map_gen/maps/crash_site/commands.lua
grilledham: If you think it is useful in the structure then it's fine to put it there.

grilledham (commented):
Just looking at the title on this PR, the data that is stored on the server is lua not json.

Jayefuu:
I think I resolved everything except passing tables to the count_enemy_entities function. It kept saying the table passed was nil, I expect I have some issues with scope. At present it works as intended, it depends on how much we care about wasting time checking for nests/worms/turrets when checking for the remaining biters.

map_gen/maps/crash_site/commands.lua
```diff
@@ -14,35 +14,67 @@ local Toast = require 'features.gui.toast'
...
+local ScoreTracker = require 'utils.score_tracker'
 
 local Public = {}
 
 function Public.control(config)
 
-local server_player = {name = '<server>', print = print}
+    local server_player = {name = '<server>', print = print}
+    local start_epoch = nil
+    local global_data = {restarting = nil, start_epoch = start_epoch}
```
grilledham: `local start_epoch = nil` isn't used.
```suggestion
    local global_data = {restarting = nil, start_epoch = nil}
```

map_gen/maps/crash_site/commands.lua
```diff
@@ -56,337 +88,331 @@ callback =
...
-            local game_time = game.ticks_played
-            Server.set_data('crash_site_data', tostring(unix_time),  game_time) -- Store the server unix time as key and total game ticks in the Scenario Data
-
+            -- local discord_crashsite_role = '<@&593534612051984431>' -- @test
+            Server.to_discord_named_raw(DiscordChannelNames.map_promotion, discord_crashsite_role
+                .. ' **Crash Site has just restarted! Previous map lasted: ' .. time_string .. '!**')
+
+            local end_epoch = Server.get_current_time()
+            if end_epoch == nil then
+                end_epoch = -1 -- for if a save was made locally so it doesn't crash on restart
```
grilledham: Comment is wrong, `end_epoch` is `nil` if the restart command is used locally rather than on the server. It doesn't matter where the save was made.

map_gen/maps/crash_site/commands.lua
```diff
@@ -56,337 +88,331 @@ callback =
...
+            set_timeout_in_ticks(60 * j, chart_area_callback, {player = player, xpos = xpos, ypos = ypos})
+        end
     end
-end
 
-Event.add(Retailer.events.on_market_purchase, function(event)
+    Event.add(Server.events.on_server_started, function()
+        if not global_data.start_epoch then
+            global_data.start_epoch = Server.get_current_time()
+            game.print("Start epoch: " .. global_data.start_epoch)
```
grilledham: Is this some test code that you forgot to remove?

grilledham (approved):

PR #1128, opened 2020-12-26
Crash site rocket tanks

Jayefuu:
Adds research, unlockable from the market, that allows tanks to fire rockets at enemy worms and spawners.

The intention is to make mid-game more fun and make tanks more useful, they're often ignored while players wait for spidertrons. Intentionally didn't allow rockets to target biters and spitters because the part of the game I want them to be enticing for is when we have few rockets just after we get the science to make them.

They don't target enemy turrets because I don't want to change the destroyer spawning behaviour.

The first level of tech allows rockets to be shot at a 3 second interval. Upgradable to level 4, 0.5 seconds per rocket.

Almost ready for a live test, one small feature to complete.

### DIFF
--- a/map_gen/maps/crash_site/rocket_tanks.lua
+++ b/map_gen/maps/crash_site/rocket_tanks.lua
@@ -82,8 +82,6 @@ local on_tick = Token.register(function()
     end
 end)
 
-local rocket_tank_level_intervals = {[2] = 180, [3] = 120, [4] = 60, [5] = 30}
-
 Event.add(Retailer.events.on_market_purchase, function(event)
     local market_id = event.group_name
     local group_label = Retailer.get_market_group_label(market_id)
@@ -103,7 +101,7 @@ Event.add(Retailer.events.on_market_purchase, function(event)
 
         Toast.toast_all_players(15, {'command_description.crash_site_rocket_tank_upgrade_success', interval_level})
         item.name_label = {'command_description.crash_site_rocket_tanks_name_label', (interval_level + 1)}
-        item.price = (interval_level + 1) * 7500
+        item.price = (interval_level + 1) * 4000
         Retailer.set_item(market_id, item) -- this updates the retailer with the new item values.
     end
     if interval_level >= 4 then -- update label, set price to 0, disable further purchases
@@ -114,8 +112,9 @@ Event.add(Retailer.events.on_market_purchase, function(event)
         Retailer.set_item(market_id, item)
     end
 
+    -- Interval for each level should decrease 30 from 120 ticks each level. 120, 90, 60, 30.
     if interval_level > 2 then
-        Event.remove_removable_nth_tick(rocket_tank_level_intervals[tank_research.interval_level - 1], on_tick)
+        Event.remove_removable_nth_tick((120-(((interval_level-1)-2)*30)), on_tick)
     end
-    Event.add_removable_nth_tick(rocket_tank_level_intervals[tank_research.interval_level], on_tick)
+    Event.add_removable_nth_tick((120-((interval_level-2)*30)), on_tick)
 end)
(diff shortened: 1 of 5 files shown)


map_gen/maps/crash_site/rocket_tanks.lua
```diff
@@ -0,0 +1,153 @@
...
+
+local static_entities_to_check = {
+    'spitter-spawner',
+    'biter-spawner',
+    'small-worm-turret',
+    'medium-worm-turret',
+    'big-worm-turret',
+    'behemoth-worm-turret'
+}
+--[[local function on_tick()
```
Jayefuu: Old function I used before using the removable nth tick event.

map_gen/maps/crash_site/rocket_tanks.lua
```diff
@@ -0,0 +1,153 @@
...
+                    s.create_entity {name = "rocket", position = tank.position, target = target[1], speed = 0.2}
+                    inv.remove({name = "rocket", count = 1})
+                end
+            end
+        end
+    end
+end
+)
+
+local rocket_tank_level_intervals = {
```
Jayefuu: Used a table rather than maths to determine the firing interval because I wanted some sub-one second options.

map_gen/maps/crash_site/rocket_tanks.lua
```diff
@@ -0,0 +1,153 @@
...
+
+        Toast.toast_all_players(15, {'command_description.crash_site_rocket_tank_upgrade_success', interval_level})
+        item.name_label = {'command_description.crash_site_rocket_tanks_name_label', (interval_level + 1)}
+        item.price = (interval_level+1)*7500
+        item.description = 'Upgrade the tank rocket firing rate to reduce the rocket interval.\n\nPlace rockets in the tank inventory to have them automatically target enemy worms and nests.'
+        Retailer.set_item(market_id, item) -- this updates the retailer with the new item values.
+    end
+    if interval_level == 5 then
+        -- This doesn't work....
+        Retailer.remove_item('Spawn', 'rocket_tanks_fire_rate')
```
Jayefuu: @grilledham or @SimonFlapse I need your help with Retailer.remove_item, I couldn't get it to work from the documentation and examples I found. Any help much appreciated. I think this is the last thing I need to change before I can do a live play test.

map_gen/maps/crash_site/rocket_tanks.lua
```diff
@@ -0,0 +1,153 @@
...
+            end
+        end
+    end
+end]]
+local on_tick
+on_tick = Token.register(
+    function()
+    for k, v in pairs(tank_entities) do
+        local unit_number = k
+        local tank = v
```
grilledham:
```suggestion
    for unit_number, tank in pairs(tank_entities) do
```

grilledham: I'd add a check that tank is valid, that way this doesn't depend on the order in which events are called.

map_gen/maps/crash_site/scenario.lua
```diff
@@ -787,6 +791,14 @@ local function init(config)
                 sprite = 'item-group/production',
                 description = {'command_description.crash_site_airstrike_radius', 1, 0, 5}
             },
+            {
+                price = 5000,
+                type = 'rocket_tanks',
+                name = 'rocket_tanks_fire_rate',
+                name_label = {'command_description.crash_site_rocket_tanks_name_label', 1},
+                sprite = 'item-group/production',
+                description = 'Upgrade the tank rocket firing rate to reduce the rocket interval.\n\nPlace rockets in the tank inventory to have them automatically target enemy worms and nests.'
```
grilledham: Use localised string

map_gen/maps/crash_site/rocket_tanks.lua
grilledham:
```suggestion
        Retailer.remove_item(market_id, 'rocket_tanks_fire_rate')
```

grilledham (commented):
I've had a quick look, I'll play test later.

grilledham:
I made some changes, I was going to suggest them as part of a review, but I needed to make changes to files not in the review for it to work.

grilledham (approved):

PR #1136, opened 2020-12-30
Updated price after balance testing

Jayefuu:
Changed to level * 1000 instead of level * 4000. Cost was far too high.


grilledham (approved):
Maybe that's too cheap? But we can always change it later. 
I'd probably make the cost double each time, but the rockets per second increase linearly (1/3, 2/3, 1, 4/3, ...). Just a thought.

Jayefuu:
> 
> 
> Maybe that's too cheap? But we can always change it later.
> I'd probably make the cost double each time, but the rockets per second increase linearly (1/3, 2/3, 1, 4/3, ...). Just a thought.

Yeah, but this is the third time I've adjusted it down, so I thought I'd just stick it low and see if it even gets used :P

PR #1138, opened 2020-12-31
Player data types updated and saved to server on restart

Jayefuu:
- Saves more kill data to Scoreboard
- Displays total kills in score board
- Saves player data to crash site scenario data set
- Added scenario name to data set so that if we later want to add data for danger ores we can filter it out


features/player_stats.lua
```diff
@@ -137,9 +174,30 @@ local function robot_built_entity()
...
+local function entity_died(event)
+    local entity = event.entity
+    if not entity or not entity.valid or entity.force.name ~= 'enemy' then
+        return
+    end
+
+    change_for_global(aliens_killed_name, 1)
+
+    local cause = event.cause
+    if not cause or not cause.valid or cause.name ~= 'character' then -- What about players inside vehicles?
```
grilledham: I left this comment
> What about players inside vehicles?

Do we want to record this too, if so I've left it as an exercise for the reader.

features/player_stats.lua
```diff
@@ -21,6 +21,11 @@ local player_console_chats_name = 'player-console-chats'
 local player_items_crafted_name = 'player-items-crafted'
 local player_distance_walked_name = 'player-distance-walked'
 local satellites_launched_name = 'satellites-launched'
+local player_units_killed_name = 'player-units-killed' -- biters and spitters
+local player_worms_killed_name = 'player-worms-killed'
+local player_spawners_killed_name = 'player-spawners-killed'
+local player_total_kills_name = 'player-total-kills'
+local player_turrets_killed_name = 'player-turrets-killed'
```
grilledham, line 28: I wonder if we should expose these variables publicly, so they can be used in command.lua.
But don't worry if it's too much effort.

grilledham:
I made some changes, but I must confess I was lazy and did minimal testing.

features/gui/player_list.lua
```diff
@@ -498,16 +531,17 @@ local function get_default_player_settings()
...
-        offset = 7
+        table.insert(columns, 6, coin_heading_name)
     end
-    columns[offset] = deaths_heading_name
-    columns[offset + 1] = poke_name_heading_name
-    columns[offset + 2] = report_heading_name
+
     return {
         columns = columns,
         sort = -3
```
grilledham, line 543: You changed this value to -4, so I changed it back to -3, did you mean to change it?

features/player_stats.lua
```diff
@@ -137,9 +174,30 @@ local function robot_built_entity()
...
+    -- Only store for players if its an allowed entity (ie not for walls, chests etc)
+    local category = entity_to_category_map[entity.name]
+    if category == nil then
+        return
+    end
+
+    local player_index = cause.player.index
+    change_for_player(player_index, player_total_kills_name, 1)
+
+    if category then
```
Jayefuu: Why do this? We already return if category is nil.....

grilledham: category could be `false`. At least the way I setup the table.

map_gen/maps/crash_site/commands.lua
```diff
@@ -90,22 +90,42 @@ function Public.control(config)
...
+                    distance_walked = ScoreTracker.get_for_player(p.index, 'player-distance-walked'),
+                    player_deaths = ScoreTracker.get_for_player(p.index, 'player-deaths'),
+                    coins_earned = ScoreTracker.get_for_player(p.index, 'coins-earned'),
+                    entities_built = ScoreTracker.get_for_player(p.index, 'built-by-players'),
+                    time_played = p.online_time
+                }
+            end
+
             local statistics = {
+                scenario = data.scenario_name,
```
```suggestion
                scenario = config.scenario_name,
```

`data.scenario_name` is the scenario that is going to be started. Could be different if an admin used the restart command.

features/player_stats.lua
Jayefuu: Fixed! I haven't managed deaths caused by manually driving trains but I don't think that's necessary.

features/player_stats.lua
Jayefuu: How do I do this and what's the advantage?

features/gui/player_list.lua
Jayefuu: Woops.

grilledham:
I made some changes:

- Refactored the get player index from cause logic. Moved into a function and reordered to reduce the amount of nested if blocks.
- Fixed the merge conflict.
- Made the player stat names public.

features/player_stats.lua
grilledham: I made the change, so you can see it in the commit.

The advantage is that if the same values are declared separately they can go out of sync, which if that ever happened would be a bug. So it's better to have a single source of truth for values that are logically the same thing.

features/player_stats.lua
grilledham: Nice!

I think that is much better, we can always add the maunal trains kills later if we want.

grilledham (approved):

PR #1139, opened 2020-12-31
Fix loader gui layout issue in multiplayer

grilledham:
- Change the loader gui logic to be more conservative about when it deletes the loader frames. Now the layout issue should only happen when the player goes from having no loader recipes unlocked to any or vice versa Which in most cases will only happen once per game.


Jayefuu (approved):
Looks fine, can't spot anything wrong.

PR #1144, opened 2021-01-03
Crash Site Arrakis

Jayefuu:
- Drier than desert. No trees or oases so wood and fish must be bought from the market.
- Placing roboports attracts sandworms. They will spawn every 6 to 12 minutes to attack each roboport so networks must be kept small or protected well.
- Repair cars. Upon placing repair packs in a car it can fire a repair beam at nearby damaged entities.


map_gen/maps/crash_site/features/repair_cars.lua
```diff
@@ -0,0 +1,95 @@
...
+    register_car(event.created_entity)
+end)
+
+Event.add(defines.events.on_built_entity, function(event)
+    register_car(event.created_entity)
+end)
+
+Event.add(defines.events.on_entity_cloned, function(event)
+    register_car(event.destination)
+end)
```
grilledham: I wonder if the <https://lua-api.factorio.com/latest/events.html#on_player_driving_changed_state> event would be better to use (Same for rocket tanks).

Assuming the event does what I think it does, it would mean a smaller list of entities to iterate through and we could remove the get_driver check.

map_gen/maps/crash_site/features/repair_cars.lua
```diff
@@ -0,0 +1,95 @@
...
+
+        local repair_amount = 150
+        for i, entity in pairs(targets) do
+            if entity.unit_number and (entity.get_health_ratio() or 1) < 1 then
+                -- Rules for when car can repair:
+                -- if the damaged entity is a turret, the turret cooldown must be complete (entity.active == true) to help reduce turret creeping
+                -- if the damaged entity is not the car. Cars can heal other cars but not themselves.
+                -- if the damaged entity is not a character
+                -- if the entity is not moving. Vehicles must have a speed of 0, entities that can't move will be nil
+                if (entity ~= car and (entity.speed == 0 or entity.speed == nil) and entity.name ~= 'character' and not turrets[entity.name]) or (turrets[entity.name] and entity.active == true) then
```
grilledham: Normal grumble about repeating the same table lookups 😛

map_gen/maps/crash_site/features/repair_cars.lua
```diff
@@ -0,0 +1,95 @@
...
+                        speed = 1,
+                        duration = 20
+                    }
+                    local max_health = entity.prototype.max_health
+                    if (max_health - entity.health) < repair_amount then
+                        repair_amount = max_health - entity.health -- so that the player doesn't lose part of a repair pack partially healing an entity
+                    end
+                    entity.health = entity.health + repair_amount
+                    stack.drain_durability(repair_amount)
+                    goto continue -- break out because we only want to heal one entity
```
grilledham: Could probably use `break` here but this is fine.

map_gen/maps/crash_site/features/sandworms.lua
```diff
@@ -0,0 +1,114 @@
...
+    end
+    for index, worm_type in pairs(sandworms) do
+        -- Determine which worm type to spawn based on the evolution
+        if (evolution > worm_type.evo_min) and (evolution <= worm_type.evo_max) then
+            local s = entity.surface
+            local worm_position = {
+                entity.position.x + math.random(max_worm_spawn_radius * -1, max_worm_spawn_radius),
+                entity.position.y + math.random(max_worm_spawn_radius * -1, max_worm_spawn_radius)
+            }
+            worm_position = s.find_non_colliding_position(index, worm_position, 5, 1)
```
grilledham: `find_non_colliding_position` can return `nil`, so we should check.

map_gen/maps/crash_site/features/sandworms.lua
```diff
@@ -0,0 +1,114 @@
...
+            s.create_entity {name = index, position = worm_position, force = "enemy"}
+            -- For the appropriate worm for each evolution region, spawn some accompanying biters to attack the roboport
+            for worm, _ in pairs(sandworm_biters) do
+                if worm == index then
+                    for biters, data in pairs(sandworm_biters[worm]) do
```
grilledham: Isn't `sandworm_biters[worm]` the value you discarded in the above for loop?

grilledham: The key called `biters` is really the name of a single biter, so I think it should be just `biter`.

map_gen/maps/crash_site/features/sandworms.lua
```diff
@@ -0,0 +1,114 @@
...
+                        local amount = math.random(data.min, data.max)
+                        local extra_chance = amount % 1
+                        if extra_chance > 0 then
+                            if math.random(0,1) <= extra_chance then
+                                amount = math.ceil(amount)
+                            else
+                                amount = math.floor(amount)
+                            end
+                        end
+                        for _ = amount,1,-1 do
```
grilledham: If you don't use the index, isn't this the same as `for _ = 1, amount do`

map_gen/maps/crash_site/features/sandworms.lua
```diff
@@ -0,0 +1,114 @@
...
+                           local pos = s.find_non_colliding_position(biters, worm_position, 5, 1)
```
grilledham: `find_non_colliding_position` could return nil

map_gen/maps/crash_site/presets/arrakis.lua
```diff
@@ -0,0 +1,39 @@
...
+            }
+        }
+    }
+}
+
+local Scenario = require 'map_gen.maps.crash_site.scenario'
+ScenarioInfo.set_map_name('Crashsite Arrakis')
+ScenarioInfo.set_map_description('Capture outposts and defend against the biters. Even drier than desert, sandworms roam the desert and will attack roboports on sight.')
+ScenarioInfo.add_map_extra_info(
+    '- Arrakis is even drier than crash site Desert.\n- Sandworms are attracted to the vibration caused by roboports and will spawn intermittently to neutralise this threat to their peace\n- Cars have repair beams\n- Outposts have enemy turrets defending them.\n- Outposts have loot and provide a steady stream of resources.\n- Outpost markets to purchase items and outpost upgrades.\n- Capturing outposts increases evolution.\n- Reduced damage by all player weapons, turrets, and ammo.\n- Biters have more health and deal more damage.\n- Biters and spitters spawn on death of entities.'
```
grilledham: Using multi line string `[[...]]` see <https://www.lua.org/pil/2.4.html> might make these easier to read.

grilledham (commented):
I had a quick look, haven't play tested yet.

map_gen/maps/crash_site/features/repair_cars.lua
Jayefuu: That makes good sense. It will be more overhead when players/passengers get in and out of vehicles but less to check per tick.

map_gen/maps/crash_site/features/repair_cars.lua
Jayefuu: I think I've fixed this in the upcoming commit. I created vars before the if statement then used those in the if statement conditions.

map_gen/maps/crash_site/features/repair_cars.lua
Jayefuu: Thanks.

map_gen/maps/crash_site/features/sandworms.lua
Jayefuu: Fixed

map_gen/maps/crash_site/features/sandworms.lua
Jayefuu: Ahhh yes. I've added biters instead of _ and sandworm_biters[worm]. That helped me understand tables better, thanks.

Jayefuu: Fixed.

map_gen/maps/crash_site/features/sandworms.lua
Jayefuu: Yes. Thanks. Fixed. Been programming in other languages a lot and couldn't remember the correct syntax. Used the first thing I found in google.

map_gen/maps/crash_site/features/sandworms.lua
Jayefuu: Updated.

map_gen/maps/crash_site/presets/arrakis.lua
Jayefuu: All updated in next commit.

map_gen/maps/crash_site/features/repair_cars.lua
```diff
@@ -0,0 +1,89 @@
...
+
+Event.add(defines.events.on_entity_destroyed, function(event)
+    car_entities[event.unit_number] = nil
+end)
+
+Event.add(defines.events.on_player_driving_changed_state, function(event)
+    local entity = event.entity
+    if (entity.name ~= 'car') or not entity or not entity.valid then
+        return
+    else
```
grilledham: Don't need an `else` after a `return`

map_gen/maps/crash_site/presets/arrakis.lua
```diff
@@ -0,0 +1,46 @@
...
+ScenarioInfo.set_map_description('Capture outposts and defend against the biters. Even drier than desert, sandworms roam the desert and will attack roboports on sight.')
+ScenarioInfo.add_map_extra_info(
+    '- Arrakis is even drier than crash site Desert.\n'
+    .. '- Sandworms are attracted to the vibration caused by roboports and will spawn intermittently to neutralise this threat to their peace\n'
+    .. '- Cars have repair beams\n'
+    .. '- Outposts have enemy turrets defending them.\n'
+    .. '- Outposts have loot and provide a steady stream of resources.\n'
+    .. '- Outpost markets to purchase items and outpost upgrades.\n'
+    .. '- Capturing outposts increases evolution.\n- Reduced damage by all player weapons, turrets, and ammo.\n'
+    .. '- Biters have more health and deal more damage.\n- Biters and spitters spawn on death of entities.'
```
grilledham: That's not what I meant by multi line strings 😄 
```suggestion
[[
- Arrakis is even drier than crash site Desert.
- Sandworms are attracted to the vibration caused by roboports and will spawn intermittently to neutralise this threat to their peace
- Cars have repair beams
- Outposts have enemy turrets defending them.
- Outposts have loot and provide a steady stream of resources.
- Outpost markets to purchase items and outpost upgrades.
- Capturing outposts increases evolution.\n- Reduced damage by all player weapons, turrets, and ammo.
- Biters have more health and deal more damage.\n- Biters and spitters spawn on death of entities.]]
```

Jayefuu:
So sandworms aren't working properly in live. I need to re-test with some debug.

Jayefuu:
Ok they're working fine, but the non_colliding radius needs turning up for in base

map_gen/maps/crash_site/features/sandworms.lua
```diff
@@ -0,0 +1,133 @@
...
+    if event.created_entity.name == 'roboport' then
+        start_worm_attacks(event.created_entity)
+        local player = game.get_player(event.player_index)
+        player.print("A sandworm approaches.....")
+    end
+end)
+
+Event.add(defines.events.on_entity_cloned, function(event)
+    if event.created_entity.name == 'roboport' then
+        start_worm_attacks(event.created_entity)
```
grilledham: on_entity_cloned event doesn't have a created_entity property.
Also when is on_entity_cloned raised, is it a map editor thing?

map_gen/maps/crash_site/features/sandworms.lua
```diff
@@ -0,0 +1,133 @@
...
+                    end
+                end
+            end
+        end
+    end
+end
+
+local thump_text_callback
+thump_text_callback = Token.register(function(entity)
+    local s = entity.surface
```
grilledham, line 90: Need an entity.valid check.

map_gen/maps/crash_site/features/sandworms.lua
```diff
@@ -0,0 +1,133 @@
...
+        local callback_timer = math.random(min_worm_period_secs * 60, max_worm_period_secs * 60)
+        set_timeout_in_ticks(callback_timer, worm_callback, entity)
+        for i = 1, 5 do -- repeat thump text 30 seconds before worm will arrive to give players visual reminder they need to protect roboports
+            set_timeout_in_ticks(callback_timer-1800+(120*i), thump_text_callback, entity)
+        end
+    end
+end)
+
+local function start_worm_attacks(entity)
+    if not entity then
```
grilledham: Is this supposed to be an entity.valid check instead?

map_gen/maps/crash_site/features/sandworms.lua
```diff
@@ -0,0 +1,133 @@
...
+local thump_text_callback
+thump_text_callback = Token.register(function(entity)
+    local s = entity.surface
+    local entity_position = entity.position
+    s.create_entity{name="flying-text", position={(entity_position.x + math.random(-3,3)),(entity_position.y + math.random(0,3))}, text="*thump*", color={r=0.6,g=0.4,b=0}}
+end)
+
+local worm_callback
+worm_callback = Token.register(function(entity)
+    if entity then -- stops the callback if the roboport has been removed
```
grilledham: Is this supposed to be an entity.valid check instead?

map_gen/maps/crash_site/features/sandworms.lua
```diff
@@ -0,0 +1,133 @@
...
+end
+
+Event.add(defines.events.on_robot_built_entity, function(event)
+    if event.created_entity.name == 'roboport' then
+        start_worm_attacks(event.created_entity)
+    end
+end)
+
+Event.add(defines.events.on_built_entity, function(event)
+    if event.created_entity.name == 'roboport' then
```
grilledham: The build events need to check the entity is valid. 
There can be multiple event handlers per event and a handler that runs before this handler might have destroyed the entity which would make it invalid. That was a difficult bug to track down we once had ~3years ago.

Jayefuu:
I think I've resolved everything you raised. :)

map_gen/maps/crash_site/features/sandworms.lua
grilledham: I think maybe you missed this one, It doesn't seem to be resolved.

map_gen/maps/crash_site/features/vehicle_repair_beams.lua
```diff
@@ -0,0 +1,89 @@
...
+    car_entities = tbl.car_entities
+end)
+
+Event.add(defines.events.on_entity_destroyed, function(event)
+    car_entities[event.unit_number] = nil
+end)
+
+Event.add(defines.events.on_player_driving_changed_state, function(event)
+    local entity = event.entity
+    if (entity.type ~= 'car') or not entity or not entity.valid then
```
grilledham: Probably should do the nil check before trying to access the property.
```suggestion
    if not entity or not entity.valid or (entity.type ~= 'car') then
```
Edit: Should probably do the valid check before the type check too.

map_gen/maps/crash_site/features/vehicle_repair_beams.lua
```diff
@@ -0,0 +1,89 @@
...
+
+        local repair_amount = 150
+        for i, entity in pairs(targets) do
+            if entity.unit_number and (entity.get_health_ratio() or 1) < 1 then
+                -- Rules for when cars/tanks can repair:
+                -- if the damaged entity is a turret, the turret cooldown must be complete (entity.active == true) to help reduce turret creeping
+                -- if the damaged entity is not the car. Cars can heal other cars but not themselves.
+                -- if the damaged entity is not a character
+                -- if the entity is not moving. Vehicles must have a speed of 0, entities that can't move will be nil
+                if (entity ~= car and (entity.speed == 0 or entity.speed == nil) and entity.name ~= 'character' and not turrets[entity.name]) or (turrets[entity.name] and entity.active == true) then
```
grilledham, line 65: We did discuss this on discord. Are there any entities that are inactive other than the turrets?
I'm wondering if the turret check is needed at all. I think you might only need the active check.

map_gen/maps/crash_site/features/sandworms.lua
```diff
@@ -0,0 +1,136 @@
...
+    end
+    local callback_timer = math.random(min_worm_period_secs * 60, max_worm_period_secs * 60)
+    set_timeout_in_ticks(callback_timer, worm_callback, entity)
+    for i = 1, 5 do
+        set_timeout_in_ticks(120*i, thump_text_callback, entity)
+    end
+end
+
+Event.add(defines.events.on_robot_built_entity, function(event)
+    if event.created_entity.name == 'roboport' and event.created_entity.valid then
```
grilledham: Needs an entity.valid check for the same reason as the on_built_entity event does.

map_gen/maps/crash_site/features/sandworms.lua
```diff
@@ -0,0 +1,136 @@
...
+        start_worm_attacks(event.created_entity)
+    end
+end)
+
+Event.add(defines.events.on_built_entity, function(event)
+    if event.created_entity.name == 'roboport' and event.created_entity.valid then
```
grilledham: The valid check should be before accessing the properties of the entity.

map_gen/maps/crash_site/features/sandworms.lua
Jayefuu: It seems this is only raised when cloning entities from one surface to another. As such I've removed it.

map_gen/maps/crash_site/features/vehicle_repair_beams.lua
Jayefuu: Belts, splitters, underground belts, chests and other entities can have active==false, and I want to repair those, so I need to check that it's a turret and that it's active so that it repairs active turrets but not while they're on turret cooldown.

Jayefuu:
Ok I think I actually got them all this time @grilledham  :D

map_gen/maps/crash_site/features/sandworms.lua
```diff
@@ -0,0 +1,130 @@
...
+            local worm_position = {
+                entity.position.x + math.random(max_worm_spawn_radius * -1, max_worm_spawn_radius),
+                entity.position.y + math.random(max_worm_spawn_radius * -1, max_worm_spawn_radius)
+            }
+            worm_position = s.find_non_colliding_position(index, worm_position, 20, 1)
+            if worm_position then
+                s.create_entity {name = index, position = worm_position, force = "enemy"}
+            end
+            -- For the appropriate worm for each evolution region, spawn some accompanying biters to attack the roboport
+            for worm, biters in pairs(sandworm_biters) do
```
grilledham, line 59: Why do you need to iterate this table, can't you just get the biters table using `local biters = sandworm_biters[index]`?

map_gen/maps/crash_site/features/sandworms.lua
```diff
@@ -0,0 +1,130 @@
...
+Event.add(defines.events.on_robot_built_entity, function(event)
+    if event.created_entity.valid and event.created_entity.name == 'roboport' then
+        start_worm_attacks(event.created_entity)
+    end
+end)
+
+Event.add(defines.events.on_built_entity, function(event)
+    if event.created_entity.valid and event.created_entity.name == 'roboport' then
+        start_worm_attacks(event.created_entity)
+        local player = game.get_player(event.player_index)
```
grilledham, line 127: Probably not needed, but a player not nil or invalid check would be good.

map_gen/maps/crash_site/features/vehicle_repair_beams.lua
grilledham: Ok, that's good to know, I think this is fine then.

grilledham (approved):

PR #1146, opened 2021-01-08
Added player_built and player_crafted to the crash site awards ceremony

Jayefuu:


features/player_stats.lua
```diff
@@ -166,12 +168,23 @@ local function player_console_chat(event)
...
 
-local function robot_built_entity()
+local function robot_built_entity(event)
     change_for_global(built_by_robots_name, 1)
+    local entity = event.created_entity
+    local player = entity.last_user
+    -- When item gets built, add to the total entities built for the player that placed the ghost
+    if player.index then
+        change_for_player(player.index, player_entities_built_name, 1)
+    end
```
grilledham: Does this mean if the the entity dies and is rebuilt by a robot this counts as another player built entity?

features/player_stats.lua
```diff
@@ -166,12 +168,23 @@ local function player_console_chat(event)
     end
 end
 
-local function player_built_entity()
+local function player_built_entity(event)
     change_for_global(built_by_players_name, 1)
+    local player_index = event.player_index
+    if player_index and (event.created_entity.is_registered_for_construction() == false)then -- only register it as built once a robot builds it
```
grilledham: Two things:
`entity.is_registered_for_construction()` looks cool, I didn't know about that, in the past we have ignored entities called 'entity-ghost' or 'tile-ghost' but this might be better.

> --only register it as built once a robot builds it

I wonder if we care. Also this will mean the player entities built wont add up to the global count. Maybe we should just count it before it's actually built. Or maybe the answer is to do the same check before adding to global.

Jayefuu: @grilledham  I think we should do the same check for global. The reason I added the check is that at the moment, no one cares about the global count as we don't use it. As we start adding player statistics, people will be come interested in having the best stat and they'll find ways to increase it. I don't want people spamming blueprints that don't get built, then undoing them, etc, just to increase their entity built count.

features/player_stats.lua
Jayefuu: Yup! But I can't see a way for players to abuse that. It will also give a wrong result if player A builts an assembler. Player B changes the reciple then it gets destroyed. Then player C or the factory bots rebuild it.

on_robot_built_entity contains robot (luaEntity), I couldn't find anything that might tell us who it belongs to though. If it did we could use that. Please can you have a look and see if you can see a way to see who a robot belongs to?

Jayefuu: robot.logistic_cell or robot.is_entity_with_owner will do it! I shall add that check so I know who built it.

Jayefuu:
@grilledham  sorry I think I'll need some help with this. Or at the very least a good discussion on what behaviour we want. I can't decide what or how to do it.

map_gen/maps/crash_site/commands.lua
```diff
@@ -202,7 +224,7 @@ function Public.control(config)
                 .. 'Players: '..statistics.total_players..'\\n'
                 )
             end
-            Server.to_discord_named_raw(map_promotion_channel, crash_site_role_mention .. ' **Crash Site has just restarted!!**')
+            Server.to_discord_named_raw(map_promotion_channel, crash_site_role_mention .. ' **'..scenario_display_name[next_scenario]..' has just restarted!!**')
```
grilledham: `next_scenario` isn't necessarily the same as the scenario that has just finished. For example in the case an admin uses the restart command.
On line 498 we have `local default_name = config.scenario_name or 'crashsite'` so I guess we should use that.

features/player_stats.lua
```diff
@@ -166,12 +169,24 @@ local function player_console_chat(event)
     end
 end
 
-local function player_built_entity()
-    change_for_global(built_by_players_name, 1)
+local function player_built_entity(event)
+    local player_index = event.player_index
+    if player_index and (event.created_entity.is_registered_for_construction() == false)then -- only register it as built once a robot builds it
```
grilledham: I don't think `player_index` can be `nil` here.

grilledham: The comment could be a bit clearer, maybe something like 'When is_registered_for_construction() is true we only register the entity as built once a robot builds it.'

features/player_stats.lua
```diff
@@ -166,12 +169,24 @@ local function player_console_chat(event)
...
+        change_for_global(built_by_players_name, 1)
+        change_for_player(player_index, player_entities_built_name, 1)
+    end
 end
 
-local function robot_built_entity()
+local function robot_built_entity(event)
     change_for_global(built_by_robots_name, 1)
+    local robot = event.robot
```
grilledham, line 182: It kinda dumb but I'd feel better if there was a check that robot was valid. Therorectailly a different event handler that runs before this could use the event to destroy the robot.

grilledham (commented):
Had a play test I think this is mostly good, just a few minor comments.

features/player_stats.lua
```diff
@@ -166,12 +169,28 @@ local function player_console_chat(event)
...
+
+    if not robot then
```
grilledham:
```suggestion
    if not robot.valid then
```
I was worried about another event handler destroying the robot, not the field being removed from the event table.

map_gen/maps/crash_site/commands.lua
```diff
@@ -202,7 +224,7 @@ function Public.control(config)
                 .. 'Players: '..statistics.total_players..'\\n'
                 )
             end
-            Server.to_discord_named_raw(map_promotion_channel, crash_site_role_mention .. ' **Crash Site has just restarted!!**')
+            Server.to_discord_named_raw(map_promotion_channel, crash_site_role_mention .. ' **'..scenario_display_name[config.scenario_name]..' has just restarted!!**')
```
grilledham: We should use the `local default_name` variable defined on line 498 (it will have to be moved higher up). Because it handles the case of `config.scenario_name` being `nil`.

Jayefuu: Thanks. Fixed both.

grilledham (approved):

PR #1148, opened 2021-01-08
Robust(?) artillery fix

Jayefuu:
A work in progress.

Needs some discussion some time.

Plan is to register artillery in global when it is picked up and create a callback to delete it again 3 seconds later.

Then if an enemy artillery dies and there is no cause and no registered artillery, no bots spawn.
If an enemy artillery dies and a player has recently picked up an artillery, send the bots at the player.

Could go badly wrong if a player picks up one artillery and biters kill another soon after. Only realised this after I wrote the code so will try think of a more robust way to stop artillery cheesing.

EDIT: I've looked at the following events...
on_trigger_fired_artillery - This can't be used because the default prototype for artillery-turret has trigger_fired_artillery == false and I can't change prototypes in vanilla. I hoped to set entity.minable and then show a cooldown arc for when it can be successfully mined.

on_pre_player_mined_item - This won't work because I can't set minable = false when the event has already triggered. I hoped to set a cooldown and then do entity.mine(player.inventory).

Other thoughts:
- If I could detect the start of mining I could set the manual_mining_speed_modifier=-0.9 so that they mine it really slowly, that would give the projectile time to arrive and ensure event.cause is not nil.
- I could try entity.clone() when on_pre_player_mined_item is called
- I could randomly add health to the artillery before it's destroyed so that it never takes the same number of shells to kill one. Then they wouldn't know when to pick it up?

### DIFF
--- a/map_gen/maps/crash_site/events.lua
+++ b/map_gen/maps/crash_site/events.lua
@@ -4,6 +4,7 @@ local Token = require 'utils.token'
 local Global = require 'utils.global'
 local math = require 'utils.math'
 local table = require 'utils.table'
+local game = require "utils.game"
 
 local random = math.random
 local set_timeout_in_ticks = Task.set_timeout_in_ticks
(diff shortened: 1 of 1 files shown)


Jayefuu:
Instead of trying to find where artillery was fired from if the death has no cause, could I just add a bit of health on to the dead entity instead? That way there'd be no point in ripping up the artillery if the final hit didn't kill the target....

Jayefuu:
Or can I use the on_pre_player_mined_item event in some way to lengthen or randomise the deconstruction time of the friendly artillery somehow?

Jayefuu:
OR. Every time an artillery turret fires, it sets the entity.mineable to false and starts a 5 second cooldown before entity.mineable becomes true again. I like this one best.

Jayefuu:
Closing this because I can't think of a good way to do this right now and it's not that important.

PR #1149, opened 2021-01-10
Change biter corpse remover.

grilledham:
The biter corpse remover was attempting to be clever about removing corpses and only remove them when they exceeded a certain threshold per chunk. For some reason it is no longer working and I can't figure out why.
So I replaced it with a simpler approach. Now we keep track of the corpses in a queue. When the queue size exceeds the threshold we remove and destroy the oldest corpse.


Jayefuu:
Looks fine to me! Much simpler approach, I like it.

Jayefuu (approved):

PR #1150, opened 2021-01-11
Updated stupid wooden chest in crash site

Jayefuu:
The wooden chest is frustrating end game on crash site because you can't fit enough loaders around it. By changing to steel chest it's more usable over time.


grilledham:
> The wooden chest is frustrating end game on crash site because you can't fit enough loaders around it.

Why does the chest change how many loaders you can use?

grilledham (approved):

Jayefuu:
> 
> 
> > The wooden chest is frustrating end game on crash site because you can't fit enough loaders around it.
> 
> Why does the chest change how many loaders you can use?

It doesn't. But you can only fit 2 for both so having a bigger chest lets it refill to more items.

PR #1151, opened 2021-01-12
Updated corpse remover to only remove biter corpses

Jayefuu:


features/biter_corpse_remover.lua
```diff
@@ -17,8 +17,19 @@ Global.register(corpse_queue, function(tbl)
...
+    ['behemoth-biter'] = true,
+    ['small-spitter'] = true,
+    ['medium-spitter'] = true,
+    ['big-spitter'] = true,
+    ['behemoth-spitter'] = true
+}
+
 local function entity_died(event)
-    if not event.unit_number then
+    if not event.unit_number or not enemy_units[event.prototype.name] then
```
grilledham: I wonder if we want to include worms and spawners too.
Or we include everything that is not trees, rocks or player corpses. We could probably look at the prototype.type to determine that.
My thinking is that building wreckage can be annoying too, but maybe not as much as you can build over it.

Jayefuu:
I thought we just cared about the biters, though perhaps we should add worms due to Hydra. 

Instead of my table, perhaps we should do `if (event.prototype.type ~= unit) or (event.prototype.type ~= turret) then`?

features/biter_corpse_remover.lua
Jayefuu: I thought we just cared about the biters, though perhaps we should add worms due to Hydra. 

Instead of my table, perhaps we should do `if (event.prototype.type ~= unit) or (event.prototype.type ~= turret) then`?

grilledham: Ok, let's do biters and worms. We can always extend it later. I made the changes.

grilledham (approved):

PR #1152, opened 2021-01-13
Gradient danger ores

grilledham:


Jayefuu (approved):
I can't see any problems. Approved.

PR #1153, opened 2021-01-19
Added white text above player when mentioned in chat

Jayefuu:
The notification sound for text chat mentions isn't useful if player is deaf or has sound off.

Recently we've had lots of players not reading the chat when they're being spoken to and often results in jailing or popups while we get their attention.

This commit adds white text above their player character (visible only to them) that disappears after 3 seconds.


Jayefuu:
image

features/chat_triggers.lua
```diff
@@ -100,8 +100,11 @@ local function mentions(event)
                         success = true
                         break
                     end
+
                     p.print(prefix .. game.get_player(event.player_index).name .. ' mentioned you!', Color.yellow)
                     p.play_sound {path = 'utility/new_objective', volume_modifier = 1}
+                    local message = game.get_player(event.player_index).name .. ' mentioned you in the chat!'
```
SimonFlapse: Could you by any chance transfer this string to the locales for translation?

Same with line 104

features/chat_triggers.lua
```diff
@@ -100,8 +100,11 @@ local function mentions(event)
                         success = true
                         break
                     end
+
                     p.print(prefix .. game.get_player(event.player_index).name .. ' mentioned you!', Color.yellow)
```
SimonFlapse: Could you by any chance transfer this string to the locales for translation?

Same with line 106

SimonFlapse (approved):
LGTM

features/chat_triggers.lua
Jayefuu: Yup no problem. Will do.

features/chat_triggers.lua
grilledham:
```suggestion
                    local message = player.name .. ' mentioned you in the chat!'
```
And the same for line 104 above.

features/chat_triggers.lua
```diff
@@ -100,8 +100,11 @@ local function mentions(event)
...
                     p.play_sound {path = 'utility/new_objective', volume_modifier = 1}
+                    local message = game.get_player(event.player_index).name .. ' mentioned you in the chat!'
+                    rendering.draw_text({text =  message, target_offset  = {0,-3},surface = p.surface, target = p.character, time_to_live = 180, alignment="center", scale=2, players={p.name}, color= {1, 1, 1, 1}})
```
grilledham: What happens if `p.character` is `nil`?

grilledham (changes requested):
Cool feature.

Some thoughts:
- For small messages (< 250 characters?) it might be better to print the message itself, rather than a message that you have a message.
- Use the chat color of the sending player.

Jayefuu:
@grilledham 
- I'm against using the colour of the sending player. White is universally visible whereas some players choose hard to read colours. If I'm adding this as an accessibility feature I'd prefer it to always be the same colour.
- re. short messages less than 250 characters appearing directly over the character, I'm not sure. I could code it up to try it but I think it will be confusing. Also if someone isn't replying and we ping them, we might just use their name to get their attention so we don't have to repeat several lines of conversation. For example asking a player not to do something, we tell them why and discuss among ourselves, then someone pings their name to get them to read the chat. This nicely tells them their attention is required.
- I'll fix the p.character issue and player.name stuff this evening :)

features/chat_triggers.lua
Jayefuu: Done.

features/chat_triggers.lua
Jayefuu: Done.

features/chat_triggers.lua
Jayefuu: I've added a check for this now. The player won't get white text when they're dead, but it would have errored before. I think this is okay since if they're dead they'll probably be bored and looking at the chat.

Jayefuu:
@grilledham I think that's everything....

features/chat_triggers.lua
```diff
@@ -100,8 +100,15 @@ local function mentions(event)
                         success = true
                         break
                     end
-                    p.print(prefix .. game.get_player(event.player_index).name .. ' mentioned you!', Color.yellow)
+
+                    --p.print({'chat_triggers.mention_success_target', prefix, game.get_player(event.player_index).name}, Color.yellow)
```
grilledham: Can you remove this commented out code.

features/chat_triggers.lua
```diff
@@ -100,8 +100,15 @@ local function mentions(event)
...
+                    p.print({'chat_triggers.mention_success_target', prefix, player.name}, Color.yellow)
                     p.play_sound {path = 'utility/new_objective', volume_modifier = 1}
+                    if p.character and p.character.valid then   -- If player is dead and they don't have a character then they won't get the hovering notification.
+                        --local message = {'chat_triggers.mention_success_target_floating', game.get_player(event.player_index).name}
```
grilledham: Can you remove this commented out code here too.

grilledham (approved):
Looks good. You seems to left in the old code commented out. After you remove that you can merge.

PR #1154, opened 2021-01-20
Made strike command less spammy, replaced map pings with tags

Jayefuu:
/strike command now only messages the player that uses it with the location. Everyone else will only see the poison capsule tags appear on the map along with the name of the player that used them. These will timeout after 30 seconds.

You can't place a tag on an uncharted part of the map so the code waits 1 second for it to be charted before placing the map tag.

### DIFF
--- a/map_gen/maps/crash_site/commands.lua
+++ b/map_gen/maps/crash_site/commands.lua
@@ -338,7 +338,7 @@ function Public.control(config)
     end)
 
     local map_chart_tag_clear_callback = Token.register(function(tag)
-        if not tag then
+        if not tag or not tag.valid then
             return -- in case a player deleted the tag manually
         end
         tag.destroy()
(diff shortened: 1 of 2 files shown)


map_gen/maps/crash_site/commands.lua
```diff
@@ -337,6 +337,26 @@ function Public.control(config)
         }
     end)
 
+    local map_chart_tag_clear_callback = Token.register(function(tag)
+        if not tag then
```
grilledham: need to also check that tag isn't invalid.

grilledham (commented):
Chart tags are good. But you only see them if you are looking at the map. Maybe we should add some text at the airstrike location 'player name has called an airstrike at this location'. That way players who aren't looking at the map know it is coming.

Jayefuu:
@grilledham I don't think the message at the location of the strike is necessary. At present it's only a tiny crosshair and then the poison arrives almost instantly. I think it's pretty clear there's been a strike called, and they will soon learn to pop in to map view to see who sent it rather than looking at the chat to find the sender.

I've updated the tag clear callback to include `or not tag.valid.`

grilledham (approved):

PR #1157, opened 2021-01-24
Testing prometheus/grafana

Jayefuu:
We've tested it works on S10 but want it on a live map for a little while so we can gather some data for testing configuring Grafana.

For now it just saves the number of players, will add more later.

### DIFF
--- a/features/prometheus/.gitignore
+++ b/features/prometheus/.gitignore
@@ -1,0 +1,40 @@
+# Compiled Lua sources
+luac.out
+
+# luarocks build files
+*.src.rock
+*.zip
+*.tar.gz
+
+# Object files
+*.o
+*.os
+*.ko
+*.obj
+*.elf
+
+# Precompiled Headers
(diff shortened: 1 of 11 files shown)


grilledham (commented):
I'm not so sure about this, I'll need to think about it. I'll confess I have a strong hatred of using files to transfer data between processes 😄.

To be clear I'm ok with using prometheus/grafana (even though I know nothing about them), my concern is with having the factorio client doing file writes. I wonder if this would be better placed on the web interface. The web interface already knows when players leave/join, so we could hook into that. There might also be the possibility of using sockets rather than files, but I don't know what prometheus/grafana offers. 

My only other thought is that, are there plans to send other stats to prometheus/grafana from factorio? In such a case we may have to extend the server module, which I appreciate wouldn't be as easy as just writing to a file in the short term.

Jayefuu:
Closed while we think about it some more.

PR #1158, opened 2021-01-24
Updated target position for destroyers when cause is artillery

Jayefuu:
Small change to the position which the destroyers are sent. To keep players on their toes because this is the new meta that makes it very easy to use artillery without having to defend it.

image

### DIFF
--- a/map_gen/maps/crash_site/events.lua
+++ b/map_gen/maps/crash_site/events.lua
@@ -347,7 +347,7 @@ local function do_bot_spawn(entity_name, entity, event)
         end
         for i = 1, repeat_cycle do
             if (cause.name == 'artillery-turret') or (cause.name == 'artillery-wagon') then
-                spawn_entity.target = cause.position    -- Overwrite target. Artillery turrets/wagons don't move so send them to entity position. Stops players from picking up the arty and the bots stopping dead.
+                spawn_entity.target = {cause.position.x + math.random(-10,10), cause.position.y + math.random(-10,10)}    -- Overwrite target. Artillery turrets/wagons don't move so send them to entity position. Stops players from picking up the arty and the bots stopping dead.
                 spawn_entity.speed = 0.2
                 -- This is particularly risky for players to do because defender-capsule quantities are not limited by the player force's follower robot count.
                 spawn_entity.name = 'defender-capsule'  -- use 'defender-capsule' (projectile) not 'defender' (entity) since a projectile can target a position but a capsule entity must have another entity as target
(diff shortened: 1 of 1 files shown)


map_gen/maps/crash_site/events.lua
```diff
@@ -347,7 +347,7 @@ local function do_bot_spawn(entity_name, entity, event)
         end
         for i = 1, repeat_cycle do
             if (cause.name == 'artillery-turret') or (cause.name == 'artillery-wagon') then
-                spawn_entity.target = cause.position    -- Overwrite target. Artillery turrets/wagons don't move so send them to entity position. Stops players from picking up the arty and the bots stopping dead.
+                spawn_entity.target = {cause.position.x + math.random(-10,10), cause.position.y + math.random(-10,10)}    -- Overwrite target. Artillery turrets/wagons don't move so send them to entity position. Stops players from picking up the arty and the bots stopping dead.
```
```suggestion
                local cause_position = cause.position
                spawn_entity.target = {cause_position.x + random(-10,10), cause_position.y + random(-10,10)}    -- Overwrite target. Artillery turrets/wagons don't move so send them to entity position. Stops players from picking up the arty and the bots stopping dead.
```

grilledham (commented):
I'm not sure I understand the objective here? Wouldn't this just spread the bots out more making the attack less concentrated?

What happens if the artillery wagon is at the edge of the map, do we send bots into the void?

map_gen/maps/crash_site/events.lua
Jayefuu: Oh yeah. I need to do the random part once before the for loop. I want all the bots to go to the same place, but slightly different to the cause position each time so that they can't just ring them in lasers that never get attacked.

map_gen/maps/crash_site/events.lua
```diff
@@ -345,9 +336,13 @@ local function do_bot_spawn(entity_name, entity, event)
         else
             repeat_cycle = 4
         end
+
+        local random_x = math.random(-10,10) -- calculate before the for loop so all go to the same place
+        local random_y = math.random(-10,10)
```
grilledham, line 341:
```suggestion
        local random_x = random(-10,10) -- calculate before the for loop so all go to the same place
        local random_y = random(-10,10)
```

grilledham (approved):
Looks fine. But I still don't really understand the point. 10 is not much variance, so does it really make much difference? But if you make the variance larger you face the problem that we could send bots far into the void.

Jayefuu:
Yeah I might just ditch this change. It chases the problem not solves it. They'll just make a bigger circle of lasers.

Jayefuu:
Ditching this. I've come up with a better way.

PR #1159, opened 2021-01-24
Added some rendering to make airstrike nicer to use

Jayefuu:
- Shows all players where the airstrike was aimed using the crosshair. Disappears after 5 seconds.
- Shows the user of airstrike the radius their strike will roughly cover Disappears after 20 seconds.

image

### DIFF
--- a/map_gen/maps/crash_site/commands.lua
+++ b/map_gen/maps/crash_site/commands.lua
@@ -358,20 +358,23 @@ function Public.control(config)
         set_timeout_in_ticks(60*30, map_chart_tag_clear_callback, tag) -- To clear the tag after 30 seconds
     end)
 
-    local function render_crosshair(position)
+    local function render_crosshair(data)
         local red = {r = 0.5, g = 0, b = 0, a = 0.5}
         local timeout = 5*60
         local line_width = 10
         local line_length = 2
-        rendering.draw_circle{color=red, radius=1.5, width=line_width, filled=false, target=position, surface="redmew", time_to_live=timeout, forces={"player"}}
-        rendering.draw_line{color=red, width=line_width, from={position.x-line_length, position.y}, to={position.x+line_length, position.y}, surface="redmew", time_to_live=timeout, forces={"player"}}
-        rendering.draw_line{color=red, width=line_width, from={position.x, position.y-line_length}, to={position.x, position.y+line_length}, surface="redmew", time_to_live=timeout, forces={"player"}}
+        local s = data.player.surface
+        local f = data.player.force
+        rendering.draw_circle{color=red, radius=1.5, width=line_width, filled=false, target=data.position, surface=s, time_to_live=timeout, forces={f}}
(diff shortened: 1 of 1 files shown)


map_gen/maps/crash_site/commands.lua
```diff
@@ -358,6 +358,22 @@ function Public.control(config)
         set_timeout_in_ticks(60*30, map_chart_tag_clear_callback, tag) -- To clear the tag after 30 seconds
     end)
 
+    local function render_crosshair(position)
+        local red = {r = 0.5, g = 0, b = 0, a = 0.5}
+        local timeout = 5*60
+        local line_width = 10
+        local line_length = 2
+        rendering.draw_circle{color=red, radius=1.5, width=line_width, filled=false, target=position, surface="redmew", time_to_live=timeout, forces={"player"}}
```
grilledham: It would be better to not hardcode the surface/force. If we pass the player as an argument to the function we can get the surface/force from that.

grilledham (commented):
This is really cool.

I still think we should add some text to say who called the airstrike, so you don't have to switch to the map view to find out.

map_gen/maps/crash_site/commands.lua
Jayefuu: image

Jayefuu: I have now:
- Removed surface and force hard coding
- Added floating text on surface so players don't need to use map view to see strike sender

map_gen/maps/crash_site/commands.lua
```diff
@@ -358,6 +358,25 @@ function Public.control(config)
...
+        local red = {r = 0.5, g = 0, b = 0, a = 0.5}
+        local timeout = 5*60
+        local line_width = 10
+        local line_length = 2
+        local s = data.player.surface
+        local f = data.player.force
+        rendering.draw_circle{color=red, radius=1.5, width=line_width, filled=false, target=data.position, surface=s, time_to_live=timeout, forces={f}}
+        rendering.draw_line{color=red, width=line_width, from={data.position.x-line_length, data.position.y}, to={data.position.x+line_length, data.position.y}, surface=s, time_to_live=timeout, forces={f}}
+        rendering.draw_line{color=red, width=line_width, from={data.position.x, data.position.y-line_length}, to={data.position.x, data.position.y+line_length}, surface=s, time_to_live=timeout, forces={f}}
+        s.create_entity{name="flying-text", position={data.position.x+3, data.position.y}, text = "[item=poison-capsule] "..data.player.name, color = {r = 1, g = 1, b = 1, a = 1}  }
```
grilledham, line 371: I'd probably use the player's color. It'd be consistent with what we do when we print the player's name for the artillery targeting.

grilledham (approved):
Looks good.

map_gen/maps/crash_site/commands.lua
Jayefuu: I like white :) It's clear and easy to read. I think so long as it's consistent for all strike usage it's clean and easy to read.

PR #1161, opened 2021-01-26
Sandworms: Fixed callback crash when no position for worm found

Jayefuu:
- If it couldn't find a non-colliding position it errored when trying to use the worm position to spawn biters
- This would cause the callback to stop working and the roboport wouldn't spawn worms.
- Fixed by adding check for nil worm position

### DIFF
--- a/map_gen/maps/crash_site/features/sandworms.lua
+++ b/map_gen/maps/crash_site/features/sandworms.lua
@@ -51,11 +51,11 @@ local function spawn_sandworms(entity)
                 entity.position.x + math.random(max_worm_spawn_radius * -1, max_worm_spawn_radius),
                 entity.position.y + math.random(max_worm_spawn_radius * -1, max_worm_spawn_radius)
             }
-            worm_position = s.find_non_colliding_position(index, worm_position, 20, 1)
-            if not worm_position then
+            local free_worm_position = s.find_non_colliding_position(index, worm_position, 20, 1)
+            if not free_worm_position then
                 return
             end
-            s.create_entity {name = index, position = worm_position, force = "enemy"}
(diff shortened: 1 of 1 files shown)


map_gen/maps/crash_site/features/sandworms.lua
```diff
@@ -52,9 +52,10 @@ local function spawn_sandworms(entity)
                 entity.position.y + math.random(max_worm_spawn_radius * -1, max_worm_spawn_radius)
             }
             worm_position = s.find_non_colliding_position(index, worm_position, 20, 1)
-            if worm_position then
-                s.create_entity {name = index, position = worm_position, force = "enemy"}
+            if not worm_position then
+                return
             end
+            s.create_entity {name = index, position = worm_position, force = "enemy"}
```
grilledham: I would suggest using a different variable name for the returned worm_position to avoid confusion with the random target location.
```lua
local free_worm_position = s.find_non_colliding_position(index, worm_position, 20, 1)
if not free_worm_position then
	return
end
s.create_entity {name = index, position = free_worm_position, force = "enemy"}
```
That way when spawning multiple worms they are all spawn within a range around the target location rather than from the previous spawned worm. Which currently means the maximum range a worm could theoretically spawn from a roboport is 20 * 4 + max_worm_spawn_radius, which I assume is unintended.

grilledham (changes requested):

map_gen/maps/crash_site/features/sandworms.lua
Jayefuu: You're right, thank you. I've updated this.

grilledham (approved):

PR #1162, opened 2021-01-27
Add Player online time to discord report/jail embeds.

grilledham:


Jayefuu:
The changes are fine but while we're at it can we make some formatting changes to make it easier for anyone reading #moderation-log?

I'll submit the changes soon if you don't have time but I would suggest the new format added some colons and line breaks to make it easier to understand....

**grilledham has reported grilledham**
Server: s10-Refactorio Beta Test
Game time: 0 minutes
Player online time: 0 minutes:
Report reason: blah blah

**grilledham has been jailed by grilledham**
Server: s10-Refactorio Beta Test
Game time: Game time 0 minutes
Player online time : 0 minutes

Jayefuu (approved):

Jayefuu:
@grilledham I made the following changes in the latest commit

- Fixed the bold first line text with your suggestion. 
- Made an attempt at escaping the characters with a function similar to your C#. I think it's working except for the @ case, the unicode isn't being inserted. It errored when I used only one \

image

grilledham:
Thanks, looks like you have done most of the work, I'll finish it up. Looks like lua doesn't support unicode escape sequences.

grilledham:
@Jayefuu I've updated this, the sanitisation should be working now.

If you are curious you can run the tests with the `/test-runner` command in game if you are in debug mode.

PR #1163, opened 2021-01-29
Update destroyer attack behaviour for artillery wagons

Jayefuu:


map_gen/maps/crash_site/events.lua
```diff
@@ -294,6 +294,20 @@ local bot_cause_whitelist = {
...
+local destroyer_callback
+destroyer_callback =
+    Token.register(
+    function(data)
+        local entity = data.entity
+        local destroyer = data.destroyer
+        if not destroyer or not destroyer.valid or not entity or not entity.valid then
+            return
+        end
+        data.entity.surface.create_entity{name = "laser", position=destroyer.position, target=entity, speed=1}
```
grilledham: Maybe we should add a range check before creating the laser. Seems unfair to the player that the bot can shot the wagon before you have an opportunity to shot it back.

```suggestion
        entity.surface.create_entity{name = "laser", position=destroyer.position, target=entity, speed=1}
```

map_gen/maps/crash_site/events.lua
```diff
@@ -294,6 +294,20 @@ local bot_cause_whitelist = {
...
+        set_timeout_in_ticks(30, destroyer_callback, {destroyer = destroyer, entity = entity})
```
```suggestion
        set_timeout_in_ticks(30, destroyer_callback, data)
```

map_gen/maps/crash_site/events.lua
Jayefuu: Done! Thought I'd try be smart and use laser-beam instead which can have a range limit but it looked terrible so I added a range check function instead.

Jayefuu:
I think that's everything @grilledham

map_gen/maps/crash_site/events.lua
```diff
@@ -294,6 +294,28 @@ local bot_cause_whitelist = {
...
+local destroyer_callback
+destroyer_callback =
+    Token.register(
+    function(data)
+        local entity = data.entity
+        local destroyer = data.destroyer
+        if not destroyer or not destroyer.valid or not entity or not entity.valid then
+            return
+        end
+        if euclidean_distance(destroyer.position, entity.position) < 10 then
```
grilledham: `math.sqrt` is an expensive calculation, what we have done in the past is return the distance squared and compared that instead.

So effectively something like
```lua
 ((p.x-q.x)^2 + (p.y-q.y)^2) < 100
```
And add a comment saying something like 100 is a distance of 10 squared.

Jayefuu: Smart! I like it. Done.

grilledham (approved):

PR #1164, opened 2021-01-29
Multi-coordinate support for /spy and /strike.

EigenMan-Factorio:
Added multi-coordinate support to the /spy and /strike commands. Also added help text to the /strike command.

### DIFF
--- a/map_gen/maps/crash_site/commands.lua
+++ b/map_gen/maps/crash_site/commands.lua
@@ -303,7 +303,7 @@ function Public.control(config)
         local coords = {}
         local spy_cost = 100
 
-        for m in string.gmatch(location_string, "%-?%d+") do
+        for m in string.gmatch(location_string, "%-?%d+") do -- Assuming the surface name isn't a valid number.
             table.insert(coords, tonumber(m))
         end
         -- Do some checks on the coordinates passed in the argument
@@ -321,17 +321,17 @@ function Public.control(config)
 
             -- Make sure player has enough coin to cover spying cost
             if coin_count < spy_cost then
-            player.print({'command_description.crash_site_spy_funds'}, Color.fail)
-            return
-        else
-            -- reveal 3x3 chunks centred on chunk containing pinged location
-            -- make sure it lasts 15 seconds
-            for j = 1, 15 do
-                set_timeout_in_ticks(60 * j, chart_area_callback, {player = player, xpos = xpos, ypos = ypos})
+                player.print({'command_description.crash_site_spy_funds'}, Color.fail)
+                return
+            else
+                -- reveal 3x3 chunks centred on chunk containing pinged location
+                -- make sure it lasts 15 seconds
+                for j = 1, 15 do
+                    set_timeout_in_ticks(60 * j, chart_area_callback, {player = player, xpos = xpos, ypos = ypos})
+                end
+                game.print({'command_description.crash_site_spy_success', player_name, spy_cost, xpos, ypos}, Color.success)
+                inv.remove({name = "coin", count = spy_cost})
             end
-            game.print({'command_description.crash_site_spy_success', player_name, spy_cost, xpos, ypos}, Color.success)
-            inv.remove({name = "coin", count = spy_cost})
-        end
 
             -- move to the next set of coordinates
             i = i+2
@@ -407,7 +407,7 @@ function Public.control(config)
         local strikeCost = count * 4 -- the number of poison-capsules required in the chest as payment
 
         -- parse GPS coordinates from map ping
-        for m in string.gmatch(location_string, "%-?%d+") do
+        for m in string.gmatch(location_string, "%-?%d+") do -- Assuming the surface name isn't a valid number.
             table.insert(coords, tonumber(m))
         end
 
@@ -417,11 +417,6 @@ function Public.control(config)
             return
         end
 
-        -- process each set of coordinates with a 10 strike limit
-        local i = 1
-        local xpos = coords[i]
-        local ypos = coords[i+1]
-        while xpos ~= nil and ypos ~= nil and i < 20 do
         -- Check that the chest is where it should be.
         local entities = s.find_entities_filtered {position = {-0.5, -3.5}, type = 'container', limit = 1}
         local dropbox = entities[1]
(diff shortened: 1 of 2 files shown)


map_gen/maps/crash_site/commands.lua
```diff
@@ -403,9 +416,12 @@ function Public.control(config)
             player.print({'command_description.crash_site_airstrike_invalid'}, Color.fail)
             return
         end
-        local xpos = coords[1]
-        local ypos = coords[2]
```
Jayefuu, line 419: Would it be sensible, for performance reasons, to limit the number of strikes usable per line to 10? For performance reasons and to stop one person hogging all the capsules at once?

Jayefuu (changes requested):
These are great changes!

I don't have any problem with the code except for a few small comments:
1) As it is but you need to add a check for if one strike location is invalid (off map, no biters present (eg spawn) then the other locations won't work.

For example:
`/strike [gps=0,0,redmew] [gps-100,100,redmew] [gps=200,200,redmew] `

In this case the first one is refused because it could be griefing but it then doesn't continue to do the second and third positions which are valid

2) I think you should limit the number of strikes in a single command to 10. This stops one person hogging all of the capsules quite so much. And I think there should be a limit for performance reasons otherwise someone could use the entire chest of capsules at once rather than spread over a few minutes.

map_gen/maps/crash_site/commands.lua
EigenMan-Factorio: Yes, I think limiting to 10 strikes per command would be good. I was actually debating whether or not I should limit to 10 with my initial changes so this is good feedback. I'll add that change.

EigenMan-Factorio:
1) I'm fine with bailing out after the first failed strike. If I were to strike 10 invalid locations then I would rather see 1 failure message instead of 10. And if they start going crazy and striking places where there aren't biters, I think it would be good to abort the rest of the strikes so they can figure out what it is they're trying to accomplish with the strikes.

I don't have strong feeling about this though, so if someone does have strong feeling about it, or if grilledham has an opinion, I'd love to hear it.

2) Along the lines of limiting to 10 coordinates, should that be applied to /spy also? I'm not as worried about someone hogging the shared resources in this case since the coins come out of the players inventory. But there is a message for each set of coordinates so that could become overwhelming.

Jayefuu:
>     1. Along the lines of limiting to 10 coordinates, should that be applied to /spy also? I'm not as worried about someone hogging the shared resources in this case since the coins come out of the players inventory. But there is a message for each set of coordinates so that could become overwhelming.

I agree. If someone wants to waste 1000 coins on 10 spy, that's on them :P I doubt spy will be used more than 1 at a time because they won't have time to use it and tag the map.

map_gen/maps/crash_site/commands.lua
```diff
@@ -307,16 +306,24 @@ function Public.control(config)
...
+        local xpos = coords[i]
+        local ypos = coords[i+1]
+        while xpos ~= nil and ypos ~= nil do
+            local coin_count = inv.get_item_count("coin")
+
+            -- Make sure player has enough coin to cover spying cost
+            if coin_count < spy_cost then
             player.print({'command_description.crash_site_spy_funds'}, Color.fail)
             return
         else
```
grilledham: The formatting is a bit off here, the `else` should have the same indentation as the `if`

map_gen/maps/crash_site/commands.lua
```diff
@@ -307,16 +306,24 @@ function Public.control(config)
         for m in string.gmatch(location_string, "%-?%d+") do
```
grilledham: It wasn't an issue before, but it might be now if we have surface names that are valid numbers. Appreciate that's unlikely to be a problem, more just awareness if this code was copied for other scenarios that make their own surfaces. Maybe a comment would be alright for now.
```suggestion
        for m in string.gmatch(location_string, "%-?%d+") do -- Assuming the surface name isn't a valid number.
```

Same again for when we process the coords in the strike function.

map_gen/maps/crash_site/commands.lua
```diff
@@ -403,9 +416,12 @@ function Public.control(config)
...
             return
         end
-        local xpos = coords[1]
-        local ypos = coords[2]
 
+        -- process each set of coordinates with a 10 strike limit
+        local i = 1
+        local xpos = coords[i]
+        local ypos = coords[i+1]
+        while xpos ~= nil and ypos ~= nil and i < 20 do
```
grilledham: Formatting: the content of the while loop needs indenting.

grilledham (changes requested):
Good stuff, a few minor comments, mostly formatting.

This code that is now inside the while loop in the strike function:
```lua
if dropbox == nil then
    player.print("Chest not found. Replace it here: [gps=-0.5,-3.5,redmew]")
    return
end
```
Should probably be a one time check before the while loop as it is invariant of the loop.

grilledham:
> 1. I'm fine with bailing out after the first failed strike. If I were to strike 10 invalid locations then I would rather see 1 failure message instead of 10. And if they start going crazy and striking places where there aren't biters, I think it would be good to abort the rest of the strikes so they can figure out what it is they're trying to accomplish with the strikes.
> 
> I don't have strong feeling about this though, so if someone does have strong feeling about it, or if grilledham has an opinion, I'd love to hear it.

I'm ok with either bailing on first failure, or processing only the valid ones.

map_gen/maps/crash_site/commands.lua
EigenMan-Factorio: I agree.

map_gen/maps/crash_site/commands.lua
EigenMan-Factorio: I agree.

EigenMan-Factorio:
> This code that is now inside the while loop in the strike function:
> 
> ```lua
> if dropbox == nil then
>     player.print("Chest not found. Replace it here: [gps=-0.5,-3.5,redmew]")
>     return
> end
> ```
> 
> Should probably be a one time check before the while loop as it is invariant of the loop.

I intentionally put this inside the loop because I was concerned that the the chest could *potentially* be removed while processing the multiple strike commands and could possibly give the player free strikes. I think the check with capCount should prevent that but I'm not familiar enough with the code to be sure. The timing would have to be just right to pull it off and the time window is now shorter with a 10 strike limit.

But I'm also seeing that the chest isn't removable so I'm fine with moving this to the outside of the loop. I'm curious if you're aware of a case where this test would fail with the scenario in it's current state?

EigenMan-Factorio:
Hey guys, thanks for your feedback and your patience while I figure out git and github and Lua and everything else. I think I've addressed all of the issues at this point so if there's something I'm missing, go ahead and point it out.

I don't know if I'm supposed to Re-request review or anything so... I'll just leave this as is for a while. I probably should have requested the initial review when I created the pull request so thank you grilledham for doing that for me. If I should have done anything else different process-wise, please let me know. I don't have plans to contribute anything else for now, but who knows.

grilledham (approved):
Looks good to me. I'll wait for @Jayefuu to approve then either Jay or I will merge.

grilledham:
> I intentionally put this inside the loop because I was concerned that the the chest could _potentially_ be removed while processing the multiple strike commands and could possibly give the player free strikes.

No player actions are performed when running the lua script code. The only way the chest would be removed is if the lua code removed it either directly or indirectly (through calling another function, or raising an event) but that is not the case here.

> Hey guys, thanks for your feedback and your patience while I figure out git and github and Lua and everything else. I think I've addressed all of the issues at this point so if there's something I'm missing, go ahead and point it out.

No thank you for taking the time to contribute this, it is very much appreciated 👍.

If this is your first time using git/github then I think you did a good job. The only thing I would say is usually you name the branch after the feature/bug you are working on, so if other people check out your branch (like I did to play test this) it's easier to know what the branch is for.

>  don't know if I'm supposed to Re-request review or anything so... I'll just leave this as is for a while. I probably should have requested the initial review when I created the pull request

We don't have a formal process, so I don't think you did anything wrong.

PR #1165, opened 2021-01-31
Remove dynamic requires from diggy.

grilledham:
I'm rewriting the save deflating functionality and I don't want to support dynamic requires e.g. this
```lua
require ('map_gen.maps.diggy.feature.' .. feature_name)
```
which in theory can only be resolved by actually running the lua code. 
I think this is the only place in the whole code base that does this, so I rewrote it to not use dynamic requires.


linaori (approved):

linaori:
Yeah this change makes a lot of sense tbh

PR #1167, opened 2021-02-02
Added configurable pollution to outposts for future testing

Jayefuu:
My plan is to make a harder crash site version where getting free resources from outposts creates pollution.

My intention is to balance it so it's still better to get items from outposts, but this commit is intended to give me a way to take saved games and test different amounts of pollution to check for balance. I don't intend to test this on live for a while but want to test it offline, this will give me the ability to turn it on half way through a game.

Need to consider how it doesn't take into account what the item is so it will be better to get free end-game items and probably limit how much iron we let into base later in the game. Might also create interesting meta where we are careful about what resources we take from outposts depending upon the pollution making setting.


map_gen/maps/crash_site/outpost_builder.lua
```diff
@@ -1247,7 +1278,11 @@ local function do_magic_crafters()
             local fcount = floor(count)
 
             if fcount > 0 then
-                entity.get_output_inventory().insert {name = data.item, count = fcount}
+                if entity.get_output_inventory().can_insert(data.item) then -- No pollution once full. Taking items out of crafters makes pollution
```
grilledham: make a local for inventory, so we don't have to call `get_output_inventory()` twice.

map_gen/maps/crash_site/outpost_builder.lua
```diff
@@ -1213,6 +1219,31 @@ local function do_artillery_turrets_targets()
...
+    game.print(args.multiplier)
+
+    local multipliers = {}
+    for m in string.gmatch(args.multiplier, "%-?%d+") do -- Assuming the surface name isn't a valid number.
+        table.insert(multipliers, tonumber(m))
+    end
+    if not multipliers or #multipliers ~= 1 then
+        game.player.print("Fail")
+        return
+    end
```
grilledham: Are you planning on having multiple multipliers?
Can this just be
```suggestion
    local multiplier = tonumber(args.multiplier)
    if not multiplier then
        player.print("Fail")
        return
    end   
```

map_gen/maps/crash_site/outpost_builder.lua
```diff
@@ -1213,6 +1219,31 @@ local function do_artillery_turrets_targets()
     end
 end
 
+local function set_pollution_multiplier(args, player)
+    game.print(args.multiplier)
```
grilledham: There are some places in this function where you use `game.print` they should probably be `player.print`.

map_gen/maps/crash_site/outpost_builder.lua
```diff
@@ -1213,6 +1219,31 @@ local function do_artillery_turrets_targets()
...
+    for m in string.gmatch(args.multiplier, "%-?%d+") do -- Assuming the surface name isn't a valid number.
+        table.insert(multipliers, tonumber(m))
+    end
+    if not multipliers or #multipliers ~= 1 then
+        game.player.print("Fail")
+        return
+    end
+
+    local old_multiplier = pollution_multiplier
+    pollution_multiplier = multipliers[1]
```
grilledham: Setting `pollution_multiplier` like this isn't desync safe. When you set `pollution_multiplier` you are not changing the value stored in global. What you need to do is box the value in a table, i.e. `local pollution_multiplier = {0}`. Then read/write the value through the table.

map_gen/maps/crash_site/outpost_builder.lua
```diff
@@ -1213,6 +1219,31 @@ local function do_artillery_turrets_targets()
...
+    for _, p in pairs(game.players) do
+        if p.admin then
+            p.print(player.name..' changed magic crafter pollution multiplier from '..old_multiplier..' to '..pollution_multiplier)
```
grilledham: We know there is always going to be one admin, the admin who ran the command, so we could do the string concatenation once outside of the loop. This is just me being picky 😄.

map_gen/maps/crash_site/outpost_builder.lua
```diff
@@ -1891,4 +1926,27 @@ Event.add(Retailer.events.on_market_purchase, do_outpost_upgrade)
...
+    set_pollution_multiplier
+)
+
+Command.add(
+    'get_pollution_multiplier',
+    {
+        description = {'command_description.get_pollution_multiplier'},
+        required_rank = Ranks.admin,
+        capture_excess_arguments = true,
+        allowed_by_server = false
```
grilledham: Not allowed by server is odd, but I see that would make the `get_pollution_multiplier` function slightly more complicated.

grilledham (changes requested):
Long term it would be good to move the commands out of the outpost_builders.lua, but as this is temporary I'll allow it for now.
The only thing that is really blocking this is that it's not desync safe. Let me know if you need any help with that.

grilledham:
Thinking about the future, I wonder if we want configurable pollution values per outpost type, or per recipe.
Also maybe the magic crafter should spawn with a full inventory, so there is no initial pollution.

Jayefuu:
> 
> 
> Thinking about the future, I wonder if we want configurable pollution values per outpost type, or per recipe.

I think that would be great. And much better for balancing. I'm not sure how to do it though.

> Also maybe the magic crafter should spawn with a full inventory, so there is no initial pollution.

That's so obvious. I should have caught that, I shall make it so.

map_gen/maps/crash_site/outpost_builder.lua
Jayefuu: I was being lazy. I'll look up how to resolve if it's from the server from another example somewhere.

map_gen/maps/crash_site/outpost_builder.lua
Jayefuu: Nah, I appreciate it, need to get into better habits and this helps with that.

map_gen/maps/crash_site/outpost_builder.lua
Jayefuu: I'm not sure I understand. In other places you directly write to the global vars. Such as outpost_count. I've had a go but I don't think I've done this right in the next commit.

Jayefuu:
@grilledham that was quite a messy last commit I made.
- I still haven't fixed the global desync issue I suspect and will need some help
- In taking a look at the magic crafters I've made them all start with full inventory so they shouldn't release pollution.
- I'm not sure how, but oil refineries now flash that they don't have power and I can't work out why

map_gen/maps/crash_site/outpost_builder.lua
grilledham: `outpost_count` isn't a global var.

map_gen/maps/crash_site/outpost_builder.lua
```diff
@@ -1213,6 +1219,29 @@ local function do_artillery_turrets_targets()
...
+local function set_pollution_multiplier(args, player)
+    local multiplier = tonumber(args.multiplier)
+    if not multiplier then
+        player.print("Fail")
+        return
+    end
+
+    local old_multiplier = pollution_multiplier.value
+    pollution_multiplier.value = multiplier
+    local message = {player.name, ' changed magic crafter pollution multiplier from ', old_multiplier, ' to ', pollution_multiplier}
```
grilledham: Are you missing a `table.concat`? You can't print a table unless it is a valid LocalisedString.

map_gen/maps/crash_site/outpost_builder.lua
```diff
@@ -1445,14 +1479,21 @@ Public.magic_item_crafting_callback =
...
         local recipe = callback_data.recipe
-        if recipe then
+        local has_fluid_output = callback_data.keep_active
+        if recipe and not has_fluid_output then -- to avoid trying to put an item into a fluid inventory
             entity.set_recipe(recipe)
-        else
-            local furance_item = callback_data.furance_item
-            if furance_item then
+            local output_inv = entity.get_output_inventory()
+            output_inv.insert {name = callback_data.output.item, count = 200}
```
grilledham: If you just want a full stack you can just pass in a string.
```suggestion
            output_inv.insert(callback_data.output.item)
```

map_gen/maps/crash_site/outpost_builder.lua
```diff
@@ -1445,14 +1479,21 @@ Public.magic_item_crafting_callback =
         entity.operable = false
 
         local recipe = callback_data.recipe
-        if recipe then
+        local has_fluid_output = callback_data.keep_active
```
grilledham: Maybe we should rename `keep_active` to `has_fluid_output`

map_gen/maps/crash_site/outpost_builder.lua
```diff
@@ -1445,14 +1479,21 @@ Public.magic_item_crafting_callback =
...
+        if recipe and not has_fluid_output then -- to avoid trying to put an item into a fluid inventory
```
grilledham: I think this whole if, elseif, else block can be simplified to:
```lua
local recipe = callback_data.recipe
if recipe then
	entity.set_recipe(recipe)
	if not callback_data.keep_active then -- to avoid trying to put an item into a fluid inventory
		entity.get_output_inventory().insert(callback_data.output.item)
	end
else
	local furnace_item = callback_data.furnace_item
	if furnace_item then
		local inv = entity.get_inventory(2) -- defines.inventory.furnace_source
		inv.insert(furnace_item)                
		entity.get_output_inventory().insert(callback_data.output.item)
	end
end
```

map_gen/maps/crash_site/outpost_builder.lua
```diff
@@ -1213,6 +1219,29 @@ local function do_artillery_turrets_targets()
...
+        if p.admin then
+            p.print(message)
+        end
+    end
+end
+
+local server_player = {name = '<server>', print = print}
+local function get_pollution_multiplier(player)
+    player = player or server_player
+    player.print('Current pollution multiplier is: '..pollution_multiplier)
```
grilledham:
```suggestion
    player.print('Current pollution multiplier is: ' .. pollution_multiplier.value)
```

grilledham (commented):
Haven't had time to play test this yet.

Thanks for renaming all the furnace typos 😄.

Not sure why oil refineries would now start flashing missing power, if you don't figure it out, I'll investigate later.

I think pollution_multiplier should be desync safe now.

Jayefuu:
@grilledham I made the suggested changes but now I get an error in my logs and this outpost fails to build properly. I didn't manage to fix it, any ideas?

 `131.012 Script @C:/Users/James/AppData/Roaming/Factorio/temp/currently-playing/utils/task.lua:77: 'stack': table expected, got nil`

image

map_gen/maps/crash_site/outpost_builder.lua
```diff
@@ -1447,11 +1481,15 @@ Public.magic_item_crafting_callback =
         local recipe = callback_data.recipe
         if recipe then
             entity.set_recipe(recipe)
+            if not callback_data.keep_active then -- to avoid trying to put an item into a fluid inventory
```
grilledham: You missed this when changing the table field name.
```suggestion
            if not callback_data.has_fluid_output then -- to avoid trying to put an item into a fluid inventory
```

grilledham:
> I'm not sure how, but oil refineries now flash that they don't have power and I can't work out why

They are flashing in the develop branch without your changes. I can't remember if they always flashed, but it doesn't seem related to your changes.

map_gen/maps/crash_site/outpost_builder.lua
```diff
@@ -1891,4 +1933,27 @@ Event.add(Retailer.events.on_market_purchase, do_outpost_upgrade)
 
 Event.add(defines.events.on_selected_entity_changed, market_selected)
 
+Command.add(
+    'set_pollution_multiplier',
```
grilledham: Minor but we normally use dash '-' rather than underscore '_' for commands.

map_gen/maps/crash_site/outpost_builder.lua
```diff
@@ -1213,6 +1219,29 @@ local function do_artillery_turrets_targets()
...
+    local message = player.name..' changed magic crafter pollution multiplier from '..old_multiplier..' to '..pollution_multiplier
+    for _, p in pairs(game.players) do
+        if p.admin then
+            p.print(message)
+        end
+    end
+end
+
+local server_player = {name = '<server>', print = print}
+local function get_pollution_multiplier(player)
```
grilledham: This doesn't work. I think it should be.
```suggestion
local function get_pollution_multiplier(args, player)
```

map_gen/maps/crash_site/outpost_builder.lua
```diff
@@ -1213,6 +1219,29 @@ local function do_artillery_turrets_targets()
...
+local function set_pollution_multiplier(args, player)
+    local multiplier = tonumber(args.multiplier)
+    if not multiplier then
+        player.print("Fail")
+        return
+    end
+
+    local old_multiplier = pollution_multiplier.value
+    pollution_multiplier.value = multiplier
+    local message = player.name..' changed magic crafter pollution multiplier from '..old_multiplier..' to '..pollution_multiplier
```
grilledham:
```suggestion
    local message = player.name..' changed magic crafter pollution multiplier from '..old_multiplier..' to '..pollution_multiplier.value
```

map_gen/maps/crash_site/outpost_builder.lua
```diff
@@ -1447,11 +1481,15 @@ Public.magic_item_crafting_callback =
         local recipe = callback_data.recipe
         if recipe then
             entity.set_recipe(recipe)
+            if not callback_data.has_fluid_output then -- to avoid trying to put an item into a fluid inventory
+                entity.get_output_inventory().insert(callback_data.output.item)
```
grilledham, line 1485: Thinking about it don't we also need to handle the fluid output case by inserting the fluid items.

map_gen/maps/crash_site/outpost_builder.lua
Jayefuu: Updated in next commit

map_gen/maps/crash_site/outpost_builder.lua
Jayefuu: Updated in next commit

map_gen/maps/crash_site/outpost_builder.lua
Jayefuu: Updated in next commit

map_gen/maps/crash_site/outpost_builder.lua
Jayefuu: I decided not to do this because a) I'm lazy and b) I thought it would be fun to see if they worked out that fluids don't produce pollution. I can't remember the last time I saw someone connect up oil or petroleum on crash site. I was hoping it might bring back flamethrower meta if using oil produced less pollution than getting ammo from magic crafters.

Mostly a) but now I've defended it I quite like b) :D

Jayefuu:
I think that's everything @grilledham

map_gen/maps/crash_site/outpost_builder.lua
grilledham: I'm ok with the laziness 😄.

grilledham (approved):

PR #1169, opened 2021-02-06
Add combat robots to corpse remover.

grilledham:


features/biter_corpse_remover.lua
```diff
@@ -41,4 +37,39 @@ local function entity_died(event)
...
+        return
+    end
+
+    local corpses = surface.find_entities_filtered{position = position, radius = 5, name = corpse_name}
+    process_corpses(corpses)
+end)
+
+local function entity_died(event)
+    local prototype_type = event.prototype.type
+    if prototype_type == 'combat-robot' then
```
Jayefuu, line 65: Should this be `if prototype_type == 'combat-robot' or prototype_type == 'projectile' then`?

Not all robots spawned are combat-robots. For the artillery turrets I use entities of type projectile because they can have their target set as a position not an entity.

Jayefuu (changes requested):
I think this code won't work for all instances and that it won't clear up robot corpses spawned when an artillery turret is used. Though I could be wrong.

features/biter_corpse_remover.lua
grilledham: The capsules turn into combat robots when they hatch, so this works. I have tested the artillery turrets case.

Jayefuu: Excellent.

Jayefuu (approved):

PR #1170, opened 2021-02-08
Player kills deposit coins in inventory

Jayefuu:
The amount of coins all over the map has been frustrating some people. With there being more stuff to spend coins on I thought I'd make this change to reduce frustration. Kills by artillery, turrets and airstrike will still spill coins but this way a nice flying text pops up and the coins go straight into the inventory of the player that caused the death if they have space in their inventory.

image

### DIFF
--- a/map_gen/maps/crash_site/events.lua
+++ b/map_gen/maps/crash_site/events.lua
@@ -439,9 +439,10 @@ local function do_coin_drop(entity_name, entity, cause)
 
     local count = random(bounds.low, bounds.high)
     if count > 0 then
-        if cause.name == 'character' and cause.player.can_insert({name="coin", count = count}) then -- try put coins into inventory
-            cause.player.insert{name="coin", count = count}
-            entity.surface.create_entity{name="flying-text", position = {entity.position.x-1, entity.position.y}, text="+"..count.." [img=item.coin]", color={1, 0.8, 0, 0.5}, render_player_index=cause.player.index}
+        local coins = {name="coin", count = count}
+        if cause and cause.player and cause.name == 'character'  and cause.player.can_insert(coins) then
+            cause.player.insert(coins)
+            entity.surface.create_entity{name="flying-text", position = {position.x-1, position.y}, text="+"..count.." [img=item.coin]", color={1, 0.8, 0, 0.5}, render_player_index=cause.player.index}
         else -- spill them on the floor
             set_timeout_in_ticks(
                 1,
--- a/map_gen/maps/crash_site/events.lua
+++ b/map_gen/maps/crash_site/events.lua
@@ -438,23 +438,31 @@ local function do_coin_drop(entity_name, entity, cause)
     end
 
     local count = random(bounds.low, bounds.high)
-    if count > 0 then
-        local coins = {name="coin", count = count}
-        if cause and cause.player and cause.name == 'character'  and cause.player.can_insert(coins) then
-            cause.player.insert(coins)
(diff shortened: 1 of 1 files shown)


Jayefuu:
image

map_gen/maps/crash_site/events.lua
```diff
@@ -439,15 +439,20 @@ local function do_coin_drop(entity_name, entity)
...
-            spill_items,
-            {
-                count = count,
-                surface = entity.surface,
-                position = position
-            }
-        )
+        if cause.name == 'character' and cause.player.can_insert({name="coin", count = count}) then -- try put coins into inventory
+            cause.player.insert{name="coin", count = count}
+            entity.surface.create_entity{name="flying-text", position = {entity.position.x-1, entity.position.y}, text="+"..count.." [img=item.coin]", color={1, 0.8, 0, 0.5}, render_player_index=cause.player.index}
```
```suggestion
            entity.surface.create_entity{name="flying-text", position = {position.x-1, position.y}, text="+"..count.." [img=item.coin]", color={1, 0.8, 0, 0.5}, render_player_index=cause.player.index}
```

map_gen/maps/crash_site/events.lua
```diff
@@ -439,15 +439,20 @@ local function do_coin_drop(entity_name, entity)
...
-        set_timeout_in_ticks(
-            1,
-            spill_items,
-            {
-                count = count,
-                surface = entity.surface,
-                position = position
-            }
-        )
+        if cause.name == 'character' and cause.player.can_insert({name="coin", count = count}) then -- try put coins into inventory
```
grilledham: It's possible, albeit rare, that cause.player could be `nil`, even when the entity is a character so we should probably check.

grilledham: We could make a local for `{name="coin", count = count}` that way we don't have to make the table twice.

grilledham: `cause` could be `nil` so we need to check that.

grilledham (changes requested):
Cool feature I'm sure it will be popular. It's not as visible obvious to other players that coins are being earned for killing worms/nests, but I think this is the right way to do it.

Some extra checks required before this can be merged.

map_gen/maps/crash_site/events.lua
```diff
@@ -439,15 +439,21 @@ local function do_coin_drop(entity_name, entity)
...
-            1,
-            spill_items,
-            {
-                count = count,
-                surface = entity.surface,
-                position = position
-            }
-        )
+        local coins = {name="coin", count = count}
+        if cause and cause.player and cause.name == 'character'  and cause.player.can_insert(coins) then
```
grilledham: The `cause.name == 'character'` check needs to happen before the `cause.player` check. Else when a non character entity is the cause `cause.player` errors.

grilledham (commented):
Maybe I'm obsessive but if you write the code like this you can avoid the repeated table lookup.
```lua
local count = random(bounds.low, bounds.high)
if count <= 0 then
    return
end

if cause and cause.name == 'character' then
    local player = cause.player
    if player and player.valid then
        local coins = {name = "coin", count = count}
        if player.can_insert(coins) then
            entity.surface.create_entity{name="flying-text", position = {position.x - 1, position.y}, text = "+" .. count .. " [img=item.coin]", color = {1, 0.8, 0, 0.5}, render_player_index = player.index}
            return
        end
    end
end

-- spill them on the floor
set_timeout_in_ticks(
    1,
    spill_items,
    {
        count = count,
        surface = entity.surface,
        position = position
    }
)
```
I'll leave it up to you.

Jayefuu:
No, that way makes sense, it's cleaner. I've updated with your advice and retested.

grilledham (approved):

PR #1171, opened 2021-02-09
Added fish consumed to player_stats and crash site awards

Jayefuu:


Jayefuu:
Uhhhh @grilledham I made this dependant on a different branch, did I do that wrong?

features/player_stats.lua
```diff
@@ -263,6 +279,15 @@ local function rocket_launched(event)
     change_for_global(satellites_launched_name, 1)
 end
 
+local function capsule_used(event)
+    local item = event.item
+
+    if not item or not item.valid or not item.name == 'raw-fish' then
```
grilledham: `not item.name == 'raw-fish'` is a bit of a lua trap, I'm guessing you think it's `not (item.name == 'raw-fish')` but it's actually `(not item.name) == 'raw-fish'` which is always `false` as `(not item.name)` is always `false`. Luckily there is an easy solution.
```suggestion
    if not item or not item.valid or item.name ~= 'raw-fish' then
```

grilledham:
> Uhhhh @grilledham I made this dependant on a different branch, did I do that wrong?

It's always a bit tricky to have two PR that touch the same code. I don't think you did anything wrong, but it'll probably be simpler if we wait for the other PR to finish first.

grilledham (commented):
I think this is good. Need to fix the merge conflict then I'll approve, let me know if you need help with that.

Jayefuu:
I think that's right @grilledham

features/player_stats.lua
```diff
@@ -186,7 +188,7 @@ local function robot_built_entity(event)
     end
 
     -- When item gets built, add to the total entities built for the player whose robot built the entity NOT for the player who placed the ghost
-    local robot_owner = robot.logistic_network.cells[1].owner
+    local robot_owner = event.robot.logistic_network.cells[1].owner
```
grilledham: Probably something that changed when resolving the merge conflict.
```suggestion
    local robot_owner = robot.logistic_network.cells[1].owner
```

grilledham (approved):
Looks good. One line seems to have changed, probably an issue with resolving the merge conflict, after fixing that you can merge.

PR #1178, opened 2021-02-18
Made map selection examples clearer for scenarios like danger ores

Jayefuu:
Barthoze pointed out that it's not clear how to select a danger ores preset and that simply selecting danger_ores doesn't work.

I've updated the documentation to make it clearer.


barthoze (approved):
It is much clearer that way.

grilledham (approved):

PR #1179, opened 2021-02-18
Fixed player coin count bug from picking up chests

Jayefuu:
Problem: Picking up a chest with coins in it added to the player's coin score. Players could place chests with coins and mine them to increase their score.

Solution: Change the event that adds the coin to the player score so that we can check what entity was mined and apply the coins to the count conditionally.

Details:
- Removed events.on_player_mined_item from outpost_builder because we don't want to add coins every time they mine and it's duplicated in player_stats.
- Removed events.on_player_mined_item from player_stats because it doesn't tell us what entity the items came from.
- Added coin tracking to events.on_pre_player_mined_item event instead. Checks if the mined item was a chest with coins in it and adds to the player stats if so.
- Tested that coins from trees and rocks still track when market is enabled
- While I was at it, added a pretty coin floating text for when RNG gives coins from mining trees/rocks

### DIFF
--- a/features/market.lua
+++ b/features/market.lua
@@ -10,6 +10,7 @@ local RS = require 'map_gen.shared.redmew_surface'
 local market_items = require 'resources.market_items'
 local fish_market_bonus_message = require 'resources.fish_messages'
 local ScoreTracker = require 'utils.score_tracker'
+local Color = require 'resources.color_presets'
 local change_for_player = ScoreTracker.change_for_player
 local get_for_player = ScoreTracker.get_for_player
 local coins_earned_name = 'coins-earned'
@@ -123,7 +124,7 @@ local function fish_earned(event, amount)
 
     local stack = {name = currency, count = amount}
     local inserted = player.insert(stack)
-    player.surface.create_entity{name="flying-text", position = {player.position.x - 1, player.position.y}, text = "+" .. amount .. " [img=item.coin]", color = {1, 0.8, 0, 0.5}, render_player_index = player.index}
+    player.surface.create_entity{name="flying-text", position = {player.position.x - 1, player.position.y}, text = "+" .. amount .. " [img=item.coin]", color = Color.gold, render_player_index = player.index}
 
     local diff = amount - inserted
     if diff > 0 then
--- a/features/market.lua
+++ b/features/market.lua
@@ -123,6 +123,7 @@ local function fish_earned(event, amount)
 
     local stack = {name = currency, count = amount}
     local inserted = player.insert(stack)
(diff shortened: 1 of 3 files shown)


features/player_stats.lua
```diff
@@ -145,14 +145,21 @@ local function picked_up_item(event)
...
-local function player_mined_item(event)
+local function pre_player_mined_item(event)
     if event.entity.type == 'simple-entity' then -- Cheap check for rock, may have other side effects
         change_for_global(rocks_smashed_name, 1)
         return
     end
     if event.entity.type == 'tree' then
         change_for_global(trees_cut_down_name, 1)
     end
+    if event.entity.type == 'container'  and event.entity.operable == false then-- We have to do it here because we can't check the entity from player_mined_item
```
grilledham: The comment doesn't seem that useful, people would probably figure out you are using this event because of the properties available on the event.
However, a comment explaining why there is an operable check would be useful. Something like 'We check the container is not operable as these are used as loot chest and players can't add coins to them.'

features/player_stats.lua
```diff
@@ -145,14 +145,21 @@ local function picked_up_item(event)
...
 end
 
-local function player_mined_item(event)
+local function pre_player_mined_item(event)
     if event.entity.type == 'simple-entity' then -- Cheap check for rock, may have other side effects
         change_for_global(rocks_smashed_name, 1)
         return
     end
     if event.entity.type == 'tree' then
         change_for_global(trees_cut_down_name, 1)
```
grilledham, line 155: Add a return after this line. No need to check type == 'container' after we know it is 'tree'.

features/player_stats.lua
```diff
@@ -145,14 +145,21 @@ local function picked_up_item(event)
     end
 end
 
-local function player_mined_item(event)
+local function pre_player_mined_item(event)
```
grilledham, line 148: For this function we should probably make a local variable for entity as it is used a fair amount.

features/market.lua
```diff
@@ -123,6 +123,7 @@ local function fish_earned(event, amount)
 
     local stack = {name = currency, count = amount}
     local inserted = player.insert(stack)
+    player.surface.create_entity{name="flying-text", position = {player.position.x - 1, player.position.y}, text = "+" .. amount .. " [img=item.coin]", color = {1, 0.8, 0, 0.5}, render_player_index = player.index}
```
grilledham: What color is that? Is it one of the colors in <https://github.com/Refactorio/RedMew/blob/develop/resources/color_presets.lua>

grilledham (approved):
Looks good 👍. Only some minor stuff.

features/market.lua
Jayefuu: I just picked one out of a colour chart that looks nice. But I can just use Color.gold, that's almost as good and much clearer.

PR #1180, opened 2021-02-18
Updated artillery behaviour to get rid of destroyer-capsule

Jayefuu:
Use of the destroyer-capsule projectile instead of destroyer capsule meant that the destroyers and defenders didn't attack player entities until they reached the destination. They also couldn't be shot down by a well defended wall, they'd fly over.

I changed to destroyer-capsule so that they couldn't pick the artillery up after the bots spawned as this would remove their target and cause them to float around.

By request I have reveret to the destroyers that can be shot down as they're more fun and at current balance the destroyers kill the artillery every time, which is not much fun when they cannot be defended against.

To mitigate against players picking up the artillery after the destroyer/defender spawns, artillery targets will become unminable for 5 seconds after the destroyer/defender deploys.

This leaves the bug where they can pick up the artillery before the last shell destroys the turret. I considered adding artillery to a table when placed and checking their ammo inventory once every few seconds. I was going to set all artillery turrets to unminable unless they have no ammo for 3 seconds, but this seems like a lot of work for a small bug.

### DIFF
--- a/map_gen/maps/crash_site/events.lua
+++ b/map_gen/maps/crash_site/events.lua
@@ -322,6 +322,10 @@ local artillery_cooldown_callback
 artillery_cooldown_callback =
     Token.register(
     function(entity)
+        if not entity.valid then
+            return
+        end
+
         entity.minable = true
     end
 )
--- a/map_gen/maps/crash_site/events.lua
+++ b/map_gen/maps/crash_site/events.lua
@@ -318,6 +318,15 @@ destroyer_callback =
     end
 )
 
+local artillery_cooldown_callback
+artillery_cooldown_callback =
+    Token.register(
+    function(data)
+        local entity = data.entity
+        entity.minable = true
+    end
+)
+
 local function do_bot_spawn(entity_name, entity, event)
     -- Return if the entity killed is not on the white list
     if not bot_spawn_whitelist[entity_name] then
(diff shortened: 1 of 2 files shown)


map_gen/maps/crash_site/events.lua
```diff
@@ -318,6 +318,15 @@ destroyer_callback =
     end
 )
 
+local artillery_cooldown_callback
+artillery_cooldown_callback =
+    Token.register(
+    function(data)
```
grilledham: `data` could just be the entity itself, don't have to put it in a table.

grilledham (approved):

map_gen/maps/crash_site/events.lua
Jayefuu: I've now updated this.

Jayefuu:
Woops. I committed some changes to this branch I didn't mean to. Those need checking too @grilledham as I don't know how to uncommit them.

map_gen/maps/crash_site/outpost_builder.lua
```diff
@@ -1219,13 +1219,16 @@ local function do_artillery_turrets_targets()
...
+local server_player = {name = '<server>', print = print}
+
 local function set_pollution_multiplier(args, player)
     local multiplier = tonumber(args.multiplier)
     if not multiplier then
         player.print("Fail")
         return
     end
 
+    player = player or server_player
```
grilledham: This is too low, it would need to happen before `player.print` 4 lines up.

map_gen/maps/crash_site/outpost_builder.lua
```diff
@@ -1953,7 +1955,7 @@ Command.add(
     'get-pollution-multiplier',
     {
         description = {'command_description.get_pollution_multiplier'},
-        required_rank = Ranks.admin,
+        required_rank = Ranks.regular,
```
grilledham: What is the value of restricting this to regulars? If we let regulars do this we might as well let everyone.

map_gen/maps/crash_site/events.lua
```diff
@@ -318,6 +318,14 @@ destroyer_callback =
     end
 )
 
+local artillery_cooldown_callback
+artillery_cooldown_callback =
+    Token.register(
+    function(entity)
+        entity.minable = true
```
grilledham, line 329: Now that I think about this again we need to check the entity is still valid, it might have killed.

grilledham (changes requested):

map_gen/maps/crash_site/outpost_builder.lua
Jayefuu: None. I've switched it to guest.

grilledham (approved):

PR #1182, opened 2021-02-22
Adjusted barrel prices to balance new meta

Jayefuu:
Players were buying a hundred barrels for 15 coins. Made blue belts V easy among other things, adjusted at request of Tigress.


grilledham (approved):
10x seems quite high, but then I'm not sure what the current meta is. Do people buy barrels?
Either way happy to try this change we can always adjust it again.

Jayefuu:
@grilledham with the old price they were getting 100 barrels for 15 coins. Now set to 100 barrels for 150 coins. The meta wasn't to bother setting up any oil processing, sulfuric or lube production as buying an inventory full of barrels only cost a few hundred  coins.

PR #1184, opened 2021-02-24
crash site: enable steel research on start

lex:
This enables the players to see the steel icons on outpost furnaces on the map.

### DIFF
--- a/map_gen/maps/crash_site/outpost_builder.lua
+++ b/map_gen/maps/crash_site/outpost_builder.lua
@@ -1920,7 +1920,7 @@ Event.add(defines.events.on_entity_died, turret_died)
 
 Event.on_init(
     function()
-        game.forces.neutral.recipes['steel-plate'].enabled = true
+        game.forces.player.recipes['steel-plate'].enabled = true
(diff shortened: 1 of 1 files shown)


grilledham:
Thanks for your contribution, but I don't think this is the right solution. Perhaps instead we could insert metal plates into the furnaces when steel technology is researched. Or we make a second player force, assuming the first player force can see the recipes tags for the other player force.

lex:
The first option could work.

I tried going the allied force way, but it seems like you're unable to see allied forces' recipes on the map, so that doesn't seem like an option either.

Jayefuu:
An alternative.... when we call magic_item_crafting_callback, if furnace_item.name is iron-plate, if steel is not researched, enable the steel recipe, insert the iron plates then disable the steel recipe.

I tested this should work by loading a new game, enabling all technology and then charting the map.

Jayefuu:
Ok my suggestion doesn't work. The icons disappear when you disable the technology

grilledham:
I think #1188 fixes this.

PR #1185, opened 2021-02-26
Fixed evo floating text bug

Jayefuu:
We managed to get into a state where the evo never quite reached 1 because the code to set the evolution was inside the if statement. Therefore +0% evo would pop up for the whole of late game


grilledham (changes requested):
We did discuss this on discord, but I'll put it here incase you have forgotten. 

You probably want to check `round(extra*100,2) >= 0.01` instead. Consider what would happen if evolution was 0.5 but extra was 0.

PR #1188, opened 2021-02-28
Fix steel recipe not showing for magic crafters if player force hasn't unlocked steel tech.

grilledham:


Jayefuu:
This works? That's a really nice and simple solution if so, nice one.

Jayefuu (approved):

PR #1189, opened 2021-03-02
Dump player inventories 

Jayefuu:
New feature to dump player inventories to a corpse after they've been gone 15 minutes. Will spawn on their last position so if no one else wants their stuff it will be there when they get back.

Remaining issues:
1) I couldn't get the config.lua settings to work as I couldn't find a good example. I want it off by default then overwrite that from crashsite. See commented block on lines 19 to 23 of dump_offline_inventories.lua for my attempt that needs correcting.
EDIT: I should have put it in scenario.lua not dump_offline_inventories.lua, right?
2) I don't think I've made the corpse_util,lua player_corpses public properly. It doesn't destroy the tag when it is removed and I added code to stop the looting message when corpse.active is false. That's not working.
EDIT: I fixed the looting message
3) There's some debug I haven't taken out yet. I will remove it all once it's working.
4) The timer's currently set to 1 minute not 15.

### DIFF
--- a/features/corpse_util.lua
+++ b/features/corpse_util.lua
@@ -23,6 +23,10 @@ Global.register(player_corpses, function(tbl)
     player_corpses = tbl
 end)
 
+local function add_tag(tag, player_index, death_tick)
+    player_corpses[player_index * 0x100000000 + death_tick] = tag
+end
+
 local function player_died(event)
     local player_index = event.player_index
     local player = game.get_player(player_index)
@@ -83,7 +87,7 @@ local function player_died(event)
         end
     end
 
-    player_corpses[player_index * 0x100000000 + tick] = tag
+    add_tag(tag, player_index, tick)
 end
 
 local function remove_tag(player_index, tick)
@@ -99,12 +103,14 @@ local function remove_tag(player_index, tick)
     tag.destroy()
 end
 
-local function corpse_expired(event)
-    local entity = event.corpse
-
-    if entity and entity.valid then
-        remove_tag(entity.character_corpse_player_index, entity.character_corpse_tick_of_death)
+local function remove_corpse_tag(corpse)
+    if corpse and corpse.valid then
+        remove_tag(corpse.character_corpse_player_index, corpse.character_corpse_tick_of_death)
     end
+end
+
+local function corpse_expired(event)
+    remove_corpse_tag(event.corpse)
 end
 
 local corpse_util_mined_entity = Token.register(function(data)
@@ -131,14 +137,14 @@ local function mined_entity(event)
     local player_index = event.player_index
     local corpse_owner_index = entity.character_corpse_player_index
 
-    if player_index == corpse_owner_index then
+    if player_index == corpse_owner_index or not entity.active then
         return
     end
 
     local player = game.get_player(player_index)
     local corpse_owner = game.get_player(corpse_owner_index)
 
-    if player and corpse_owner and entity.active then
+    if player and corpse_owner then
         local message = table.concat {player.name, ' has looted ', corpse_owner.name, "'s corpse"}
         Utils.action_warning('[Corpse]', message)
     end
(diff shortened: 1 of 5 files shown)


features/dump_offline_inventories.lua
```diff
@@ -0,0 +1,84 @@
...
+--[[local config = global.config
+config.dump_offline_inventories = {
+    enabled = true,
+    offline_timout_mins = 1,   -- time after which a player logs off that their inventory is provided to the team
+}]]--
+
+local spawn_player_corpse =
+    Token.register(
+    function(player)
+        if player and player.valid and player.connected == false and offline_player_queue[player.index] then
```
grilledham: There should probably be some check that the player still has a character and the character is valid.

Also a small style thing but I'd invert this if statement and early return, that way the whole function body is one block less indented.

features/dump_offline_inventories.lua
```diff
@@ -0,0 +1,84 @@
...
+            end
+            for item_name, count in pairs(inv_trash_contents) do
+                inv_corpse.insert({name = item_name, count = count})
+            end
+
+            inv_main.clear()
+            inv_trash.clear()
+
+                local text = player.name .. "'s inventory (offline)"
+                local tag = player.force.add_chart_tag(player.surface, {
```
grilledham: tag could be `nil` I.e. it could fail to create the tag.

features/dump_offline_inventories.lua
grilledham: We need to remove the entry from `offline_player_queue` at some point before or after spawning the corpse else the player is still in the queue?

features/dump_offline_inventories.lua
```diff
@@ -0,0 +1,84 @@
...
+        offline_player_queue[event.player_index] = nil -- ensures they're not in the offline_player_queue for wealth redistribution
+        game.print("Player rejoined. Removed from offline list")
+    end
+)
+
+Event.add(
+    defines.events.on_pre_player_left_game,
+    function(event)
+        local player_index = event.player_index
+        local player = game.get_player(player_index)
```
grilledham: We should check if the player is `nil` or not valid.

features/corpse_util.lua
```diff
@@ -138,7 +138,7 @@ local function mined_entity(event)
     local player = game.get_player(player_index)
     local corpse_owner = game.get_player(corpse_owner_index)
 
-    if player and corpse_owner then
+    if player and corpse_owner and entity.active == true then
```
grilledham: Minor style issue.
```suggestion
    if player and corpse_owner and entity.active then
```

features/dump_offline_inventories.lua
```diff
@@ -0,0 +1,84 @@
+-- This feature allows you to turn on anti-hoarding so that X minutes after a player leaves the game 
+-- the resources in their inventory are returned to the teams
+
+-- To do
+-- What if player has no body when they leave?
+-- What if players is kicked?
+-- Do we want to allow donators and admins to keep their inventories as a perk? Or is that too pay to win?
```
grilledham: No to pay to win.

features/dump_offline_inventories.lua
```diff
@@ -0,0 +1,84 @@
+-- This feature allows you to turn on anti-hoarding so that X minutes after a player leaves the game 
+-- the resources in their inventory are returned to the teams
+
+-- To do
+-- What if player has no body when they leave?
+-- What if players is kicked?
```
grilledham: If a player is ban we should dump immediately. No special treatment for kicked.

features/dump_offline_inventories.lua
Jayefuu: I've inverted the statement. Is this part right? `or not offline_player_queue[player.index]`? I seem to remember me getting something like that wrong before.

Also do I really need a character check? I thought when offline the player doesn't have a character....

Jayefuu:
Remaining:

-- test what happens when timer reaches 0 
-- Check that I haven't messed up anything in corpse_util. Test normal corpses and tags still despawn and that the looting message still works
-- Investigate desync seen on S10. Incorrect setup of offline_player_queue?
-- Implement switch of state to active to allow timeout and automatic deletion of corpse and tag.

Jayefuu:
@grilledham one bug remains (that I can find).

If you open GUI or mine a corpse when the corpse is less than 15 minutes old, the corpse despawns as expected and the tag is removed from the map.

If the corpse is older than 15 minutes then the corpse despawns on mining/GUI empty but the tag remains.

Jayefuu:
Nevermind I found the cause. I changed the tick of death in order to test faster, so it didn't match the tag. Woops. It was a bug I created :P

grilledham (approved):

PR #1214, opened 2021-05-19
Added a Circles preset for Danger Ores.

Soggs:


map_gen/maps/danger_ores/presets/danger_ore_circles.lua
```diff
@@ -0,0 +1,151 @@
...
+concrete_on_landfill({tile = 'blue-refined-concrete'})
+
+local main_ores_builder = require 'map_gen.maps.danger_ores.modules.main_ores_circles'
+
+local config = {
+    spawn_shape = b.circle(64),
+    start_ore_shape = b.circle(68),
+    main_ores_builder = main_ores_builder,
+    main_ores = main_ores_config,
+    main_ores_shuffle_order = true,
```
grilledham, line 127: So this doesn't do anything for this preset, right? We should either remove or set to `false`.

map_gen/maps/danger_ores/modules/main_ores_circles.lua
```diff
@@ -0,0 +1,57 @@
...
+
+
+return function(config)
+    local main_ores = config.main_ores
+    local main_ores_split_count = config.main_ores_split_count or 1
+
+    main_ores = Helper.split_ore(main_ores, main_ores_split_count)
+
+    local function condition_factory(ore_name)
+      return function(x, y, _)
```
grilledham, line 30: Everything inside the `function(x, y, _)` function gets called per tile. I think lines 19 to 25 could be in the `condition_factory` function instead.

map_gen/maps/danger_ores/presets/danger_ore_circles.lua
```diff
@@ -0,0 +1,151 @@
+local RS = require 'map_gen.shared.redmew_surface'
+local MGSP = require 'resources.map_gen_settings'
+local Event = require 'utils.event'
+local b = require 'map_gen.shared.builders'
+local Config = require 'config'
+
+local ScenarioInfo = require 'features.gui.info'
+ScenarioInfo.set_map_name('Danger Ore Waves')
```
grilledham:
```suggestion
ScenarioInfo.set_map_name('Danger Ore Circles')
```

map_gen/maps/danger_ores/modules/main_ores_circles.lua
```diff
@@ -0,0 +1,57 @@
...
+          local ore_name = ore_data.name
+          local tiles = ore_data.tiles
+          local land = tile_builder(tiles)
+
+          local ratios = ore_data.ratios
+          local weighted = b.prepare_weighted_array(ratios)
+          local amount = ore_data.start
+
+          local ore = ore_builder(ore_name, amount, ratios, weighted)
+          local condition = condition_factory(ore_name)
```
grilledham, line 56: This could be moved outside the loop. That way the check would be once per tile, rather than once per tile per ore. But doing that is a bit complicated, so we could leave for later.

grilledham (commented):
Cool, a few minor tweaks and some (optional) optimization would be nice. The lack of starting ore (see below) is probably my only concern.

Can you add this to <https://github.com/Refactorio/RedMew/tree/develop/scenario_templates> That way it will appear on the web interface when merged.

All the danger ore maps have unsorted starting ores. For this one the starting ore is only coal. It would be good to add starting ore for iron/copper, but I'm not sure how best to do that for this ore pattern.

map_gen/maps/danger_ores/presets/danger_ore_circles.lua
Soggs: Shuffle order should be back. I am just not sure it actually does anything of significance.

Rest gets used again especially ``start_ore_shape``.

map_gen/maps/danger_ores/modules/main_ores_circles.lua
Soggs: I think i moved all that can be moved outside, outside.

Soggs:
The starting ores are not 100% reliable. But otherwise it should be working.

map_gen/maps/danger_ores/modules/main_ores_circles.lua
```diff
@@ -0,0 +1,68 @@
...
+          local shape = b.choose(condition, b.apply_entity(land, ore), b.empty_shape)
+          table.insert(shapes, shape)
+      end
+
+      local ores = b.any(shapes)
+
+      local starting_ores = b.change_collision_tile(start_ore_shape, 'water-tile', 'grass-1') -- For some reason that does not always overwrite water tiles?
```
grilledham: You might want `b.change_map_gen_collision_tile` instead. `change_collision_tile` only changes tile created by the shapes, whereas `change_map_gen_collision_tile` changes tiles created by factorio's map generation.
```suggestion
      local starting_ores = b.change_map_gen_collision_tile(start_ore_shape, 'water-tile', 'grass-1')
```

grilledham (commented):
I'll give this a proper check tomorrow, but on first inspection it looks good.

map_gen/maps/danger_ores/modules/main_ores_circles.lua
```diff
@@ -0,0 +1,68 @@
...
+    local function condition_factory(ore_name)
+      local scale = config.circle_scale or 1
+      --local randomize_scale = config.randomize_scale or false
+      local ore_table = {}
+      for _, ore_data in pairs(main_ores) do
+        table.insert(ore_table, {name = ore_data.name, weight = ore_data.weight})
+      end
+      local weighted = b.prepare_weighted_array(ore_table)
+      return function(x, y, _)
+        local i =  floor(sqrt(x * x + y * y) * scale) % weighted.total + 1
```
grilledham, line 31:
```suggestion
        local i =  floor(sqrt(x * x + y * y) * scale + 1) % weighted.total + 1
```
So the first ring is iron rather than coal. It's minor but makes the early game slightly easier.

grilledham: Ignore this, I'm dumb. This won't work when shuffle ore is being used.

Soggs: Also at the moment it just works out to be coal since technically I am building from the center of the map. If you change the throttle or types, weights and/or order in the supplied ``ore_table`` it will just change. It is also the reason that first ring usually does not have same width as the other rings as it gets clipped by the spawn and the starting ores.

grilledham (approved):

PR #1216, opened 2021-05-22
Fix the desync on Danger Ores Circles.

Soggs:
Moved the tables.shuffle around so it gets the ``random_gen`` passed. Which required a slight change to the ``condition_factory``. According to tests I just did on s10 it should now be desync proof.


map_gen/maps/danger_ores/modules/main_ores_circles.lua
```diff
@@ -2,67 +2,50 @@ local Helper = require 'map_gen.maps.danger_ores.modules.helper'
...
+        if config.main_ores_shuffle_order then
+            table.shuffle_table(starting_ores_list, random_gen)
+            table.shuffle_table(shapes, random_gen)
+        end
 
-      local ores = b.any(shapes)
+        local starting_ores = b.segment_weighted_pattern(starting_ores_list)
+        local ores = b.ring_weighted_pattern(shapes, circle_thickenss)
 
-      local starting_ores = b.change_map_gen_collision_tile(start_ore_shape, 'water-tile', 'grass-1')
```
Soggs, line 63: Does the change in line 49 to the order in ``b.any(...)`` prevent water from overwriting the start ores?

map_gen/shared/builders.lua
```diff
@@ -1517,6 +1517,44 @@ function Builders.segment_weighted_pattern(pattern)
     end
 end
 
+function Builders.ring_pattern(pattern, thickness)
```
Soggs, line 1520: Do you want documentation for this on the wiki part and if so, how do I do that? :D

map_gen/maps/danger_ores/modules/main_ores_circles.lua
grilledham: No, I changed the starting ore to use the land shape line 27. land being the same shape that produces the tiles for the main ores. Whereas before starting ore was using the start_ore_shape for the tiles.

land returns the tile name to use, whereas start_ore_shape returns true or false. true means to use the vanilla map gen tile, false I think of as transparent void.

map_gen/shared/builders.lua
grilledham: You don't have to, but if you did that would be cool. You should be able to edit the wiki directly <https://github.com/Refactorio/RedMew/wiki/Using-the-Builders/_edit> no PR required. If not it's probably a permission issue, let me know and I'll fix it.

grilledham (approved):

PR #1230, opened 2021-09-10
Landfill danger ores preset bug

Jayefuu:
Zergslayer found a bug in the danger_ore_landfill preset. Conversation starts here:
https://discord.com/channels/432567222481846283/432570636544507905/885327812611952680

I found that the ore preset was misspelled. This branch updates it.

Untested. Will update when I hear back from Zergslayer if it works.


grilledham (approved):
LGTM. Thanks for doing this.

PR #1237, opened 2021-09-12
Showing image(details)

Claude477:
When creating map is showing image(details) is make easier dev-build maps


grilledham:
image.png is a too generic name. Call it something like map_guide.png. Also put a `.` at the front of the name i.e. `.map_guide.png` so that the image doesn't add 0.5MBs to everyone saves.

Claude477:
Tryed change name but Factorio not showing up image if different name T_T

PR #1241, opened 2021-09-14
Changes to display names

Jayefuu:
- Added venice
- Changed part of print to discord to display names instead of names

UNTESTED

### DIFF
--- a/map_gen/maps/crash_site/commands.lua
+++ b/map_gen/maps/crash_site/commands.lua
@@ -78,7 +78,8 @@ function Public.control(config)
         ['crashsite-world'] = 'Crash Site World Map',
         ['crashsite-desert'] = 'Crash Site Desert',
         ['crashsite-arrakis'] = 'Crash Site Arrakis',
-        ['crashsite-venice'] = 'Crash Site Venice'
+        ['crashsite-venice'] = 'Crash Site Venice',
+        ['crashsite-manhattan'] = 'Crash Site Manhattan'
     }
 
     local function can_restart(player)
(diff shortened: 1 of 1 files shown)


map_gen/maps/crash_site/commands.lua
```diff
@@ -77,7 +77,8 @@ function Public.control(config)
         ['crashsite'] = 'Crash Site',
         ['crashsite-world'] = 'Crash Site World Map',
         ['crashsite-desert'] = 'Crash Site Desert',
-        ['crashsite-arrakis'] = 'Crash Site Arrakis'
+        ['crashsite-arrakis'] = 'Crash Site Arrakis',
+        ['crashsite-venice'] = 'Crash Site Venice'
```
grilledham: Also add Manhattan.

```suggestion
        ['crashsite-venice'] = 'Crash Site Venice',
        ['crashsite-manhattan'] = 'Crash Site Manhattan'
```

map_gen/maps/crash_site/commands.lua
```diff
@@ -243,7 +244,7 @@ function Public.control(config)
 
         Server.to_discord_named_raw(map_promotion_channel,
             crash_site_role_mention .. ' **' .. scenario_display_name[config.scenario_name] .. ' has just restarted!!\\n'
-                .. 'Next map: ' .. new_map_name .. '**')
+                .. 'Next map: ' .. scenario_display_name[new_map_name] .. '**')
```
grilledham: In case the lookup fails.
```suggestion
                .. 'Next map: ' .. scenario_display_name[new_map_name] or new_map_name .. '**')
```

map_gen/maps/crash_site/commands.lua
```diff
@@ -243,7 +244,7 @@ function Public.control(config)
 
         Server.to_discord_named_raw(map_promotion_channel,
             crash_site_role_mention .. ' **' .. scenario_display_name[config.scenario_name] .. ' has just restarted!!\\n'
```
grilledham: In case the lookup fails.
```suggestion
            crash_site_role_mention .. ' **' .. scenario_display_name[config.scenario_name] or config.scenario_name .. ' has just restarted!!\\n'
```

grilledham (changes requested):
Thanks for investigating, just a few minor comments.
We should probably also test this. If you are feeling lazy I can do that after you make the changes.

grilledham (approved):

PR #1244, opened 2021-09-25
Update venice

Claude477:
Remove bottom text in map


grilledham (changes requested):
We could crop the image to remove the water at the bottom.
Also can you commit the new `map_gen/data/.source_images/Venice.png`

grilledham (approved):

PR #1247, opened 2021-09-27
Updated the starting pollution multiplier

Jayefuu:
The new meta sucks. No one builds any walls or defences. Players can go 3 or 4 hours with no attacks so long as they craft nothing.

Now that the pollution from outposts is generated at position = {0,0}.... let's try a very small amount of pollution for every outpost captured to force at least a small amount of defences to be built.

This will be MUCH less pollution than if all of the items they get free were made in base, but it will still force at least a few defenses or clearing of nearby bases.

Starting with a value of 10 to see how it goes.

### DIFF
--- a/map_gen/maps/crash_site/outpost_builder.lua
+++ b/map_gen/maps/crash_site/outpost_builder.lua
@@ -1235,7 +1235,7 @@ local function set_pollution_multiplier(args, player)
         end
         -- update the value to the minimum and continue to message all admins
         player.print("Setting magic crafter pollution multiplier to the minimum value of " .. base_pollution_multiplier)
-        multiplier = base_pollution_multiplier        
+        multiplier = base_pollution_multiplier
     end
 
     local old_multiplier = pollution_multiplier.value
--- a/map_gen/maps/crash_site/outpost_builder.lua
+++ b/map_gen/maps/crash_site/outpost_builder.lua
@@ -1228,10 +1228,14 @@ local function set_pollution_multiplier(args, player)
         return
     end
 
-    if multiplier < base_pollution_multiplier then
-        pollution_multiplier.value = base_pollution_multiplier
-        player.print("Cannot set the multiplier lower than the base multiplier of " .. base_pollution_multiplier)
-        return
+    if multiplier < base_pollution_multiplier then  
+        if base_pollution_multiplier == pollution_multiplier.value then -- no change, so not necessary to message all admins and update the value
+            player.print("Magic crafter pollution is already at minimum value of " .. base_pollution_multiplier)
+            return
+        else -- update the value to the minimum and continue to message all admins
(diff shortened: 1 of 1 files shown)


map_gen/maps/crash_site/outpost_builder.lua
```diff
@@ -1227,6 +1228,11 @@ local function set_pollution_multiplier(args, player)
         return
     end
 
+    if multiplier < base_pollution_multiplier then
+        pollution_multiplier.value = base_pollution_multiplier
+        player.print("Cannot set the multiplier lower than the base multiplier of " .. base_pollution_multiplier)
```
grilledham: Missing a `return`?
```suggestion
        pollution_multiplier.value = base_pollution_multiplier
        player.print("Cannot set the multiplier lower than the base multiplier of " .. base_pollution_multiplier)
        return
```

Jayefuu: Yis. Fixed.

map_gen/maps/crash_site/outpost_builder.lua
```diff
@@ -1227,6 +1228,12 @@ local function set_pollution_multiplier(args, player)
         return
     end
 
+    if multiplier < base_pollution_multiplier then
+        pollution_multiplier.value = base_pollution_multiplier
```
grilledham:
```suggestion        
```
Looking at this again, I'm not sure what this line is for. What if `pollution_multiplier.value` was previously set to a valid value that is greater than `base_pollution_multiplier`. Do you want to silently change `pollution_multiplier.value` to a different value because someone try to set it to an invalid value?

Or am I missing something?

grilledham (changes requested):

map_gen/maps/crash_site/outpost_builder.lua
```diff
@@ -1227,6 +1228,16 @@ local function set_pollution_multiplier(args, player)
...
     end
 
+    if multiplier < base_pollution_multiplier then
+        if base_pollution_multiplier == pollution_multiplier.value then -- no change, so not necessary to message all admins and update the value
+            player.print("Magic crafter pollution is already at minimum value of " .. base_pollution_multiplier)
+            return
+        else -- update the value to the minimum and continue to message all admins
+            player.print("Setting magic crafter pollution multiplier to the minimum value of " .. base_pollution_multiplier)
+            multiplier = base_pollution_multiplier
+        end
```
grilledham:
```suggestion
        end
        -- update the value to the minimum and continue to message all admins
        player.print("Setting magic crafter pollution multiplier to the minimum value of " .. base_pollution_multiplier)
        multiplier = base_pollution_multiplier        
```

grilledham (approved):
Looks good.

One minor style point, you don't really need the `else` after the `return`.

PR #1248, opened 2021-09-30
Added barrage feature to give players more end-game clearing options

Jayefuu:
- Changed spawn to add 2 more chests after feedback from Tigress about end-game use of strike command
- Added floating sprites above chests to indicate what to put in each. 
- Added barrage command and items in market
- Barrage sends explosive rockets at enemy nests. Players need to judge whether there are enough nests for it to be worth a strike and decide if there are too many nests and that the current level of barrage won't kill them all.
- First pass at balancing costs made.

### DIFF
--- a/map_gen/maps/crash_site/commands.lua
+++ b/map_gen/maps/crash_site/commands.lua
@@ -29,9 +29,11 @@ function Public.control(config)
     Restart.set_start_game_data({type = Restart.game_types.scenario, name = config.scenario_name or 'crashsite'})
 
     local airstrike_data = {radius_level = 1, count_level = 1}
+    local barrage_data = {radius_level = 1, count_level = 1}
 
-    Global.register({airstrike_data = airstrike_data}, function(tbl)
+    Global.register({airstrike_data = airstrike_data, barrage_data = barrage_data}, function(tbl)
         airstrike_data = tbl.airstrike_data
+        barrage_data = tbl.barrage_data
     end)
 
     local static_entities_to_check = {
@@ -333,7 +335,7 @@ function Public.control(config)
         local ypos = data.ypos
         local player = data.player
         local tag = player.force.add_chart_tag(player.surface, {
-            icon = {type = 'item', name = 'poison-capsule'},
+            icon = {type = 'item', name = data.item},
             position = {xpos, ypos},
             text = player.name
         })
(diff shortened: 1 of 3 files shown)


Jayefuu:
New spawn look:

image

map_gen/maps/crash_site/commands.lua
```diff
@@ -425,18 +427,19 @@ function Public.control(config)
...
         local dropbox = entities[1]
 
         if dropbox == nil then
-            player.print("Chest not found. Replace it here: [gps=-0.5,-3.5,redmew]")
+            player.print("Chest not found. Replace it here: [gps=2.5,-7.5,redmew]")
             return
         end
 
-        -- process each set of coordinates with a 10 strike limit
+        -- process each set of coordinates with a 20 strike limit
```
grilledham: I think the limit was 10 because you need two indexes per coordinate (one for x, one for y). So either this comment is now wrong, or `i < 20` below is wrong.

map_gen/maps/crash_site/commands.lua
```diff
@@ -450,32 +453,115 @@ function Public.control(config)
...
+        -- Check that the chest is where it should be.
+        local entities = s.find_entities_filtered {position = {2.5, -1.5}, type = 'logistic-container', limit = 1}
+        local dropbox = entities[1]
+
+        if dropbox == nil then
+            player.print("Chest not found. Replace it here: [gps=2.5,-1.5,redmew]")
+            return
+        end
+
+        -- process each set of coordinates from the arguments with a 20 strike limit
```
grilledham: Same again here with 20 strike limit but `i < 20` implies a 10 strike limit.

map_gen/maps/crash_site/scenario.lua
```diff
@@ -872,6 +893,11 @@ Global.register_init(
     {},
     function(tbl)
         game.create_force(cutscene_force_name)
+
+        -- Sprites for the spawn chests. Is there a better place for these?
```
grilledham, line 898: Seems reasonable enough to me, it's near the code that has the position of the chests.

grilledham (commented):
I haven't tested this yet, but on first inspection of the code this looks good.

grilledham:
Does this still have the issue where destroyed nested don't spawn biters/spitters?

Jayefuu:
> 
> 
> Does this still have the issue where destroyed nested don't spawn biters/spitters?

No, I didn't fix it, I just didn't test it right. I think the issue I saw must have been due to low evolution when I tested it. I just hopped on the live test and it spawned biters fine when nests died.

I'll fix the comments you mention above.

grilledham (approved):

PR #1249, opened 2021-10-16
Fixed russian translation

SaLLIa666:
Text has grammatical errors, now it's better


Claude477:
Yeah thx for translation, but we have different type for translation if want check out ->https://crowdin.com/project/redmew

SimonFlapse (changes requested):
Thanks for submitting your contribution!

I would love to approve these changes but I need to do some work before we can accept manual translations.

For future reference we prefers the use of our crowdin project to limit the workload.

Thanks again for the initiative. You do not need to do anything else. However you are free to input these suggestions on crowdin too. Otherwise I will manually add them ASAP.

PR #1250, opened 2021-11-15
qol when mining an inserter, pick up item an inserter dropped on the ground

jacobwatkinsgit:


grilledham (approved):
Very nice! Thanks for making this PR.

PR #1262, opened 2021-12-14
Updated donor features

Jayefuu:
Features:
- Team buffs for donators with tier 2 or higher Patreon status
- Tier 2 (coal): +10% team hand mining bonus per online donator, up to a maximum of 10 online donators
- Tier 3 (solid fuel): +10% team hand crafting bonus per online donator, up to a max of 10 online donators
- Tier 4 (rocket fuel): +10% team run speed per online donator, up to a max of 10 online donators
- Tier 5 (nuclear fuel): +5 inventory slots for all. Up to max of 30 extra. Does not reduce down when donator logs off.

To do:
- How do I call a function on_load? The commend on line 344 didn't work. I want to check the list of applied bonuses when the server loads after a crash or stop to see who is online and apply the correct buffs.
- Update donator_tiers[X].max to 10, 10, 10, 10 and 6. The low values are only there for testing.

Tests to perform before merging:
- Use this command to check the current modifiers: /c local force = game.player.force game.print("Mining modifier: "..force.manual_mining_speed_modifier.." Crafting modifier: "..game.player.force.manual_crafting_speed_modifier.." Run modifier: "..game.player.force.character_running_speed_modifier.." Extra slots: "..force.character_inventory_slots_bonus)
- Check it doesn't interact badly with diggy mining speed
- Confirm that it works correctly with the market mining and running boosts
- Ensure that the max values work correctly. For now the limit has been set to max 2 donators per level. This will change back once tested.


features/donator.lua
```diff
@@ -52,19 +66,81 @@ local function player_joined(event)
         return nil
     end
 
-    local messages = d.welcome_messages
-    if not messages then
+    local perk_flag = d.perk_flags
```
grilledham, line 102: `perk_flags` isn't the donator tier, it's a flag enum for which perks are enabled for the player. We would either need to add a donator tier to the scenario data for donators, or add entries for each feature to the perk_flags. I prefer the second option.

grilledham (commented):
There are going to be some annoying edge cases we will have to solve, some of which you have noticed yourself.

- Works with other features that change these values (diggy, market).
- More than one 'player' force, i.e. when there are multiple teams.
- Donators gaining/losing donators status/tier when they are in game.
- Admins/mods changing theses values. Basically situations when it's not under our control because it's not in the Redmew codebase.

Jayefuu:
@grilledham:

> Works with other features that change these values (diggy, market).
- Market isn't a problem as this modifies player.character_running_speed_modifier whereas the new feature modifies player.force.character_running_speed_modifier. I will run some tests to confirm this is fine.
- From my quick look it seems it's compatible with diggy as diggy keeps track of modifiers and increments and decrements them rather than setting them to an absolute value. Again, I'll devise some tests and report back.
> More than one 'player' force, i.e. when there are multiple teams.
- I think I solved this by always using player.force. It should apply different bonuses to different forces depending upon which the player joins. Though I think this is moot since we don't have any multiplayer maps. Unless the player joins a lobby on join as is the case with space race. Maybe I need to add a function for when a player changes force to move the buffs from one force to another?
> Donators gaining/losing donators status/tier when they are in game.
- Good catch. I will add a condition to the on_leave event that checks if their name is in the donator_temp table instead of the donators table.
> Admins/mods changing theses values. Basically situations when it's not under our control because it's not in the Redmew codebase.
- I don't think we can solve for this and it's rarely used, right? If you've got suggestions to implement something I can look at it.

grilledham:
> I think I solved this by always using player.force. It should apply different bonuses to different forces depending upon which the player joins.

Doesn't that make this p2w if not applied to all players?
Though I agree we rarely play pvp maps, I bet when an admin randomly decides to launch one in the future, they won't think about this donator feature.

> I don't think we can solve for this and it's rarely used, right? If you've got suggestions to implement something I can look at it.

It's a tough one, probably unsolvable. At the very least it means we have to check we are not setting the values to something invalid (as we can't rely on the values not being changed out of our control) as I think the game errors if you try that.

We also need to check the 'performance-scale-set' command.

Jayefuu:
@grilledham 
 
> Doesn't that make this p2w if not applied to all players? Though I agree we rarely play pvp maps, I bet when an admin randomly decides to launch one in the future, they won't think about this donator feature.

Hmmm. I took a look at the code and possibilities and wrote some suggested solutions before I implement them. Which one do you prefer?
1) We ignore it and accept that one side could possibly have a small advantage, depending on the scenario
2) We provide an option in config for whichever admin is setting it up to disable donator features easily.
3) We iterate through all the forces each time we apply or remove buffs and apply them to each force. 
4) Same as 3) but we check it has players on the force with LuaForce players().

I like the idea that 3 or 4 would solve the problem, but we can't be sure what is happening with future scenarios and how they handle players moving teams or admins shifting about to balance things. What if someone joins the game on one team, is moved to another then leaves and the buff is removed from the wrong team?

I think 2) is the best option as if it's something in our repo the config can adjusted from within the multiplayer scenario.

> It's a tough one, probably unsolvable. At the very least it means we have to check we are not setting the values to something invalid (as we can't rely on the values not being changed out of our control) as I think the game errors if you try that.
> 
> We also need to check the 'performance-scale-set' command.

I will add code to ensure we don't try set the multipliers below 0 and will look into performance-scale-set

features/donator.lua
Jayefuu: What are the benefits of using the enum? Having the donator tier would be easier for the people that are updating the scenario data wouldn't it?

Jayefuu:
@grilledham I added the checks that you asked for. Please could you help me with checking compatibility with performance-scale-set?

config.lua
```diff
@@ -433,6 +433,9 @@ global.config = {
         print_to_force = true, -- print a message to force chat when that force finishes a new research.
         print_to_discord = true, -- print a message to the discord channel when the player force finishes a new research.
         ignore_script = false -- ignore researches unlocked by commands or by code.
+    },
+    donator_perks = {
+        enabled = true
     }
```
grilledham: I think these where by module (file) then by feature inside that module
```suggestion
    donator = {
        donator_perks = {
            enabled = true
        }
    }
```

features/donator.lua
```diff
@@ -11,17 +11,32 @@ local concat = table.concat
 local remove = table.remove
 local set_data = Server.set_data
 local random = math.random
+local config = global.config.donator_perks
```
grilledham: This needs to be restored on_load as the global table used during the control stage and after on_laod are different instances.
You can do that in Global.register below
```lua
config = tbl.global.config.donator_perks
```
Same problem in features/donator_commands.lua

features/donator.lua
grilledham: Allows donators to opt out of a feature. Currently for each donator perk we have a way to turn it off per player.

grilledham (commented):
> Hmmm. I took a look at the code and possibilities and wrote some suggested solutions before I implement them. Which one do you prefer?
> 1. We ignore it and accept that one side could possibly have a small advantage, depending on the scenario
> 2. We provide an option in config for whichever admin is setting it up to disable donator features easily.
> 3. We iterate through all the forces each time we apply or remove buffs and apply them to each force.
> 4. Same as 3) but we check it has players on the force with LuaForce players().

Option 2) seems reasonable to me.

features/donator.lua
Jayefuu: Did [this ](https://github.com/Refactorio/RedMew/pull/1262/commits/134c15f4db5e26b55f26f3b4ceef8a42a3477e1a) commit fix this issue how you wanted it? I'm not sure I fully understood the why but think I got all your comments in there?

Jayefuu:
@grilledham
>Allows donators to opt out of a feature. Currently for each donator perk we have a way to turn it off per player.

Is this up for debate? I've been thinking about this. I don't agree there is a need to turn features on and off by player. If they don't want a feature, why would they buy that level of patreon?

My main problem with it is ease of maintenance. The value in the scenario data should be the straight patreon level value. I'm sure whoever is maintaining the Patreon perk levels (currently Tigress) is capable of understanding binary masks and applying the right number corresponding to a perk level, but should they have to go to a post in the moderation guide to get it right?

Is it acceptable if I create a new record at donators[player_name].patreon_tier for this feature, or do you really want me to add new flags to donator_perks.lua ?

grilledham:
@Jayefuu
> Is this up for debate? I've been thinking about this. I don't agree there is a need to turn features on and off by player. If they don't want a feature, why would they buy that level of patreon?

I would still like the functionality to turn donator perks off per player. I don't think all donators choose to donator just for the donator perks, some donator to support, for example the top tier donators that never wanted a map made. Also if a situation occured where a donator wants a t4 feature but doesn't want a t3 feature at what level should they donate? I would prefer to be able to accommodate such a request rather than tell them to not donate. 

> My main problem with it is ease of maintenance....

We should add something to moderator guide explaining how it works. Something with what values for which perks and just add all the values together.

> Is it acceptable if I create a new record at donators[player_name].patreon_tier for this feature, or do you really want me to add new flags to donator_perks.lua ?

I think adding extra flags to donator_perks.lua is the way to go, the code already exists.

features/donator.lua
grilledham: Yes, I think that should work.

Jayefuu:
No problem. I'll just not be lazy then I guess :)

Jayefuu:
@grilledham I managed a bit of progress over last couple of weeks, but haven't done much testing or checking the logic. I pushed it anyway since I won't have time for the next few weeks and didn't want to lose the changes I'd made.

Claude477:
Hi checked lua-error "perk_flag" is needed or can to make comment ? for it so CI/build give green light

grilledham (changes requested):
I think we still need to address player's donator perks changing while they are in game. Let me know if you need any help with that.

Jayefuu:
> I think we still need to address player's donator perks changing while they are in game. Let me know if you need any help with that.

I've got no plans to work on this soon, I've been busy and not had time /interest.

features/donator.lua
```diff
@@ -76,6 +76,16 @@ local print_after_timeout =
     end
 )
 
+-- Just in case we want to turn it off mid-game, we can use /sc package.loaded['features.donator'].toggle_perks()
+function Public.toggle_perks()
+    config.donator_perks.enabled = not config.donator_perks.enabled
+    if config.donator_perks.enabled == true then
+        game.print("Donator perks now enabled")
```
grilledham, line 82: Might want to also do `print("Donator perks now enabled")`. `game.print` is not shown in the server console.

features/donator.lua
```diff
@@ -121,8 +129,8 @@ local function player_joined(event)
    end
 
     -- Update team perks
-    if not donatator_perks_temp[player.name] then -- check they're not already in donator_perks_temp table, this keeps track of bonuses that are added and removed as players join and leave
-        local donator_perk_msg = concat({"Donator Tier: ",donator_tiers[d.patreon_tier].name, ". Team bonuses applied: "})
+    if not donator_perks_temp[player.name] then -- check they're not already in donator_perks_temp table, this keeps track of bonuses that are added and removed as players join and leave
+        local donator_perk_msg = concat({"Donator Tier: ",donator_tiers[d.patreon_tier].name, ". Team bonuses applied for "..player.name..": "})
```
grilledham: You are mixing two different ways to do string concatenation here, I'd suggest this:
`local donator_perk_msg = concat({"Donator Tier: ", donator_tiers[d.patreon_tier].name, ". Team bonuses applied for ", player.name, ": "})`

features/donator.lua
```diff
@@ -138,40 +146,44 @@ local function player_joined(event)
...
     local player = game.get_player(event.player_index)
-    if not player or not player.valid or not config.enabled then
+
+    if not player or not player.valid or not config.donator_perks.enabled then
         return
     end
 
     local d = donators[player.name]
-    if not d then
+    if donator_perks_temp[player.name] then
```
grilledham: This might be simpler to read as
```lua
if not donator_perks_temp[player.name] and not d then
    return
end
```

features/donator_commands.lua
```diff
@@ -88,11 +88,12 @@ end
 -- Prints a list of currently active perks to the player
 local function print_perks(_, player)
     if not config.enabled then
-        player.print("Donator perks have been disabled for this map. It may be multiplayer or affect balance.")
+        player.print("Donator perks have been disabled for this map.")
```
grilledham: Probably should've mentioned this earlier but this won't work from the server console. I'd recommend putting something like:
```lua
local print = print
if player and player.valid then
    print = player.print
end
```
at the top of the function and use that instead of `player.print`

grilledham (changes requested):

grilledham:
I think we should do a live test, then if that is successful we can merge.

features/donator.lua
```diff
@@ -52,19 +99,116 @@ local function player_joined(event)
...
+            donator_tiers[3].count = donator_tiers[3].count + 1
+            donator_perk_msg = concat({donator_perk_msg, "+10% hand crafting. "})
+        end
+        if running_flag and donator_tiers[4].count < donator_tiers[4].max  then  -- Apply Tier 4 (Rocket Fuel) reward: + 10 % team running speed bonus per online tier 4+ donator
+            player.force.character_running_speed_modifier = player.force.character_running_speed_modifier + 0.1
+            donator_tiers[4].count = donator_tiers[4].count + 1
+            donator_perk_msg = concat({donator_perk_msg, "+10% run speed. "})
+        end
+        donator_perks_temp[player.name] = perk_flag
+        if inventory_flag and not donator_perks_perm[player.name] and donator_tiers[5].count < donator_tiers[5].max  then
```
Jayefuu, line 151: This needs changing. The message is right the first time a tier 5 donator joins, but it needs print it regardless of whether they're in the permanent upgrades table.

grilledham (approved):

PR #1268, opened 2021-12-26
Eating fish causes a little flying text to show other players you're …

Jayefuu:
Eating fish causes a little flying fish to appear above the player. Allows other players to see who's trying to heal themselves so they can focus on helping that player if needed.

### DIFF
--- a/features/fish_burps.lua
+++ b/features/fish_burps.lua
@@ -11,10 +11,8 @@ local function capsule_used(event)
         return
     end
 
-    local dx = math.abs(event.position.x - player.position.x)
-    local dy = math.abs(event.position.y - player.position.y)
-
-    if dx > 0.5 or dy > 1 then  -- Only want to create the flying text if the player clicks near their character for healing
+    local health_ratio = player.character.get_health_ratio()
+    if health_ratio == 1 then -- to stop people spamming it
         return
     end
 
--- a/features/fish_burps.lua
+++ b/features/fish_burps.lua
@@ -7,7 +7,7 @@ local function capsule_used(event)
     end
 
     local player =  game.get_player(event.player_index)
-    if not player or not player.valid then
+    if not player or not player.valid or not player.character or not player.character.valid then
         return
     end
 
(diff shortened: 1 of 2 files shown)


features/fish_burps.lua
```diff
@@ -0,0 +1,29 @@
...
+
+    local player =  game.get_player(event.player_index)
+    if not player or not player.valid then
+        return
+    end
+
+    local dx = math.abs(event.position.x - player.position.x)
+    local dy = math.abs(event.position.y - player.position.y)
+
+    if dx > 0.5 or dy > 1 then  -- Only want to create the flying text if the player clicks near their character for healing
```
grilledham: > Only want to create the flying text if the player clicks near their character for healing

Just curious, what is the reason for this?

Jayefuu: You can use a raw-fish capsule from any range. Click anywhere and it will use up a fish.

I only wanted to make the fish signal when they actually use a fish to heal not any time a fish is spammed.

grilledham: > I only wanted to make the fish signal when they actually use a fish to heal not any time a fish is spammed.

But doesn't it heal you regardless of where you click relative to your character?

Jayefuu: Oh I'm not sure. I'll test that.

Jayefuu: @grilledham you were right! Can't believe I've played 5 years and didn't know or forgot that :D

Jayefuu:
@grilledham I made a couple of changes to stop it being spammy if people keep eating fish when they have full health. And removed the requirement for them to click on their character with the fish capsule. Then I retested it and it works as intended.

features/fish_burps.lua
```diff
@@ -0,0 +1,27 @@
...
+    if event.item.name ~= "raw-fish" then
+        return
+    end
+
+    local player =  game.get_player(event.player_index)
+    if not player or not player.valid then
+        return
+    end
+
+    local health_ratio = player.character.get_health_ratio()
```
grilledham, line 14: character could be `nil` or invalid.

Jayefuu: Thanks. Fixed.

grilledham (approved):

PR #1277, opened 2022-01-16
toxic_jungle

Claude477:
toxic_jungle(Scenario making)


grilledham:
The name of the directory in scenario_templates should be `toxic-jungle` with a `-` instead of `_` to match the convention of the other scenarios.

Claude477:
ohh ok

PR #1278, opened 2022-01-16
beach

Claude477:
beach(Scenario-making)


grilledham:
Can we change the directory name to beach with a lower case b.

Claude477:
yes

PR #1290, opened 2022-02-12
Updated donor features

oof2win2:
Implements most parts of the image below, excluding biter pets and the sales in shop.
Image for reference

Todo:
- Figure out how to get the data to be recalculated when it gets changed in the database whilst a player is online


grilledham:
Thanks for the PR. After discussing it with the team we decided to go with #1262 instead.

PR #1294, opened 2022-03-10
Fixed error in multiline barrage command

Jayefuu:
- Reported by Clinkstur
- If the barrage antigrief found no nests in an area then it returned from the function
- This broke the command if the player used a multi-coordinate barrage to barrage more than one location
- New behaviour is that it just skips that location and continues to next coordinates
- I looked at how strike works and it seems it doesn't have any anti-grief so no changes to that one

### DIFF
--- a/map_gen/maps/crash_site/commands.lua
+++ b/map_gen/maps/crash_site/commands.lua
@@ -546,23 +546,22 @@ function Public.control(config)
             inv.remove({name = "explosive-rocket", count = strikeCost})
             if nest_count == 0 then
                 player.print({'command_description.crash_site_barrage_no_nests',}, Color.fail)
-                return
+            else
+
+                player.force.chart(s, {{xpos - 32, ypos - 32}, {xpos + 32, ypos + 32}})
+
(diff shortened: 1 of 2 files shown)


map_gen/maps/crash_site/commands.lua
```diff
@@ -546,23 +546,22 @@ function Public.control(config)
             inv.remove({name = "explosive-rocket", count = strikeCost})
             if nest_count == 0 then
                 player.print({'command_description.crash_site_barrage_no_nests',}, Color.fail)
```
grilledham: Now that this is specific to the coordinates rather than the whole command, I wonder if we should include which coordinates this failed for?

grilledham (commented):
This seems fine but I have a suggestion.

map_gen/maps/crash_site/commands.lua
Jayefuu: It's easy to add. Why not. I'll add that soon.

grilledham (approved):

PR #1295, opened 2022-03-11
A point-and-click method to call AOE functions

Jayefuu:
Allows players to point and click to call strike and barrage instead of using text commands.

https://streamable.com/4e2ed6

- Create a new decon planner
- Change its icon to the desired method
- When decon planner is used, check they only clicked not dragged an area
- Check what the icon is and call the respective function  from commands.lua

Known issues:
- If you put the decon plannerin the game blueprints then take it out again it  doesn't work. I get this error: LuaItemStack API call when LuaItemStack was invalid for read. It's definitely a Lua item stack but it's invalid for some reason. This isn't an issue with the new code to work but I know players will want to store the decon planner in game blueprints.


map_gen/maps/crash_site/features/discharge_targetting_remote.lua
```diff
@@ -0,0 +1,37 @@
...
+
+Event.add(defines.events. on_player_deconstructed_area , function(event)
+
+    local player = game.get_player(event.player_index)
+    if not player or not player.valid then
+        return
+    end
+
+    -- check they actually have a decon planner in their cursor
+    local item_name = player.cursor_stack.name
```
grilledham: `cursor_stack` can be `nil` I'd also add a valid check because why not.

map_gen/maps/crash_site/features/discharge_targetting_remote.lua
```diff
@@ -0,0 +1,37 @@
...
+    if item_name ~= "deconstruction-planner" then
+        return
+    end
+
+    -- check if the player has given the decon planner an icon. This is how  we will determine  their intention
+    if not player.cursor_stack.blueprint_icons[1].signal.name then
```
grilledham: `blueprint_icons` could return an empty array. So we either need to check the size of the array or check the [1] doesn't return `nil`

I'd probably make a `local icon = player.cursor_stack.blueprint_icons[1]`

map_gen/maps/crash_site/features/discharge_targetting_remote.lua
```diff
@@ -0,0 +1,37 @@
...
+        return
+    end
+
+    local icon_name = player.cursor_stack.blueprint_icons[1].signal.name
+
+    -- only continue if they do a small click. We don't want them selecting a huge area
+    if (math.abs(event.area.left_top.x -  event.area.right_bottom.x) < 1) and (math.abs(event.area.left_top.y -  event.area.right_bottom.y) < 1)  then
```
grilledham: I'd make locals for x and y.

map_gen/maps/crash_site/features/discharge_targetting_remote.lua
```diff
@@ -0,0 +1,37 @@
+local Event = require 'utils.event'
```
grilledham: Why is the file called "discharge_targetting_remote". I'm not sure what I'd call it but maybe just "targetting_remote"?

map_gen/maps/crash_site/commands.lua
```diff
@@ -570,6 +570,14 @@ function Public.control(config)
         end
     end
 
+    function Public.call_strike(args, player)
```
grilledham: In cases like this you can just do:
```suggestion
    Public.call_strike = strike
```

grilledham (changes requested):
We also discussed adding the deconstruction planners to the market. Feel free to add that as part of this PR or leave it for later.

map_gen/maps/crash_site/features/discharge_targetting_remote.lua
Jayefuu: I originally tried to use the capsule event to trigger stuff and was going to have a gui to switch between modes. I forgot to rename it.

map_gen/maps/crash_site/commands.lua
Jayefuu: Fixed

map_gen/maps/crash_site/features/discharge_targetting_remote.lua
Jayefuu: Fixed

Jayefuu:
A few changes made but market for decon planners a work in progress.

map_gen/maps/crash_site/commands.lua
```diff
@@ -630,8 +633,15 @@ function Public.control(config)
                 }
                 item.price = math.floor(math.exp(airstrike_data.radius_level ^ 0.8) / 2) * 1000
                 Retailer.set_item(market_id, item) -- this updates the retailer with the new item values.
+            elseif name == 'airstrike_planner' then
+                local player = event.player
+                player.cursor_stack.import_stack("0eNpdjUEKwjAURO8y6xRMtVZzFREJ6ad8SH9Ckhal5O42Czfu5j0ezI6JXJBc0uoKB3lFb0UowezIVArLnNvmFsE8DsuzWN9c+USCARdaoCB2aRQD5yCdszGvnlAVWCZ6w+j6bNBi83fa/U4VNkr5MDD9TV/Gez8OetDn66nWL23MOqk=")
+
```
grilledham: Couldn't this instead be
```lua
local cursor_stack = player.cursor_stack
cursor_stack.set_stack({name = 'deconstruction-planner'})
cursor_stack.blueprint_icons = {{index = 1, signal = {type = 'item', name = 'poison-capsule'}}}
```

Also I guess we need to check that the cursor stack is empty first.

Jayefuu:
@grilledham I think I've now addressed all your comments.

map_gen/maps/crash_site/features/deconstruction_targetting.lua
```diff
@@ -0,0 +1,39 @@
+local Event = require 'utils.event'
+local Commands = require 'map_gen.maps.crash_site.commands'
+
+Event.add(defines.events. on_player_deconstructed_area , function(event)
+
+    local player = game.get_player(event.player_index)
+    if not player or not player.valid or not player.cursor_stack or not player.cursor_stack.valid then
```
grilledham: LuaItemStacks are weird I think we also have to check [valid_for_read](https://lua-api.factorio.com/latest/LuaItemStack.html#LuaItemStack.valid_for_read)

I'd probably also make a local for cursor_stack.

map_gen/maps/crash_site/features/deconstruction_targetting.lua
```diff
@@ -0,0 +1,39 @@
...
+    end
+
+    -- check they actually have a decon planner in their cursor
+    local item_name = player.cursor_stack.name
+    if item_name ~= "deconstruction-planner" then
+        return
+    end
+
+    -- check if the player has given the decon planner an icon. This is how  we will determine  their intention
+    if not player.cursor_stack.blueprint_icons or not player.cursor_stack.blueprint_icons[1].signal.name then
```
grilledham: `blueprint_icons[1]` could be `nil`

map_gen/maps/crash_site/scenario.lua
```diff
@@ -770,6 +771,15 @@ local function init(config)
             upgrade_rate = 0.5,
             upgrade_base_cost = 500,
             upgrade_cost_base = 2,
+            {
+                price = 5,
```
grilledham: Why is there a price for items players can make for free?

grilledham (changes requested):
I still need to give this a test.

Just a though, we might need to change the antigrief code (in nuke_control.lua) that removes deconstruction planners from guests, else they will have a bad time using this.

map_gen/maps/crash_site/scenario.lua
Jayefuu: I don't know. I thought it might prevent some spamming of them and people dropping them in chests. But since it goes to cursor now not into inventory that's less likely.

PR #1296, opened 2022-03-11
Moved box position again

Jayefuu:
As requested by players after feedback from first change.

image


Jayefuu:
@grilledham if it gets straight approved again, please merge.

grilledham (approved):

PR #1297, opened 2022-03-12
Introduce unique prefix for Redmew GUI elements

Soggs:
Due to collisions with mods when naming GUI elements with just numbers, I added a prefix when naming elements with ``utils.token.uid()``.

### DIFF
--- a/utils/gui.lua
+++ b/utils/gui.lua
@@ -324,12 +324,12 @@ if _DEBUG then
         local filepath = info.source:match('^.+/currently%-playing/(.+)$'):sub(1, -5)
         local line = info.currentline
 
-        local token = tostring(Token.uid())
+        local token = gui_element_prefix .. tostring(Token.uid())
 
         local name = concat {token, ' - ', filepath, ':line:', line}
         names[token] = name
 
-        return gui_element_prefix .. token
+        return token
     end
 
     function Gui.set_data(element, value)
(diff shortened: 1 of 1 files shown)


utils/gui.lua
```diff
@@ -6,6 +6,8 @@ local Styles = require 'resources.styles'
 local tostring = tostring
 local next = next
 
+local gui_element_prefix = "Redmew_GUI_element_"
```
grilledham: Maybe we could go shorter like just `"Redmew_"`

grilledham (commented):
Have you tested this fixes the problem you were having when running mods?

utils/gui.lua
Soggs: We can definitely shorten it. I wanted to be more verbose rather than less to start out.

Soggs:
Yeah I tested it. It fixes the problems with Module Inserter. I am still uncertain however why it fired in the first place and also why it only fires both with the MI buttons but not with the Redmew GUI button.

I am at the same time investigating how I could change the MI buttons but the GUI there uses the ``name`` strings as numeric indices in a few places, which I hope we do not do anywhere. Though I think you should never do something like that in the first place and other elements in our GUI do use strings and not numbers like the diggy exp button is internally called ``Diggy.Experience.Button``.

PR #1303, opened 2022-03-17
Changes to spy reduce chat spam and improve usage visibility on map

Jayefuu:
- Added 30 second cooldown to spy chat message
- Still shows message on first use so that new players know about the command
- Added spy fish tag to map so that players can see easily which are the newly revealed areas.

### DIFF
--- a/map_gen/maps/crash_site/commands.lua
+++ b/map_gen/maps/crash_site/commands.lua
@@ -29,7 +29,7 @@ function Public.control(config)
 
     local airstrike_data = {radius_level = 1, count_level = 1}
     local barrage_data = {radius_level = 1, count_level = 1}
-    local spy_message_cooldown = false
+    local spy_message_cooldown = {false}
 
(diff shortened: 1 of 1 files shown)


map_gen/maps/crash_site/commands.lua
```diff
@@ -29,10 +29,12 @@ function Public.control(config)
 
     local airstrike_data = {radius_level = 1, count_level = 1}
     local barrage_data = {radius_level = 1, count_level = 1}
+    local spy_message_cooldown = false
```
grilledham: This is desync risk. In lua Booleans are value types so this is a copy of the value stored in global. The trick is to store the value inside a table which are reference types e.g.
```suggestion
    local spy_message_cooldown = {false}
```
And then the places where this is set would be `spy_message_cooldown[1] = true/false`

grilledham (changes requested):

map_gen/maps/crash_site/commands.lua
Jayefuu: Cool. Thanks

grilledham (approved):

PR #1304, opened 2022-03-18
A deconstruction planner control to easily group spiders to follow a …

Jayefuu:
…leader

Quicker and better than using the vanilla spidertron remote behaviour.


features/spidertron_group_control.lua
```diff
@@ -0,0 +1,114 @@
...
+        if #target_spider == 1 then
+            for i, spidertron in pairs(spidertrons) do
+                spidertron.follow_target = target_spider[1]
+            end
+            spider_army[player.name] = {} -- clear spidertrons from table once they've been assigned to follow another spidey lad
+            cursor_stack.label = 'Select a group of your spidertrons! 0 selected.'
+        else
+            player.surface.create_entity {
+                name = 'flying-text',
+                text = '0 [img=item.spidertron] found',
```
grilledham: Use localised string. Also maybe "No [img=item.spidertron] found" instead of 0.

features/spidertron_group_control.lua
```diff
@@ -0,0 +1,114 @@
...
+    end
+
+    local left_top = event.area.left_top
+    local right_bottom = event.area.right_bottom
+    local width = math.abs(left_top.x -  right_bottom.x)
+    local height = math.abs(left_top.y -  right_bottom.y)
+
+    -- A single (small area) click is interpretted as instruction to assign a group of spiders to follow a spider
+    if width <= 1 and height <= 1  then
+        local spidertrons = spider_army[player.name]
```
grilledham, line 64: `spidertrons` could be `nil` here. Do we want to detect that and print a helpful message, like select the the spiders to follow first?

features/spidertron_group_control.lua
```diff
@@ -0,0 +1,114 @@
...
+                position = left_top,
+                render_player_index = player.index
+            }
+        end
+    else -- else the area is bigger than 1x1 and so we assume the player is selecting which spiders to assign
+        local spidertrons = player.surface.find_entities_filtered{area = {left_top, right_bottom}, name="spidertron"}
+        if #spidertrons > 0 then
+            for i, spidertron in pairs(spidertrons) do
+                if spidertron.last_user.name ~= player.name then
+                    spidertrons[i] = {} -- remove a spidertron from the table of found entities if the player isn't the last user. Anti-grief.
```
grilledham: Might make more sense to build a new array of `allowed_spidertrons` instead. Then your count below wouldn't be wrong

features/spidertron_group_control.lua
```diff
@@ -0,0 +1,114 @@
...
+
+            spider_army[player.name] = spidertrons
+            cursor_stack.label = #spidertrons..' selected. Click a spidertron for them to follow.'
+        else
+            cursor_stack.label = 'Select a group of your spidertrons! 0 selected.'
+        end
+        -- Flying text to appear at top left of selection area showing player how many spidertrons they selected
+        player.surface.create_entity {
+            name = 'flying-text',
+            text = '+'..#spidertrons..' [img=item.spidertron]',
```
grilledham: Use localised string

grilledham (changes requested):
I haven't play tested it yet, but I looked through the code and have some suggestions.

features/spidertron_group_control.lua
Jayefuu: This one doesn't need a localised string. It's an icon and a number.

Jayefuu: Oh you meant the lines above. Got it

grilledham: I did mean this line. I was actually just saying it because the type supports localised strings. But what if in a different culture '+' should be something else.

features/spidertron_group_control.lua
```diff
@@ -0,0 +1,116 @@
...
+                render_player_index = player.index
+            }
+        end
+    else -- else the area is bigger than 1x1 and so we assume the player is selecting which spiders to assign
+        local spidertrons = player.surface.find_entities_filtered{area = {left_top, right_bottom}, name="spidertron"}
+        local spidertrons_valid = {}
+        if #spidertrons > 0 then
+            for i, spidertron in pairs(spidertrons) do
+                if spidertron.last_user.name == player.name then
+                    spidertrons_valid[i] = spidertron
```
grilledham:
```suggestion
                    spidertrons_valid[#spidertrons_valid + 1] = spidertron
```
We can't use `i` because it could create gaps in the `spidertrons_valid` array which would cause the count to be incorrect.

features/spidertron_group_control.lua
```diff
@@ -0,0 +1,116 @@
...
+                        time_to_live = 60*4,
+                        players = {player.name}
+                    }
+                end
+            end
+
+            spider_army[player.name] = spidertrons_valid
+            --cursor_stack.label = {#spidertrons..' selected. Click a spidertron for them to follow.'}
+            cursor_stack.label = #spidertrons_valid..' selected. Click a spidertron for them to follow.'
+            --cursor_stack.label = {{'spidertron_group_control.spiders_selected_success',7}}
```
grilledham: Remove the commented out code

grilledham (changes requested):

features/spidertron_group_control.lua
Jayefuu: So if I do:
local test = {}
test[1] = 7
test[709] = 7

Then #test is 709 ?

But if I do 
local test = {}
test["one"] = 7
test["709"] = 7

Then #test is 2?

*sigh* I hate lua

grilledham: > So if I do:
local test = {}
test[1] = 7
test[709] = 7
Then #test is 709 ?

`#test` would be 1

> But if I do
local test = {}
test["one"] = 7
test["709"] = 7
Then #test is 2?

#test would be 0.

`#` is the number of consecutive entries from 1 .. n.
0, negative numbers, string and other keys are ignored.

features/spidertron_group_control.lua
```diff
@@ -0,0 +1,123 @@
...
+                position = left_top,
+                render_player_index = player.index
+            }
+            return
+        end
+        local spidertrons = spider_army[player.name]
+        local target_spider = player.surface.find_entities_filtered{name="spidertron", area={{left_top.x-1,left_top.y-1}, {right_bottom.x+1,right_bottom.y+1}}, limit=1}
+        if #target_spider == 1 then
+            for i, spidertron in pairs(spidertrons) do
+                spidertron.follow_target = target_spider[1]
```
grilledham: Just noticed we need to check `spidertron` is valid here. E.g. you select a spidertron it is then mined or dies, then you try and target another spidertron.
image

features/spidertron_group_control.lua
```diff
@@ -0,0 +1,123 @@
...
+        if not spider_army[player.name] then
+            player.surface.create_entity {
+                name = 'flying-text',
+                text = {'spidertron_group_control.none_selected'},
+                position = left_top,
+                render_player_index = player.index
+            }
+            return
+        end
+        local spidertrons = spider_army[player.name]
```
grilledham: You should move this line `local spidertrons = spider_army[player.name]` above
`if not spider_army[player.name] then` so you can use it there, instead of doing the same table lookup twice.

map_gen/maps/crash_site/commands.lua
```diff
@@ -702,6 +702,18 @@ function Public.control(config)
                 cursor_stack.entity_filters = {'sand-rock-big'}
             end
         end
+        if item.type == 'spidertron' then
+            if item.name=='spidertron_planner' then
```
grilledham: I'd combine this with the previous if.

grilledham (changes requested):

features/spidertron_group_control.lua
Jayefuu: Good to know. Thank you.

grilledham (approved):

PR #1305, opened 2022-03-21
Game statistics and player awards for danger ores.

Jayefuu:


features/player_stats.lua
```diff
@@ -51,6 +54,9 @@ ScoreTracker.register(player_turrets_killed_name, {'player_stats.player_turrets_
 ScoreTracker.register(player_total_kills_name, {'player_stats.player_total_kills'})
 ScoreTracker.register(player_entities_built_name, {'player_stats.player_entities_built'})
 ScoreTracker.register(player_fish_eaten_name, {'player_stats.player_fish_eaten'})
+ScoreTracker.register(player_resources_hand_mined_name, {'player_stats.player_resources_hand_mined'})
+ScoreTracker.register(resources_hand_mined_name, {'player_stats.resources_hand_mined'})
+ScoreTracker.register(resource_tiles_exhausted_name, {'player_stats.resource_tiles_exhausted'})
```
grilledham: Maybe `resources_exhausted_name` instead. Tiles is slightly odd because a resource could take up more than one tile.

map_gen/maps/danger_ores/modules/restart_command.lua
```diff
@@ -29,6 +34,140 @@ return function(config)
...
+        end
+        ore_totals_message = ore_totals_message:sub(1, -3)..')'
+        ore_totals_message = "Total ore mined: "..string.format("%.2fM",total_ore/1000000).. " "..ore_totals_message
+
+        Server.to_discord_named_embed(map_promotion_channel, 'Danger ores map completed!\\n\\n'
+        .. 'Statistics:\\n'
+        .. 'Map time: '..time_string..'\\n'
+        .. 'Total entities built: '..statistics.entities_built..'\\n'
+        .. 'Total ore mined:'..ore_totals_message..'\\n'
+        .. 'Total ore tiles exhausted: '..statistics.resource_tiles_exhausted..'\\n'
```
grilledham: Maybe `'Total resources exhausted'`. This would also include non ore resources that are exhausted like oil.

map_gen/maps/danger_ores/modules/restart_command.lua
```diff
@@ -29,6 +34,140 @@ return function(config)
...
+            end
+        end
+
+        local time_string = Core.format_time(game.ticks_played)
+
+        local resource_prototypes = game.get_filtered_entity_prototypes({{filter = "type", type = "resource"}})
+        local total_ore = 0
+        local ore_totals_message = '('
+        for ore_name,_ in pairs(resource_prototypes) do
+            if game.entity_prototypes[ore_name].resource_category == 'basic-solid' then
```
grilledham: Is `game.entity_prototypes[ore_name]` not just the variable you are discarding in the for loop above?

map_gen/maps/danger_ores/modules/restart_command.lua
```diff
@@ -29,6 +34,140 @@ return function(config)
...
+                local count = game.forces["player"].item_production_statistics.get_input_count(ore_name)
+                total_ore = total_ore + count
+                ore_totals_message = ore_totals_message..ore_name:gsub( "-ore", "")..": "..string.format("%.2fM, ",count/1000000)
```
grilledham: You could use `format_number` here instead, handles the suffixes for you. See <https://github.com/Refactorio/RedMew/blob/develop/features/gui/score.lua#L9> for an example.

map_gen/maps/danger_ores/modules/restart_command.lua
```diff
@@ -29,6 +34,140 @@ return function(config)
...
+            end
+        end
+        ore_totals_message = ore_totals_message:sub(1, -3)..')'
```
grilledham: Maybe it bothers me more than it should given this is a danger ore map, but in theory the number of ore prototypes could be zero, which would make this message odd. Feel free to ignore.

grilledham (changes requested):
A few suggestions. I haven't play tested it but it mostly looks good.

map_gen/maps/danger_ores/modules/restart_command.lua
Jayefuu: I chose to exclude oil because it's never really exhausted, just reaches a minimum output. I want this to be more a figure of how much space has been cleared.

map_gen/maps/danger_ores/modules/restart_command.lua
Jayefuu: Will do :D

features/player_stats.lua
Jayefuu: Good point. Fixed.

map_gen/maps/danger_ores/modules/restart_command.lua
Jayefuu: I looked at the documentation for Util.format_num = function(amount, decimal, prefix, neg_prefix). It doesn't look as clear to use to me as the C style string.format which is pretty universal. What I've done works and I think is clearer.

map_gen/maps/danger_ores/modules/restart_command.lua
Jayefuu: Doh. Thank you. Fixed.

map_gen/maps/danger_ores/modules/restart_command.lua
grilledham: Wouldn't `string.format("%.2fM, ",count/1000000)` just become `format_number(count, true)`.
The advantage of format_number is it picks the right suffix based on the magnitude.

Not sure where you found that documentation `format_number` is `function util.format_number(amount, append_suffix)`

Jayefuu: Oh! That's cool.

grilledham (approved):

PR #1307, opened 2022-03-25
Add no_bots preset for crashsite

Soggs:
The preset removes access to construction and logistic bots on crashsite


map_gen/maps/crash_site/scenario_no_bots.lua
```diff
@@ -0,0 +1,954 @@
...
+            loot_box.insert {name = 'iron-plate', count = 50}
+        end
+    end
+    cutscene_builder('market', -4, 0)
+    local furnace = cutscene_builder('electric-furnace', 4, 0)
+    furnace.insert('iron-ore')
+    game.surfaces.cutscene.set_tiles(tiles)
+end
+
+local function init(config)
```
grilledham: config could have a flag for no_bot_outposts. Then we just don't add `mini_t1_robotics_factory` to `mini2` and `mini3` tables. The order in the tables doesn't matter, so it could be the last entry.

Is that the only different in this file compared to scenario.lua, would that allow us to avoid the duplication?

map_gen/maps/crash_site/presets/no_bots.lua
```diff
@@ -0,0 +1,53 @@
+local MGSP = require 'resources.map_gen_settings'
+local ScenarioInfo = require 'features.gui.info'
+local Event = require 'utils.event'
+
+local config = {
+    scenario_name = 'crashsite',
```
grilledham: We should also add the preset to scenario_templates. Whatever you call that preset, probably "crashsite-no-bots", should be the name here.

map_gen/maps/crash_site/presets/no_bots.lua
```diff
@@ -0,0 +1,53 @@
...
+        },
+        MGSP.starting_area_very_low,
+        MGSP.ore_oil_none,
+        MGSP.enemy_none,
+        MGSP.cliff_none
+    }
+}
+
+local Scenario = require 'map_gen.maps.crash_site.scenario_no_bots'
+ScenarioInfo.set_map_name('Crashsite')
```
grilledham: Change the name here too.

grilledham (changes requested):

map_gen/maps/crash_site/scenario_no_bots.lua
Soggs: So had a moment to think and I remembered when going over it, that it might be an idea to generally move the list of outposts that get loaded into the config. So we are generally able to alter outposts based on config. Partially because it might interesting to have different levels of allowed bots. This current preset is literally no bots but maybe we only want to prohibit logistic bots.

Also would leave for presets like no science outposts or no military outposts.

grilledham: Sounds like a good idea, more effort though. I'll leave it up to you if you want to do that.

Soggs:
I think I have everything updated. In the mean time I did add a second preset, that is just personal construction bots instead of no bots at all.

map_gen/maps/crash_site/scenario_no_bots.lua
grilledham: I would prefer not to have a duplicate scenario file. I think adding a flag to config and not adding the bot outposts would be easy to do.

grilledham (approved):

PR #1311, opened 2022-04-12
Duplicated end of game awards messages to respective channels

Jayefuu:
- The awards being in the map-promotion channel is great because people see it when they're pinged and there being no chatter in there keeps it clean and easy to find/read.
- However, it doesn't foster community and discussion.
- This PR duplicates the embedded message so it's posted once to map-promotion and once to the respective channels for danger ores and crash site.
- Will give players somewhere to discuss post-match. I hope it will lead to more discussion of game play and some more friendly competition.

### DIFF
--- a/map_gen/maps/crash_site/commands.lua
+++ b/map_gen/maps/crash_site/commands.lua
@@ -213,10 +213,10 @@ function Public.control(config)
                 awards.fish_eaten.player = v.name
             end
         end
-
+        local statistics_message = {}
         local time_string = Core.format_time(game.ticks_played)
         if statistics.enemy_entities < 1000 then
-            local statistics_message = 'Crash Site map won!\\n\\n'
+            statistics_message = 'Crash Site map won!\\n\\n'
             .. 'Statistics:\\n'
             .. 'Map time: '..time_string..'\\n'
             .. 'Total kills: '..statistics.biters_killed..'\\n'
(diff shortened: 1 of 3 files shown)


map_gen/maps/crash_site/commands.lua
```diff
@@ -214,7 +216,7 @@ function Public.control(config)
 
         local time_string = Core.format_time(game.ticks_played)
         if statistics.enemy_entities < 1000 then
-            Server.to_discord_named_embed(map_promotion_channel, 'Crash Site map won!\\n\\n'
+            local statistics_message = 'Crash Site map won!\\n\\n'
```
grilledham: If you declare `local statistics_message` above the `if` then you can move the `Server.to_discord_named_embed` below the `end` and remove the duplication.

grilledham (changes requested):

grilledham (approved):

map_gen/maps/crash_site/commands.lua
```diff
@@ -213,10 +213,10 @@ function Public.control(config)
                 awards.fish_eaten.player = v.name
             end
         end
-
+        local statistics_message = {}
```
grilledham: Lua check is complaining about this
```suggestion
        local statistics_message
```

You don't need to assign it because it's always assigned before use.

grilledham (changes requested):

grilledham (approved):

PR #1312, opened 2022-04-12
Rewrite of crash site awards to include top 3 in each category.

Jayefuu:
I'll do danger ores and diggy at a later date a) when I have time and b) once I see how this goes down.

### DIFF
--- a/map_gen/maps/crash_site/commands.lua
+++ b/map_gen/maps/crash_site/commands.lua
@@ -14,6 +14,7 @@ local ScoreTracker = require 'utils.score_tracker'
 local PlayerStats = require 'features.player_stats'
 local Restart = require 'features.restart_command'
 local set_timeout_in_ticks = Task.set_timeout_in_ticks
+local format_number = require 'util'.format_number
 
 -- Use these settings for live
 local map_promotion_channel = Discord.channel_names.map_promotion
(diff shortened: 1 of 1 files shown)


map_gen/maps/crash_site/commands.lua
```diff
@@ -114,6 +117,28 @@ function Public.control(config)
...
     end
 
+    local places = {'1st','2nd','3rd','4th','5th'}
+
+    -- builds a message for top 3 or 4 etc players for a given stat to use in the embedded awards
+    local function get_top_players(players_as_array, statistic_name)
+        local return_message = ""
+        table.sort(players_as_array, function (a, b) return a[statistic_name] > b[statistic_name] end )
+        for i=1,3 do -- currently only showing top 3 but might want to show more later?
+            if players_as_array[i] then -- for the unlikely event that only 1 or 2 players joined the map
```
grilledham: Alternatively you could do `for i = 1, math.min(3, #players_as_array) do` above, instead of the if check.

map_gen/maps/crash_site/commands.lua
```diff
@@ -151,97 +176,46 @@ function Public.control(config)
...
-            if v.coins_earned > awards.coins_earned.value then
-                awards.coins_earned.value = v.coins_earned
-                awards.coins_earned.player = v.name
-            end
-            if v.fish_eaten > awards.fish_eaten.value then
-                awards.fish_eaten.value = v.fish_eaten
-                awards.fish_eaten.player = v.name
-            end
+        local players_as_array = {}
+        for name, data in pairs(statistics.player_data) do
```
```suggestion
        for name, data in pairs(player_data) do
```

grilledham (changes requested):

map_gen/maps/crash_site/commands.lua
Jayefuu: Clever. I like.

grilledham (approved):

PR #1320, opened 2022-05-15
Add wooden-chest 4 BP digging to allowed and change ore generation back to tendrils_impure

FuzTheCat:


map_gen/maps/diggy/presets/danger_ores.lua
```diff
@@ -471,7 +471,8 @@ local config = {
 				'stone-wall',
 				'small-lamp',
 				'pipe',
-				'pipe-to-ground'
+				'pipe-to-ground',
+				'wooden-chest'
```
grilledham, line 475: Curious, why wooden chests on ore?

FuzTheCat: We use wooden-chest ghosts for mining rocks since they are generally not made/used anywhere, they will never be placed. BPs for mining usually use lines of wooden-chest between the walls leaving gaps of supporting rocks (not fully filled in with wooden-chest which generally has a higher rate of collapses as the rocks get mined often B4 the walls are placed.) It's just a nuisance having chat fully fill with the message about not being able to place the wooden-chest on ores.

grilledham (approved):
Thanks for the PR.

PR #1321, opened 2022-05-20
Add map name to stats

Duko:
Added map name to the title of the Discord embed

### DIFF
--- a/map_gen/maps/danger_ores/modules/restart_command.lua
+++ b/map_gen/maps/danger_ores/modules/restart_command.lua
@@ -62,7 +62,7 @@ return function(config)
         end
 
         local statistics = {
-            scenario = ScenarioInfo.get_map_name(),
+            scenario = config.scenario_name or 'Danger ore',
             start_epoch = Server.get_start_time(),
             end_epoch = end_epoch, -- stored as key already, useful to have it as part of same structure
             game_ticks = game.ticks_played,
--- a/map_gen/maps/danger_ores/modules/restart_command.lua
+++ b/map_gen/maps/danger_ores/modules/restart_command.lua
@@ -6,6 +6,7 @@ local ShareGlobals = require 'map_gen.maps.danger_ores.modules.shared_globals'
 local ScoreTracker = require 'utils.score_tracker'
 local PlayerStats = require 'features.player_stats'
 local format_number = require 'util'.format_number
+local ScenarioInfo = require 'features.gui.info'
 
 return function(config)
     local map_promotion_channel = Discord.channel_names.map_promotion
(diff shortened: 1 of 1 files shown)


map_gen/maps/danger_ores/modules/restart_command.lua
```diff
@@ -61,7 +62,7 @@ return function(config)
         end
 
         local statistics = {
-            scenario = config.scenario_name,
+            scenario = ScenarioInfo.get_map_name(),
```
grilledham:
```suggestion
            scenario = config.scenario_name or 'Danger ore',
```
It'd be consistent with the names that get used for the next map.

grilledham (changes requested):
Thanks for the PR, could I trouble you for a change?

map_gen/maps/danger_ores/modules/restart_command.lua
Duko: Wasn't sure if `statistics.scenario_name` was used.
isn't `config.scenario_name` the next map?
and `ScenarioInfo.get_map_name()` is formatted for display, so it's looks better in discord.

> Danger Ore Gradient

instead of

> danger_ore_gradient

could add both to the statistics array if needed when sent to the server

grilledham: > Wasn't sure if `statistics.scenario_name` was used.

There is a key/value store on the server, which is what `Server.set_data('danger_ores_data', tostring(end_epoch), statistics)` is doing. The data isn't needed, but is just collected if anyone (Jayefuu) what's to look at them.

> isn't config.scenario_name the next map?

It's the map that is set in the presets e.g. here <https://github.com/Refactorio/RedMew/blob/develop/map_gen/maps/danger_ores/presets/danger_ore_deadlock_beltboxes_ore_only.lua#L119>. So assuming the presets are correct it would be the current map. By default the maps restart to themselves unless an admin changes it with `/config-restart`. 
```lua
        local start_game_data = Restart.get_start_game_data()
        local new_map_name = start_game_data.name
```
On line 36 determines the next map.

> and ScenarioInfo.get_map_name() is formatted for display, so it's looks better in discord.

There are pros and cons to using `ScenarioInfo.get_map_name()`. My intital thought is it would be inconsistent with the name of the map we show for the next map in map-promotion. e.g:
image. Technically admins can also change what `ScenarioInfo.get_map_name()` returns by editing the textbox in the map info gui (for admins the fields are editable).

> could add both to the statistics array if needed when sent to the server

I'd probably stick with sending `config.scenario_name` as it's what was being sent before.

Duko: Should I move it out of the title then? Doesn't look good in title when it's dashed.

image
image
image

grilledham: I think it's fine in the title.

grilledham (approved):

PR #1322, opened 2022-05-29
Change on_destroy message on danger_ores

constantoine:
Before, the message was hardcoded. Now, the message is generated from the list
of entities actually whitelisted, and prints the list of icons

It now looks like this on `map_gen.maps.danger_ores.presets.danger_ore_gradient`
image

### DIFF
--- a/map_gen/maps/danger_ores/modules/banned_entities.lua
+++ b/map_gen/maps/danger_ores/modules/banned_entities.lua
@@ -29,7 +29,35 @@ return function(allowed_entities, message)
     local function on_destroy(event)
         local p = event.player
         if p and p.valid then
-            p.print(message or 'You cannot build that on top of ores, only belts, mining drills, and power poles are allowed.')
+            if message then
+                p.print(message)
+                return
+            end
+            local items = {}
+            local len = 0
+            for _, v in pairs(allowed_entities) do
+                local entity = game.entity_prototypes[v]
+                for _, v in pairs(entity.items_to_place_this) do
+                    if not (items[v.name] ~= nil) then --- Avoid duplication for straight-rail and curved-rail, which both use rail
+                        items[v.name] = v
+                        len = len + 1
+                    end
+                end
+            end
+            local i = 1
+            local str = "You cannot build that on top of ores, only "
+            for k, v in pairs(items) do
+                str = str.."[img=item." .. k .."]"
+                if len == i then
+                    str = str.."."
+                elseif i == len - 1 then
+                    str = str.." and "
+                else
+                    str = str..", "
+                end
+                i = i + 1
+            end
+            p.print(str)
         end
     end
 
--- a/map_gen/maps/danger_ores/modules/banned_entities.lua
+++ b/map_gen/maps/danger_ores/modules/banned_entities.lua
@@ -35,8 +35,9 @@ return function(allowed_entities, message)
             end
             local items = {}
             local len = 0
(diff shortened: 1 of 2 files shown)


map_gen/maps/danger_ores/modules/banned_entities.lua
```diff
@@ -29,7 +29,35 @@ return function(allowed_entities, message)
...
         local p = event.player
         if p and p.valid then
-            p.print(message or 'You cannot build that on top of ores, only belts, mining drills, and power poles are allowed.')
+            if message then
+                p.print(message)
+                return
+            end
+            local items = {}
+            local len = 0
+            for _, v in pairs(allowed_entities) do
```
grilledham: `allowed_entities` is the entities that were allowed at the start of the map. This could change if an admin adds or removes entities from the list. You would need to get `allowed_entities` from `RestrictEntities` to get the up to date list. You will probably need to add a public getter function.

map_gen/maps/danger_ores/modules/banned_entities.lua
```diff
@@ -29,7 +29,35 @@ return function(allowed_entities, message)
...
+                local entity = game.entity_prototypes[v]
+                for _, v in pairs(entity.items_to_place_this) do
+                    if not (items[v.name] ~= nil) then --- Avoid duplication for straight-rail and curved-rail, which both use rail
```
grilledham:
```suggestion
                    if not items[v.name] then --- Avoid duplication for straight-rail and curved-rail, which both use rail
```

map_gen/maps/danger_ores/modules/banned_entities.lua
```diff
@@ -29,7 +29,35 @@ return function(allowed_entities, message)
...
+            local i = 1
+            local str = "You cannot build that on top of ores, only "
+            for k, v in pairs(items) do
+                str = str.."[img=item." .. k .."]"
+                if len == i then
+                    str = str.."."
+                elseif i == len - 1 then
+                    str = str.." and "
+                else
+                    str = str..", "
```
grilledham: I probably wouldn't separate a list of images with `,` and `and` I think it would look better without. And it would simplify the code slightly. 

Also it might be worth considering using `table.concat(array)` to reduce string allocations.

grilledham (commented):
Thanks for the PR. A few suggestions.

map_gen/maps/danger_ores/modules/banned_entities.lua
constantoine: Yes, this is the remnants of first iterations where I printed the names of the entities. Will remove

map_gen/maps/danger_ores/modules/banned_entities.lua
constantoine: In the code I can see https://github.com/Refactorio/RedMew/blob/develop/map_gen/shared/entity_placement_restriction.lua#L314-L321

Is the global allowed_entities a copy of allowed_entities, or do they share the same reference?

grilledham: > Is the global allowed_entities a copy of allowed_entities, or do they share the same reference?

The global allowed_entities is a copy.

constantoine: Should be good now

grilledham (approved):
LGTM! Thanks for the PR.

PR #1327, opened 2022-06-29
Updated commands.lua

Soggs:
Changed the regex on the spy, strike and barrage command to properly process decimals.

### DIFF
--- a/map_gen/maps/crash_site/commands.lua
+++ b/map_gen/maps/crash_site/commands.lua
@@ -270,7 +270,7 @@ function Public.control(config)
         local coords = {}
         local spy_cost = 100
 
-        for m in string.gmatch(location_string, "(%d*%.?%d+)") do -- Assuming the surface name isn't a valid number.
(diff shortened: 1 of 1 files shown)


map_gen/maps/crash_site/commands.lua
```diff
@@ -270,7 +270,7 @@ function Public.control(config)
         local coords = {}
         local spy_cost = 100
 
-        for m in string.gmatch(location_string, "%-?%d+") do -- Assuming the surface name isn't a valid number.
+        for m in string.gmatch(location_string, "(%d*%.?%d+)") do -- Assuming the surface name isn't a valid number.
```
grilledham: Doesn't match on negative numbers would need to be something like `"(%-?%d*%.?%d+)"`

grilledham (changes requested):

map_gen/maps/crash_site/commands.lua
Soggs: Resolved with 3189f70

grilledham (approved):

PR #1331, opened 2022-07-31
Terra latest update

Tarsassum:
After playtesting with the community in 2019, changes were made but never updated to the Refactorio branch. So here is the PR with the latest file per 2022.


grilledham:
@Tarsassum is there a reason you closed this? Is it not ready for review?

Tarsassum:
Another patch is incoming, i succesfully changed the ore generation, and currently am working on a bug with the bugs.

PR #1332, opened 2022-08-08
feat: Add blackfroest scenario

Bienenstock:
I hope there are no bugs left, I created this scenario a long time ago. 
Have Fun!


locale/en/redmew_maps.cfg
```diff
@@ -177,3 +177,34 @@ biters_disabled=Launching the first [item=satellite] has killed all the biters.
...
+gui_buff_mining=+__1__% mining speed (up to: __2__%)
+gui_buff_inv=+__1__ inventory slot(s) (up to: __2__)
+gui_buff_health=+__1__ max health (up to: __2__)
+gui_buff_other=+__1__ __2__
+gui_experience_button_tip=Black Forest leveling progress
+gui_close_btn=Close
+toast_new_level=Your team has reached level __1__!
+score_cave_collapses=Cave collapses
+score_mine_size=Mine size
+score_experience_lost=Experience lost
```
grilledham: These 3 are missing from the German translation. Any reason for that?

grilledham (commented):
Cool, thanks for making the PR.

I had a quick look and this mostly seems fine.

One thing I did notice is if I start the map with `_DEBUG = true` I get this error.
image

locale/en/redmew_maps.cfg
Bienenstock: ups, idk, I created it a long time ago. i don't think they are used. I will check tomorrow

Bienenstock:
> Cool, thanks for making the PR.
> 
> I had a quick look and this mostly seems fine.
> 
> One thing I did notice is if I start the map with `_DEBUG = true` I get this error. image

I have a look tomorrow

Bienenstock:
> Cool, thanks for making the PR.
> 
> I had a quick look and this mostly seems fine.
> 
> One thing I did notice is if I start the map with `_DEBUG = true` I get this error. 

fixed

locale/en/redmew_maps.cfg
Bienenstock: fixed

map_gen/maps/black_forest/feature/alien_spawner.lua
```diff
@@ -0,0 +1,208 @@
...
+            particle_count = 15
+        elseif rock_count == 2 then
+            particle_count = 10
+        else
+            particle_count = 5
+        end
+
+        for rock_index = rock_count, 1, -1 do
+            local rock = rocks[rock_index]
+            raise_event(defines.events.on_entity_died, {entity = rock})
```
grilledham: I ran into an issue with trying to raise the `on_entity_died` event. Diggy uses `defines.events.script_raised_destroy` here instead.

grilledham (commented):
There are some warnings in the [luacheck](https://github.com/Refactorio/RedMew/runs/8070879689?check_suite_focus=true). The "cyclomatic complexity of function on_tick is too high" might be hard to fix but the others are probably worth looking at. I can bypass the checks if we think it's not worth the effort but probably worth a look first.

jacobwatkinsgit:
I have a bunch of patches but I can't commit them because git wants write permissions to someone elses repository.

@Bienenstock Here, before I throw my computer out the window:

Summary:
add destroy_tree to create_particles to make leaf-particles, the way destroy_rock creates stone-particles, fix function description (create_particle, not create_entity)

refactor black_forest_grow: iterates every tree on the map, doesn't error on invalid entities any more, checks every tree and reduces redundant tree checks

general code quality cleaning, like using entity.destroy{raise_destroy = true}, instead of manual script raised destroy + an empty destroy().


[changed_files.zip](https://github.com/Refactorio/RedMew/files/9457018/changed_files.zip)

Bienenstock:
> refactor black_forest_grow: iterates every tree on the map, doesn't error on invalid entities any more, checks every tree and reduces redundant tree checks

@jacobwatkinsgit did you test your refac on black_forest_grow? the only trees that grow back in your version are the one at the starting area. expand into the void, trees don't grow back then on your refac

PR #1333, opened 2022-08-09
Terra, the resource update 2022

Tarsassum:
After play testing the map with the community in 2019, several changes were made, but never submitted to the refactorio branch.
Among other changes, biters are now wiped from the robot islands on map generation, however remnant biters still make the situation harder then intended. 
So although the map is still not in its final shape, now the resource update is included i will not hold back this update any longer.
Compared with the 2019 file, i am happy to present the current state of the map.


map_gen/terra.lua
```diff
@@ -0,0 +1,382 @@
+--Terra by T-A-R 2022
```
grilledham, line 1: Is this file just a duplicate of `map_gen/maps/terra.lua`? In which it can be removed.

Tarsassum: Not sure if i understand correctly, this file (terra.lua, Terra by TAR 2022) should replace the original terra.lua (from 2019) in map_gen/maps. That one has become a depricated version of the map, which can be removed.

grilledham: In this PR the two files have the same content (expect some minor whitespace differences). So I assume we only need the one in `map_gen/maps/`. Either way which ever version is the latest should be in `map_gen/maps/` and the file in `map_gen/` can be removed.

Tarsassum: I see what you mean now, correct. Thanks in advance.

PR #1337, opened 2022-12-20
Player warnings limited to once per minute.

FuzTheCat:
Once per minute limitation was automatically written by ChatGPT during a chat session.


map_gen/maps/diggy/feature/danger_ore.lua
```diff
@@ -29,10 +29,20 @@ local function banned_entities(allowed_entities)
     )
 
     --- Warning for players when their entities are destroyed
+    --- Note: Edit to limit warning once per minute per player produced completely automatically by ChatGPT
+    local last_warning_time = {}
```
grilledham: This would desync in multiplayer, it needs to go inside global. See `player_collapses` in https://github.com/Refactorio/RedMew/blob/develop/map_gen/maps/diggy/feature/antigrief.lua for an example.

map_gen/maps/diggy/feature/danger_ore.lua
```diff
@@ -29,10 +29,20 @@ local function banned_entities(allowed_entities)
...
         if p and p.valid then
-            p.print('You cannot build that on top of ores, only belts, mining drills, and power poles are allowed.')
+            if not last_warning_time[p.index] then
+                last_warning_time[p.index] = 0
+            end
+            local current_time = game.tick
+            if current_time > last_warning_time[p.index] + (60 * 60) then  -- 60 seconds * 60 ticks per second
+                last_warning_time[p.index] = current_time
+                p.print('You cannot build that on top of ores, only belts, mining drills, and power poles are allowed.')
+            end
```
grilledham: Minor but I think doing it that way would mean you don't get the warning in the first minute of gameplay.
I'd write it like this (untested):
```suggestion
            local current_time = game.tick
            local player_last_warning_time = last_warning_time[p.index]
            if player_last_warning_time ~= nil and current_time < player_last_warning_time + (60 * 60) then  -- 60 seconds * 60 ticks per second
                return
            end            
            last_warning_time[p.index] = current_time
            p.print('You cannot build that on top of ores, only belts, mining drills, and power poles are allowed.')
           
```

grilledham (changes requested):
Not bad for a chat bot 😂. But I guess it needs more context to know about Factorio desyncs.

map_gen/maps/diggy/feature/danger_ore.lua
FuzTheCat: Thank you. I'll see what I can do about getting local variables registered globally.

map_gen/maps/diggy/feature/danger_ore.lua
FuzTheCat: Yes, there would be no initial warning for the first minute. It could almost take a minute for the first ore to be uncovered also. But it's good to have the code doing the right thing, thank you.

FuzTheCat: OK, I've done some more edits now and tested locally in multiplayer with two accounts. Does it look good now?

grilledham (approved):

PR #1346, opened 2023-02-10
Rich item copy for inventory drops

Soggs:
Any item-with-data now keeps that when being moved to the corpse. This preserves equipment grids, spidertron remote settings, etc.

Adds an option to exclude items from dropping.

Selection tool items (blueprints, deconstruction planner, upgrade planner) and blue print books are always excluded from dropping.

### DIFF
--- a/features/dump_offline_inventories.lua
+++ b/features/dump_offline_inventories.lua
@@ -26,9 +26,10 @@ end)
 local function move_items(source, target)
     if not source.is_empty() then
         for i = 1, #source do
-            if source[i].valid_for_read and not ignored_items_set[source[i].name] then
-                target.insert(source[i])
-                source[i].clear()
+            local stack = source[i]
+            if stack.valid_for_read and not ignored_items_set[stack.name] then
+                target.insert(stack)
+                stack.clear()
             end
         end
     end
--- a/features/dump_offline_inventories.lua
+++ b/features/dump_offline_inventories.lua
@@ -27,7 +27,7 @@ local function move_items(source, target)
     if not source.is_empty() then
         for i = 1, #source do
             local stack = source[i]
-            if stack.valid_for_read and not ignored_items_set[stack.name] then
+            if stack.valid_for_read and not stack.is_selection_tool() and not stack.is_blueprint_book() and not ignored_items_set[stack.name] then
                 target.insert(stack)
                 stack.clear()
             end
--- a/features/dump_offline_inventories.lua
+++ b/features/dump_offline_inventories.lua
@@ -12,7 +12,7 @@ local set_timeout_in_ticks = Task.set_timeout_in_ticks
 local config = Config.dump_offline_inventories
 
 local ignored_items_set = {}
-for _ , k in pairs(config.ignored_items)
+for _ , k in pairs(config.ignored_items) do
     ignored_items_set[k] = true
 end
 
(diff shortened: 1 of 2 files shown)


features/dump_offline_inventories.lua
```diff
@@ -11,13 +11,30 @@ local Config = require 'config'
...
+
 local offline_player_queue = {}
 
 Global.register({offline_player_queue = offline_player_queue}, function(tbl)
     offline_player_queue = tbl.offline_player_queue
     config = Config.dump_offline_inventories
 end)
 
+local function move_items(source, target)
+    if not source.is_empty() then
```
grilledham, line 27:
```suggestion
    if source.valid and not source.is_empty() then
```
I think with the way you changed the checks it's technically possible to get here with only one of the inventories being valid.

features/dump_offline_inventories.lua
```diff
@@ -11,13 +11,30 @@ local Config = require 'config'
 local set_timeout_in_ticks = Task.set_timeout_in_ticks
 local config = Config.dump_offline_inventories
 
+local ignored_items_set = {}
+for _ , k in pairs(config.ignored_items) do
```
grilledham, line 15:
```suggestion
for _ , k in pairs(config.ignored_items or {}) do
```
There are a lot maps that directory set `dump_offline_inventories` with a new table that are missing this new field, search for `dump_offline_inventories =`. Another option would be to go to all those sites and add an empty table.

features/dump_offline_inventories.lua
```diff
@@ -11,13 +11,30 @@ local Config = require 'config'
...
 Global.register({offline_player_queue = offline_player_queue}, function(tbl)
     offline_player_queue = tbl.offline_player_queue
     config = Config.dump_offline_inventories
 end)
 
+local function move_items(source, target)
+    if not source.is_empty() then
+        for i = 1, #source do
+            local stack = source[i]
+            if stack.valid_for_read and not stack.is_selection_tool() and not stack.is_blueprint_book() and not ignored_items_set[stack.name] then
```
grilledham, line 30:
```suggestion
            if stack.valid and stack.valid_for_read and not stack.is_selection_tool and not stack.is_blueprint_book and not ignored_items_set[stack.name] then
```
According to the API these are boolean values not functions.
https://lua-api.factorio.com/latest/LuaItemStack.html#LuaItemStack.is_selection_tool
https://lua-api.factorio.com/latest/LuaItemStack.html#LuaItemStack.is_blueprint_book

Also I think bizarrely you have to check valid as well as valid for read.

grilledham (changes requested):
Thanks for the PR. A few minor issues that need fixing before we can merge.

features/dump_offline_inventories.lua
Soggs: Since we never let go of control between obtaining the reference and using it, it should never become invalid. However we do need a ``nil`` check at that point. Which gives me an idea to rework the whole control flow of how we handle things. So I will come back to it.

features/dump_offline_inventories.lua
Soggs: I will add that.

features/dump_offline_inventories.lua
Soggs: Got carried over by the ``is_empty()``

As for ``stack`` being invalid. That would imply that the inventory we pull from contains invalid references which should never be an constitute a major bug in Factorio.

grilledham: There might be an edge case if the stack gets cleared on the same tick by another event. The stack might still be there but not valid? I'm not sure, but the Factorio devs always recommend adding the valid check and in this case there is no performance concern and it doesn't make the code more complicated. So, I'd be happier with the check, same above for the inventory check.

PR #1365, opened 2023-06-28
Added 2 DO Map presets

Cinecraft:
First try to generate new Maps for DO.

### DIFF
--- a/map_gen/maps/danger_ores/presets/danger_ore_lazyone_beltboxes_ore_only.lua
+++ b/map_gen/maps/danger_ores/presets/danger_ore_lazyone_beltboxes_ore_only.lua
@@ -1,0 +1,186 @@
+local RS = require 'map_gen.shared.redmew_surface'
+local MGSP = require 'resources.map_gen_settings'
+local Event = require 'utils.event'
+local b = require 'map_gen.shared.builders'
+local Config = require 'config'
+
+local ScenarioInfo = require 'features.gui.info'
+ScenarioInfo.set_map_name('Danger Ore Lazy One Beltboxes (ore only)')
+ScenarioInfo.set_map_description([[
+Clear the ore to expand the base,
+focus mining efforts on specific sectors to ensure
+proper material ratios, expand the map with pollution!
+]])
+ScenarioInfo.add_map_extra_info([[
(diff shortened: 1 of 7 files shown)


map_gen/maps/danger_ores/presets/danger_ore_lazyone_beltboxes_ore_only.lua
```diff
@@ -0,0 +1,186 @@
...
+end)
+
+local terraforming = require 'map_gen.maps.danger_ores.modules.terraforming'
+terraforming({start_size = 8 * 32, min_pollution = 600, max_pollution = 24000, pollution_increment = 9})
+
+local rocket_launched = require 'map_gen.maps.danger_ores.modules.rocket_launched_simple'
+rocket_launched({win_satellite_count = 1000})
+
+local restart_command = require 'map_gen.maps.danger_ores.modules.restart_command'
+restart_command({scenario_name = 'danger-ore-landfill-beltboxes-ore-only'})
```
grilledham:
```suggestion
restart_command({scenario_name = 'danger-ore-lazyone-beltboxes-ore-only'})
```

grilledham (changes requested):
Cool, the maps look great! Thanks for making the PR.

Just a minor change and some additions required.

We need to add the scenario templates, see https://github.com/Refactorio/RedMew/tree/develop/scenario_templates

This is a redmew sever backend thing. The server uses the scenario templates to create the scenarios on the server. Each folder determines the name of the scenario and the content is the files that the scenario uses that are different to the base repo. So in this case just create folders called `danger-ore-for-the-swarm-beltboxes-ore-only` and `danger-ore-lazyone-beltboxes-ore-only` and add the appropriate map_selection.lua file.

This is what scenario_name refers to in the restart command and the name in the map poll.

If you want players to be able to vote for this map in the poll, you should add these maps to https://github.com/Refactorio/RedMew/blob/develop/map_gen/maps/danger_ores/modules/map_poll.lua#L26

If you can make the changes that would be great, otherwise if you prefer I can make them.

Cinecraft:
will add the changes, currently trying to get luacheck to work on Win10 and VSCodium.

grilledham (approved):

PR #1366, opened 2023-06-30
Added Expensive Mode Map

Cinecraft:


map_gen/maps/danger_ores/modules/main_ores_grid_factory.lua
```diff
@@ -0,0 +1,115 @@
...
+        -- left
+        walk_pattern[32][ 1] = right_tile -- b.tile('orange-refined-concrete')
+        walk_pattern[33][ 1] = left_tile  -- b.tile('pink-refined-concrete')
+        -- right
+        walk_pattern[32][64] = left_tile  -- b.tile('purple-refined-concrete')
+        walk_pattern[33][64] = right_tile -- b.tile('cyan-refined-concrete')
+        local hazard_grid =  b.grid_pattern_no_offset(walk_pattern,64,64,1,1)
+        hazard_grid = b.translate(hazard_grid,33,33)
+
+        return b.any {hazard_grid, spawn_shape, gear_shape, cc_circle, ores}
```
grilledham: I'm not sure what `cc_circle` does, the map looks the same without it?

map_gen/maps/danger_ores/modules/main_ores_grid_factory.lua
```diff
@@ -0,0 +1,115 @@
...
+        -- ores = b.translate(ores, 31, 31)
+        -- local concrete_map = b.throttle_xy(ores, 62, 64, 62, 64)
+        -- concrete_map = b.translate(concrete_map, -31, -31)
+        -- local water_map = b.change_tile(concrete_map, false, 'water-shallow')
+
+        -- spawn setup::
+        local gear_shape = b.add(b.scale(gear, 1), b.circle(16))
+        local square_o = b.subtract(b.rectangle(64, 64), b.rectangle(62, 62))
+        gear_shape = b.subtract(gear_shape, square_o)
+        gear_shape = b.overlay_tile_land(gear_shape, b.tile('orange-refined-concrete'))
```
grilledham: `overlay_tile_land` only replaces land (not water) tiles. This means if the base map that factorio generates has water it won't get replaced with concrete. For example if you generate the map with seed `464946681` you get 
image

`gear_shape = b.choose(gear_shape, b.tile('orange-refined-concrete'), b.empty_shape)`
should fix it as we later fall back to `ores` below in the `return b.any {...`.

map_gen/maps/danger_ores/modules/main_ores_grid_factory.lua
grilledham: As we are covering the whole map in concrete we need to also set the 'hidden' tile that sits below the concrete if the concrete is removed. For example using the same seed `464946681` if we remove the concrete north of spawn we reveal water which is quite unexpected.

image

I'd fix this by using:
```lua
local map = b.any {hazard_grid, spawn_shape, gear_shape, ores}
return b.set_hidden_tile(map, 'grass-1')
```

grilledham (changes requested):
Map looks great, thanks for the PR.

A few minor things I noticed that we probably want to fix.

map_gen/maps/danger_ores/modules/main_ores_grid_factory.lua
Cinecraft: that explains why i only saw it randomly, was wondering where that water came from since i didn't generate any.

map_gen/maps/danger_ores/modules/main_ores_grid_factory.lua
Cinecraft: it's a shadow for the gear, forgot i switched the grid to concrete as well, thats why it was invisible . Fixed in next commit.

grilledham (approved):

PR #1377, opened 2023-10-03
DangerOre Krastorio2 preset

RedRafe:
Hi, I would like to contribute with a DangerOre-Krastorio2 compatible preset.

The commits involve two major changes:
1. Compatibility added to the turret-autofill shared module 
2. A map preset for DangerOres to be used with [Krastorio2](https://mods.factorio.com/mod/Krastorio2) and [k2-danger-ores](https://github.com/RedRafe/k2-danger-ores) mods

The development was based off Redmew's the already present preset "[danger_ore_lazy_one_beltboxes_ore_only](https://github.com/Refactorio/RedMew/blob/develop/map_gen/maps/danger_ores/presets/danger_ore_lazy_one_beltboxes_ore_only.lua)", and still features:

- terraforming
- landfill everything
- no handcrafting
- no oil
- 10x tech cost multiplier
- 1000 rocket launches to win
- deadlock's loaders&beltboxes compatibility for essential raw/processed items

Some changes were necessarily made at data-stage by the side mod [k2-danger-ores](https://github.com/RedRafe/k2-danger-ores), which is kinda required to run the map with K2 mod, but presents adjustable settings for some features to be enabled/disabled at one's preferences.

### DIFF
--- a/features/autofill.lua
+++ b/features/autofill.lua
@@ -19,6 +19,13 @@ local ammo_locales = {
     ['piercing-rounds-magazine'] = {'item-name.piercing-rounds-magazine'},
     ['firearm-magazine'] = {'item-name.firearm-magazine'}
 }
+local k2_ammo_locales = {
+    ['rifle-magazine'] = {'item-name.rifle-magazine'},
+    ['armor-piercing-rifle-magazine'] = {'item-name.armor-piercing-rifle-magazine'},
+    ['uranium-rifle-magazine'] = {'item-name.uranium-rifle-magazine'},
+    ['imersite-rifle-magazine'] = {'item-name.imersite-rifle-magazine'}
+}
+if script.active_mods["Krastorio2"] then ammo_locales = k2_ammo_locales end
 Public.ammo_locales = ammo_locales
 
 local player_ammos = {} -- player_index -> dict of name -> bool
@@ -35,6 +42,14 @@ local default_ammos = {
     ['piercing-rounds-magazine'] = true,
     ['firearm-magazine'] = true
 }
+
+local k2_default_ammos = {
+    ['rifle-magazine'] = true,
+    ['armor-piercing-rifle-magazine'] = true,
+    ['uranium-rifle-magazine'] = true,
+    ['imersite-rifle-magazine'] = true
+}
+if script.active_mods["Krastorio2"] then default_ammos = k2_default_ammos end
 
 local function copy(tbl)
     local result = {}
--- a/map_gen/maps/danger_ores/modules/rocket_launched_krastorio2.lua
+++ b/map_gen/maps/danger_ores/modules/rocket_launched_krastorio2.lua
@@ -3,11 +3,9 @@ local RS = require 'map_gen.shared.redmew_surface'
 local Server = require 'features.server'
 local ShareGlobals = require 'map_gen.maps.danger_ores.modules.shared_globals'
 
-return function(config)
+return function()
     ShareGlobals.data.biters_disabled = false
     ShareGlobals.data.map_won = false
-
-    local win_satellite_count = config.win_satellite_count or 1000
 
     local function disable_biters()
         if ShareGlobals.data.biters_disabled then
@@ -26,12 +24,6 @@ return function(config)
         }
         game.print({'danger_ores.biters_disabled_k2'})
         Server.to_discord_bold(message)
-    end
-
-    local function print_satellite_message(count)
-        -- Just log the no. of satellite(s) launched
-        game.print({'danger_ores.satellite_logger', count})
-        Server.to_discord_bold('Already launched ' .. tostring(count) .. ' satellie(s)')
     end
 
     local function rocket_launched(event)
@@ -56,10 +48,6 @@ return function(config)
 
         if satellite_count == 1 then
             disable_biters()
-        end
-
-        if (satellite_count % 50) == 0 then
-            print_satellite_message(satellite_count)
         end
     end
 
--- a/map_gen/maps/danger_ores/modules/rocket_launched_krastorio2.lua
+++ b/map_gen/maps/danger_ores/modules/rocket_launched_krastorio2.lua
@@ -1,0 +1,92 @@
+local Event = require 'utils.event'
+local RS = require 'map_gen.shared.redmew_surface'
+local Server = require 'features.server'
+local ShareGlobals = require 'map_gen.maps.danger_ores.modules.shared_globals'
+
+return function(config)
+    ShareGlobals.data.biters_disabled = false
+    ShareGlobals.data.map_won = false
+
+    local win_satellite_count = config.win_satellite_count or 1000
+
+    local function disable_biters()
+        if ShareGlobals.data.biters_disabled then
+            return
+        end
+
+        ShareGlobals.data.biters_disabled = true
+        game.forces.enemy.kill_all_units()
+        for _, enemy_entity in pairs(RS.get_surface().find_entities_filtered({force = 'enemy'})) do
+            enemy_entity.destroy()
+        end
+
+        local message = table.concat {
+            'Launching the first satellite has killed all the biters. ',
+            'Launch ',
+            win_satellite_count,
+            ' satellites to win the map.'
+        }
+        game.print({'danger_ores.biters_disabled', win_satellite_count})
+        Server.to_discord_bold(message)
+    end
+
+    local function print_satellite_message(count)
+        -- Just log the no. of satellite(s) launched
+        game.print({'danger_ores.satellite_logger', count})
+        Server.to_discord_bold('Already launched ' .. tostring(count) .. ' satellie(s)')
+    end
+
+    local function rocket_launched(event)
+        if ShareGlobals.data.map_won then
+            return
+        end
+
+        local entity = event.rocket
+        if not entity or not entity.valid or not entity.force == 'player' then
+            return
+        end
+
+        local inventory = entity.get_inventory(defines.inventory.rocket)
+        if not inventory or not inventory.valid then
+            return
+        end
+
+        local satellite_count = game.forces.player.get_item_launched('satellite')
+        if satellite_count == 0 then
+            return
+        end
+
+        if satellite_count == 1 then
+            disable_biters()
+        end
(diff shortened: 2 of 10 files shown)


RedRafe:
I've kept the autofill changes separate, as yes, it would be cool to autofill k2 ammos, but the scenario is in peaceful mode so it's not strictly required. The autofill module is shared with all other maps, so I don't really know if that would be a good place to dump non-vanilla code

map_gen/maps/danger_ores/presets/danger_ore_krastorio2.lua
```diff
@@ -0,0 +1,230 @@
...
+    p.technologies['mining-productivity-2'].enabled = false
+    p.technologies['mining-productivity-3'].enabled = false
+    p.technologies['mining-productivity-4'].enabled = false
+    p.technologies['mining-productivity-11'].enabled = false
+    p.technologies['mining-productivity-16'].enabled = false
+    p.technologies['kr-decorations'].enabled = false
+
+    p.manual_mining_speed_modifier = 1
+
+    game.difficulty_settings.technology_price_multiplier = 20
```
grilledham: PR description says 10x tech multiplier. Did you intend 10 or 20?

map_gen/maps/danger_ores/presets/danger_ore_krastorio2.lua
```diff
@@ -0,0 +1,230 @@
...
+
+    Task.set_timeout_in_ticks(60, kr_remote)
+end)
+
+local terraforming = require 'map_gen.maps.danger_ores.modules.terraforming'
+terraforming({start_size = 8 * 32, min_pollution = 600, max_pollution = 24000, pollution_increment = 9})
+
+--[[ Win condition in K2: build intergalactic transceiver ]]
+-- local rocket_launched = require 'map_gen.maps.danger_ores.modules.rocket_launched_simple'
+-- rocket_launched({win_satellite_count = 1000})
```
grilledham: The `rocket_launched_simple` file is used to active the restart map command. Sets `ShareGlobals.data.map_won = true` and prints the win message.

I guess we will need to replace it with a krastorio version. Is there an event or way we can detect when krastorio/the game is won?

features/autofill.lua
```diff
@@ -19,6 +19,13 @@ local ammo_locales = {
     ['piercing-rounds-magazine'] = {'item-name.piercing-rounds-magazine'},
     ['firearm-magazine'] = {'item-name.firearm-magazine'}
 }
+local k2_ammo_locales = {
+    ['rifle-magazine'] = {'item-name.rifle-magazine'},
+    ['armor-piercing-rifle-magazine'] = {'item-name.armor-piercing-rifle-magazine'},
+    ['uranium-rifle-magazine'] = {'item-name.uranium-rifle-magazine'},
+    ['imersite-rifle-magazine'] = {'item-name.imersite-rifle-magazine'}
+}
+if script.active_mods["Krastorio2"] then ammo_locales = k2_ammo_locales end
```
grilledham, line 28: FYI, I think this a reasonable solution.

grilledham (changes requested):
Good stuff!

There are some changes/additions it would be good to make. Either you can make the changes, or I can make them and push to your PR, or we can merge your PR and I'll submit another PR with the changes. Let me know what you would prefer.

The `achievement/there-is-no-spoon` sprite doesn't seem to be available when using Krastorio2? So we might need to use another sprite for for the `gui/score.lua` button. Note: If you add `_DEBUG = true` to the top of your `map_selection.lua` file you will see error in the redmew scenario, rather than them being swallowed.

Can we add this scenario to the `scenario_templates` directory. This is a redmew backend server thing that it uses to know which scenarios can be launched on the server. Note: The name of the scenario in the restart command `restart_command({scenario_name = 'danger-ore-krastorio2'})` needs to be the same as the directory in `scenario_templates`.

We could add this map to the `map_poll.lua` if you want people to be able to vote to play this scenario. We can use a placeholder like `krastorio_mod_pack` for the mod pack. I'll add the mod pack to the redmew server.

grilledham:
Some luacheck failures I can fix if you would prefer. <https://github.com/Refactorio/RedMew/actions/runs/6394626624/job/17365979698?pr=1377>

Looks like `active_mods` needs to be added to script -> fields in `.luacheckrc`.

map_gen/maps/danger_ores/presets/danger_ore_krastorio2.lua
RedRafe: Yeah it's 20x sorry. Tho I could turn it down a bit for the first release and have it 10x now that I think about it

map_gen/maps/danger_ores/presets/danger_ore_krastorio2.lua
RedRafe: TL;DR: I'll look into it

Factorio Freeplay has its win con at 1st rocket launched.

Normally, Krastorio calls the remote to disable it and calls the win con when its condition is met (build the transceiver, late-late game. Requires ~10 launches of tech or so to complete). But it also offers its own API to enable/disable. So there could be a way to catch when scenario is won to restart the map.

K2 condition bypasses both vanilla and Redmew conditions, so I've tweaked the tech cost to be both unaffected by multipliers, and be as expensive as 1000x rockets. The numbers could be adjusted via side-mod settings. RN it auto computes the cost and 1000 rockets is about 3.9 M science packs

RedRafe: Precisely, here: https://github.com/RedRafe/k2-danger-ores/blob/master/prototypes/technology/intergalactic-transceiver.lua

RedRafe:
- [x] fixed [achievement sprite](https://github.com/RedRafe/k2-danger-ores/commit/e84de65fe44fa0b39efb1b0e9079e1afeee29c18)
- [x] fixed some workflow issues & added missing commands to `.luacheckrc`
- [x] added [scenario to map poll](https://github.com/Refactorio/RedMew/pull/1377/commits/d75ddbef27be4948a6f2d48f38f0b1e709000f95)
- [x] added [scenario to templates](https://github.com/Refactorio/RedMew/pull/1377/commits/7ead4174edb67ab5d3da3c70e85c8e166fa5184e)
- [x] implemented a custom `rocket_launched` module ([here](https://github.com/Refactorio/RedMew/pull/1377/commits/8319bc95a18db28494f7a52154934d6b322c309a) and on [side-mod](https://github.com/RedRafe/k2-danger-ores/commit/7919510685263c16d36f4ae49313fe86fbe36baa))
- [x] changed [tech multiplier](https://github.com/Refactorio/RedMew/pull/1377/commits/da3ac610a443db7500616a67dad7d92a2409aa7e) 20x -> 10x (it is possible to set a tech multiplier in map preview, they will now combine) 
- [x] removed the no-handcrafting/LazyB setting. Maybe that will be for another preset, but for now I think would be easier with this option enabled (also changed the starting items accordingly)

grilledham:
This is great stuff!
I have taken a quick look but I won't be able to review this fully until Friday at the earliest, failing that I should have time at the weekend.

map_gen/maps/danger_ores/modules/rocket_launched_krastorio2.lua
```diff
@@ -0,0 +1,92 @@
...
+        game.forces.enemy.kill_all_units()
+        for _, enemy_entity in pairs(RS.get_surface().find_entities_filtered({force = 'enemy'})) do
+            enemy_entity.destroy()
+        end
+
+        local message = table.concat {
+            'Launching the first satellite has killed all the biters. ',
+            'Launch ',
+            win_satellite_count,
+            ' satellites to win the map.'
```
grilledham: I think we should change this message. Instead it should say:
```lua
'Launching the first satellite has killed all the biters. ',
'Build and activate the intergalactic transceiver to win.'
```
My understanding is that you only need to launch satellites to get the space science to research the intergalactic transceiver. So launching the satellites isn't enough to win. Also the amount you have to launch depends on productivity modules used in the research labs, so probably best to not say the amount.

map_gen/maps/danger_ores/modules/rocket_launched_krastorio2.lua
```diff
@@ -0,0 +1,92 @@
...
+        }
+        game.print({'danger_ores.biters_disabled', win_satellite_count})
```
grilledham: Likewise, we should make a krastorio version of this message, that is the same/similar as above.

map_gen/maps/danger_ores/modules/rocket_launched_krastorio2.lua
```diff
@@ -0,0 +1,92 @@
...
+        Server.to_discord_bold(message)
+    end
+
+    local function print_satellite_message(count)
+        -- Just log the no. of satellite(s) launched
+        game.print({'danger_ores.satellite_logger', count})
+        Server.to_discord_bold('Already launched ' .. tostring(count) .. ' satellie(s)')
```
grilledham: We don't need to print theses messages, we have another script that already does this. Before this was useful because it would print the remaining count to win, but I don't think we want to do that in this case.

grilledham (changes requested):
Just a few changes I think we should make tot he rocket_launched_krastorio2 script, then I think we can merge.

grilledham:
Thanks. I think we still need to remove the `print_satellite_message` function though.

RedRafe:
> Thanks. I think we still need to remove the `print_satellite_message` function though.

That function only prints "Launched X rockets" without any target. Could be useful to know the progression out of the 1000 estimated rockets (from discord at least). But tyou can comment it out.

I'll be out for the weekend so I pass the ball to you now ;) I'll be on my phone tho so I'll be able to reply to messages, but not code

grilledham:
Cool. I can make the last small change.

`score.lua` <https://github.com/Refactorio/RedMew/blob/develop/features/gui/score.lua#L143> already prints the satellite counts, so we don't need `rocket_launched_krastorio2.lua` to do that. Previously the rocket_launched script printed the remaining count to win, but the krastorio version doesn't do that.

grilledham (approved):

PR #1383, opened 2023-10-27
Add new DangerOre presets

RedRafe:
(I'm back)

Add new DangerOre modded maps with the following mods:
- BZ mods
- Exotic Industries
- Omnimatter
- Industrial Revolution 3
- Pyanodon

[Here](https://www.notion.so/RedMew-project-9195bb3f1555452a8e35f59d0d72f9e9#0ad66d25f52c408c91d80ac4b2df24a5) is the full preview.

This is the data part: [redmew-data](https://github.com/RedRafe/redmew-data)


map_gen/maps/danger_ores/modules/biter_drops_exotic_industries.lua
```diff
@@ -0,0 +1,69 @@
...
+    for _=0, weight+1 do
+      weighted_table[total] = item
+      total = total + 1
+    end
+  end
+
+  table.shuffle_table(weighted_table)
+
+  return weighted_table[random:next_int(1, total)]
+end
```
grilledham: I don't think using `map_gen.shared.random` like that is desync safe in multiplayer. I think this can be easy converted to use `math.random` instead.

Also in my suggestion I changed the for loop range to `1, weight` because for loop ranges are inclusive so `0, weight+1` would be two extra?

```suggestion
local weights = {
  ['ei_energy-crystal'] =  10,
  ['ei_gold-chunk']     =  10,
  ['ei_alien-resin']    = 150,
  ['ei_alien-seed']     =   4,
  ['ei_alien-beacon']   =   1,
}

local weighted_table = {}
for item, weight in pairs(weights) do
  for _ = 1, weight do
    weighted_table[#weighted_table + 1] = item
  end
end

local function draw_random()
  return weighted_table[math.random(#weighted_table)]
end
```

map_gen/maps/danger_ores/presets/danger_ore_omnimatter.lua
```diff
@@ -0,0 +1,157 @@
...
+})
+
+-- Config.lazy_bastard.enabled = true
+Config.market.enabled = false
+Config.player_rewards.enabled = false
+Config.player_create.starting_items = {
+    { count =  1, name = 'stone-furnace'},
+    { count =  2, name = 'burner-mining-drill' },
+    { count = 50, name = 'wood' },
+    { count =  1, name = 'burner-omnitraction' },
```
grilledham: 'burner-omnitraction' isn't a valid item name?

image

map_gen/maps/danger_ores/modules/concrete_on_landfill_omni.lua
```diff
@@ -0,0 +1,113 @@
+local Event = require 'utils.event'
+
+return function(config)
+    local replace_tile = config.tile or 'blue-refined-concrete'
+
+    local replace_tiles = {['landfill'] = true, [replace_tile] = true}
+    local brush_tools = {['omnite-refined-concrete'] = true,} -- , ['refined-hazard-concrete'] = true}
```
grilledham: Is this the only way this differs from `concrete_on_landfill.lua`? If so maybe we should pass in `{['omnite-refined-concrete'] = true}` by the `config` and have it default to `{['refined-concrete'] = true}`.

map_gen/maps/danger_ores/modules/map_without_gaps.lua
```diff
@@ -0,0 +1,141 @@
...
+    return b.empty_shape
+end
+
+return function(config)
+    local start_ore_shape
+    local resource_patches
+    local no_resource_patch_shape
+    local dense_patches
+
+    local function ore_builder(ore_name, amount, ratios, weighted)
```
grilledham: Is `ore_builder` the only way this differs from `map.lua`? If so, could we add this version of `ore_builder` to the module folder called something like `ore_builder_without_gaps.lua` then pass it in through `config`.

map_gen/shared/builders.lua
```diff
@@ -1065,6 +1065,18 @@ end
...
+--- it'll return that shape. Otherwise, empty shape is returned
+function Builders.exclusive_pattern(pattern)
+    return function(x, y, world)
+        for index, shape in pairs(pattern or {}) do
+            local tile = shape.shape(x, y, world)
+            if tile and tile.entities then return shape.shape(x, y, world) end
+        end
+        return Builders.empty_shape(x, y, world)
+    end
+end
```
grilledham:
```suggestion
--- Loops through the shapes, if any shape is able to place any entities,
--- it'll return that tile. Otherwise, false is returned.
function Builders.any_entity_pattern(pattern)
    return function(x, y, world)
        for index, shape in pairs(pattern or {}) do
            local tile = shape.shape(x, y, world)
            if tile and tile.entities then return tile end
        end
        return false
    end
end
```

map_gen/maps/danger_ores/modules/map_poll.lua
```diff
@@ -18,11 +18,21 @@ end)
...
+local bz_mod_pack = 'bz_mod_pack'
+local ei_mod_pack = 'ei_mod_pack'
+local py_short_mod_pack = 'py_short_mod_pack'
+local ir3_mod_pack = 'ir3_mod_pack'
 
 local mod_packs = {
     normal_mod_pack = 'danger_ore29',
     bobs_mod_pack = 'danger_ore_bobs3',
-    krastorio_mod_pack = 'danger_ore_krastorio2'
+    krastorio_mod_pack = 'danger_ore_krastorio2',
```
grilledham: Just bumping the mod pack version.

```suggestion
    krastorio_mod_pack = 'danger_ore_krastorio3',
```

grilledham (changes requested):
Some minor changes, then I think this is good to go.
Also I've uploaded all the mod packs to the server.

map_gen/maps/danger_ores/modules/biter_drops_exotic_industries.lua
RedRafe: Yeah it totally makes sense. I don't even know where the 0/+1 came from

map_gen/maps/danger_ores/modules/concrete_on_landfill_omni.lua
```diff
@@ -0,0 +1,113 @@
...
+                surface.set_hidden_tile(tile.position, 'landfill')
+            end
+
+            if old_name == replace_tile then
+                refund_count = refund_count + 1
+            end
+        end
+
+        if inventory and inventory.valid and refund_count > 0 then
+            inventory.insert {name = 'omnite-refined-concrete', count = refund_count}
```
RedRafe: @grilledham this and L52 are the other places where the vanilla concrete was hardcoded. I guess we could pass as param `refound_tile` alongside `replace_tile`

grilledham (approved):
Fantastic work as always! Thanks for making the changes.

PR #1390, opened 2023-12-20
Fix DO+EI restart conditions

RedRafe:
- fixed typo in the rocket message
- fixed conditions for the restart command not firing


map_gen/maps/danger_ores/modules/rocket_launched_exotic_industries.lua
```diff
@@ -64,7 +64,7 @@ return function()
     end
 
     local function on_win_condition_met()
-        if ShareGlobals.show_reset_message and game.finished_but_continuing then
+        if ShareGlobals.data.show_reset_message and game.finished_but_continuing then
```
grilledham: Does `show_reset_message` ever get set to anything other than `true`?

RedRafe: Now that I look at it... No 🤔
I guess I could use `not SharedG.data.map_won` instead

grilledham (approved):

PR #1392, opened 2024-01-12
Enhance tasklist GUI with updated number of tasks

RedRafe:
Enhance tasklist GUI by adding the current number of tasks displayed on top of the icon
Following the suggestion on discord [discord/RedMew/devtalk](https://discord.com/channels/432567222481846283/432570869039104012/1194789531694682152)

Future improvements ideas (not included):
- add setting to display, instead, only the # on **new** tasks not read (notification system)
- add setting to display, instead, only the # of tasks which dont have volunteers yet (allocation system)

Here's a quick reference of hot it looks, pretty simple

Screenshot from 2024-01-12 12-26-01

### DIFF
--- a/.luacheckrc
+++ b/.luacheckrc
@@ -42,7 +42,7 @@
 --[Set Defaults]--
 -------------------------------------------------------------------------------
 local LINE_LENGTH = false -- It is 2017 limits on length are a waste
-local IGNORE = {'21./%w+_$', '21./^_%w+$', '213/[ijk]', '213/index', '213/key'}
+local IGNORE = {'21./%w+_$', '21./^_%w+$', '213/[ijk]', '213/index', '213/key', '58[1-2]'}
 local NOT_GLOBALS = {'coroutine', 'io', 'socket', 'dofile', 'loadfile'} -- These globals are not available to the factorio API
 
 local STD_CONTROL = 'lua52c+factorio+factorio_control+stdlib+factorio_defines'
--- a/features/gui/tasklist.lua
+++ b/features/gui/tasklist.lua
@@ -213,7 +213,7 @@ end
 local function update_top_gui(player)
     local button = player.gui.top[main_button_name]
     if button and button.valid then
-        button.number = table_size(tasks) or 0
+        button.number = #tasks or 0
     end
 end
 
@@ -630,7 +630,7 @@ local function player_created(event)
             name = main_button_name,
             sprite = 'item/repair-pack',
             tooltip = {'tasklist.tooltip'},
-            number = table_size(tasks) or 0,
+            number = #tasks or 0,
         }
     )
 end
@@ -1117,8 +1117,6 @@ Gui.on_click(
             elseif notify then
                 draw_main_frame(left, p)
             end
-
-            update_top_gui(p)
 
             if notify then
                 p.print(message)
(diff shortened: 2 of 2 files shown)


.luacheckrc
```diff
@@ -42,7 +42,7 @@
 --[Set Defaults]--
 -------------------------------------------------------------------------------
 local LINE_LENGTH = false -- It is 2017 limits on length are a waste
-local IGNORE = {'21./%w+_$', '21./^_%w+$', '213/[ijk]', '213/index', '213/key'}
+local IGNORE = {'21./%w+_$', '21./^_%w+$', '213/[ijk]', '213/index', '213/key', '58[1-2]'}
```
grilledham, line 45: > '58[1-2]'

What is this? Not that I'm objecting, I just don't know what this means 😄.

features/gui/tasklist.lua
```diff
@@ -210,6 +210,13 @@ local function update_volunteer_button(button, task)
     end
 end
 
+local function update_top_gui(player)
+    local button = player.gui.top[main_button_name]
+    if button and button.valid then
+        button.number = table_size(tasks) or 0
```
grilledham:
```suggestion
        button.number = #tasks
```
We can use the `#` operator in this case as the keys are consecutive. We use it below too. So for consistency and because it should perform better we should use it.

features/gui/tasklist.lua
```diff
@@ -620,7 +629,8 @@ local function player_created(event)
             type = 'sprite-button',
             name = main_button_name,
             sprite = 'item/repair-pack',
-            tooltip = {'tasklist.tooltip'}
+            tooltip = {'tasklist.tooltip'},
+            number = table_size(tasks) or 0,
```
grilledham:
```suggestion
            number = #tasks
```

features/gui/tasklist.lua
```diff
@@ -1104,6 +1118,8 @@ Gui.on_click(
                 draw_main_frame(left, p)
             end
 
+            update_top_gui(p)
```
grilledham: `update_top_gui` probably not needed for this case either. I would think we only need it for create/delete, unless I'm missing something?

grilledham (commented):
Nice feature. Some minor comments.

.luacheckrc
RedRafe: [Luacheck docs/list of warnings](https://luacheck.readthedocs.io/en/stable/warnings.html) 

They're `prone negation` and inversion of conditions... Since GitHub workflows is an older version of luackeck which doesn't track those two (that are mainly style warning, not errors), I'm adding them to the ignore list or else when running it locally with never version I get flooded by excess warnings not tracked by the repo 😄

features/gui/tasklist.lua
RedRafe: Actually, the `table_size` methods is provided by the Factorio lib and implemented in C++, should be safer than the # operator

features/gui/tasklist.lua
RedRafe: Yeah true, I added it whenever all players GUI were iterated and updated for consistency 😄

features/gui/tasklist.lua
grilledham: `table_size` is O(n) whereas `#` is O(1). The reason for `table_size` is you can't use `#` when you use the table as a dict, but in this case we use it as an array.

RedRafe: Alright, I will change it to # then, thanks for the clarification. I guess table_size is just the c++ interpretation of the Lua counterpart but not faster than # then 😄

PR #1394, opened 2024-01-18
Add popup chat feature

RedRafe:
Add chat message to map surface too as temporarily popups attached to the player and with the same color.

Configurable by `config`
- `enabled`: true/false
- `min_lifetime`: 5s
- `max_lifetime`: 20s
- allows rich-text 

The longer the message is, the longer it lasts (clamped between min/max config values).

Screenshot from 2024-01-18 19-15-25

---

Screenshot from 2024-01-18 19-15-57

---

Screenshot from 2024-01-18 19-16-35

### DIFF
--- a/features/popup_chat.lua
+++ b/features/popup_chat.lua
@@ -4,8 +4,8 @@ local Global = require 'utils.global'
 local config = global.config.popup_chat
 local MIN_LIFETIME = config.min_lifetime or 06 * 60 -- 06s
 local MAX_LIFETIME = config.max_lifetime or 20 * 60 -- 20s
-local MIN_MESSAGE_LENGTH = 40
-local MAX_MESSAGE_LENGTH = 92
+local MIN_MESSAGE_LENGTH = config.min_length or 40
+local MAX_MESSAGE_LENGTH = config.max_length or 92
 local TIME_PER_CHAR = 3 -- about +1 sec every 20 chars (60/20 ticks/chars)
 
 local data = {
@@ -17,7 +17,7 @@ Global.register(data, function(tbl)
 end)
 
 ---@param message string
-local function message_lifetime(message)
+local function get_message_lifetime(message)
   local length = message:len()
   if length <= MIN_MESSAGE_LENGTH then
     return MIN_LIFETIME
@@ -27,7 +27,7 @@ local function message_lifetime(message)
 end
 
 ---@param message string
-local function safe_message(message)
+local function get_safe_message(message)
   local length = message:len()
   if length <= MAX_MESSAGE_LENGTH then
     return message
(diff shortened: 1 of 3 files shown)


RedRafe:
After testing it in MP with TheKid we decided it was safer to set a max length to message being displayed (& adjusted the offset to not overlap with player's name)

Screenshot from 2024-01-18 19-56-15

features/popup_chat.lua
```diff
@@ -0,0 +1,76 @@
...
+---@param event defines.event.on_console_chat
+local function on_console_chat(event)
+  local index = event.player_index
+  local message = event.message
+  if not (index and message) then
+    return
+  end
+
+  local player = game.players[event.player_index]
+  if not (player and player.character) then
```
grilledham:
```suggestion
  if not (player and player.valid and player.character) then
```
Very unlikely but technically player can be not valid.

features/popup_chat.lua
```diff
@@ -0,0 +1,76 @@
...
+
+  popup_ID = rendering.draw_text({
+    text = safe_message(message),
+    surface = player.surface,
+    target = player.character,
+    target_offset = {0, -4},
+    color = color,
+    font = 'compilatron-message-font',
+    scale = 1.75,
+    time_to_live = message_lifetime(message),
```
grilledham: Should we calculate time based on the safe message, not the original message?

features/popup_chat.lua
```diff
@@ -0,0 +1,76 @@
...
+    return
+  end
+
+  local popup_ID = data.popup_chat[index]
+  if popup_ID then
+    rendering.destroy(popup_ID)
+    data.popup_chat[popup_ID] = nil
+  end
+
+  local color = table.deepcopy(player.color)
```
grilledham: You don't need the `table.deepcopy`. I think whenever you get back a regular lua table from the Factorio API you don't need to make a copy.

features/popup_chat.lua
```diff
@@ -0,0 +1,76 @@
...
+
+---@param event defines.event.on_console_chat
+local function on_console_chat(event)
+  local index = event.player_index
+  local message = event.message
+  if not (index and message) then
+    return
+  end
+
+  local player = game.players[event.player_index]
```
grilledham:
```suggestion
  local player = game.players[index]
```

grilledham (commented):
Nice feature, I'm sure people will like this.

Just some minor comments.

RedRafe:
Yeah makes sense to pass the safer message to the time computing as well. I've added the min/max length to the config file along some info commentary too, could be useful

grilledham (approved):

PR #1397, opened 2024-02-09
Safety Ores Scenario

R-Nukem:
Adding a new scenario, Safety Ores. Inspired by Zengief.

The opposite of danger ores. You can ONLY build on the ores, with a small list of exceptions to facilitate rail connections between ore patches, and to make water and oil gathering possible. 

When ore patches are consumed, you lose the ability to build on those tiles. Existing entities will not be moved or destroyed, but if mined, they will not be able to be replaced.

### DIFF
--- a/map_gen/maps/safety_ores.lua
+++ b/map_gen/maps/safety_ores.lua
@@ -2,7 +2,6 @@
 --- all local resources
 local RS = require 'map_gen.shared.redmew_surface'
 local MGSP = require 'resources.map_gen_settings'
-local MSP = require 'resources.map_settings'
 local ScenarioInfo = require 'features.gui.info'
 local config = global.config
 local RE = require 'map_gen.shared.entity_placement_restriction'
@@ -152,7 +151,7 @@ local function on_built_pumpjack(event)
 					entity.surface.create_entity({name="stone", amount=density, position={x, y}})
 				end
 			end
-		end	
+		end
 	end
 end
 Event.add(defines.events.on_robot_built_entity, on_built_pumpjack)
--- a/map_gen/maps/safety_ores.lua
+++ b/map_gen/maps/safety_ores.lua
@@ -1,6 +1,5 @@
 ---Safety Ores, by R. Nukem, inspired by Zengief
 ---all local resources
-local b = require 'map_gen.shared.builders'
 local RS = require 'map_gen.shared.redmew_surface'
 local MGSP = require 'resources.map_gen_settings'
 local ScenarioInfo = require 'features.gui.info'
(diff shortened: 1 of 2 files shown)


map_gen/maps/safety_ores.lua
```diff
@@ -0,0 +1,97 @@
...
+    [[
+2024-02-08 - R. Nukem
+- Initial Map Creation
+]]
+)
+--- Market Config
+config.currency = nil
+config.market.enabled = false
+config.player_rewards.enabled = false
+config.redmew_qol.loaders = false
```
grilledham: Why are loaders turned off? If it's because the map is intended to be played with a mod that adds loaders then that is fine.

grilledham (commented):
Code look good to me, but I haven't tested it. One minor request and question though.

Can you add a folder under `scenario_templates` with a `map_selection.lua` file. The name of the folder is the name of the scenario and the `map_selection.lua` just points to this file. See https://github.com/Refactorio/RedMew/blob/develop/scenario_templates/beach/map_selection.lua for an exmaple.

The Redmew server uses the scenario templates to create scenarios on the server.

map_gen/maps/safety_ores.lua
R-Nukem: Loaders being turned off was more a design choice than anything. We can keep them on if players are used to them and will be upset they are missing. The thought was loaders would allow placing stations in much smaller spaces, but maybe that's not a bad thing.

R-Nukem:
I'll work on the scenario templates folder. 

I also have a few changes to make after some playtesting today that will resolve some issues players found. I will turn loaders back on when I merge those changes.

map_gen/maps/safety_ores.lua
```diff
@@ -0,0 +1,159 @@
...
+--- Warning for players when their entities are destroyed
+local function on_destroy(event)
+	local p = event.player
+	if p and p.valid then
+		p.surface.create_entity{name="flying-text", position = p.position, text = 'You can only build that on top of ores, the ground is too soft'}
+	end
+end
+Event.add(RE.events.on_restricted_entity_destroyed, on_destroy)
+--- Spawn stone around pumpjacks so oil can be defended
+local function on_built_pumpjack(event)
```
grilledham, line 135: This is fine, if this is what you want to do, but it is a bit odd that you effectively let players spawn stone. An alternative solution would be to check for oil with a larger area in the `set_keep_alive_callback`. But only a suggestion.

grilledham (approved):
LGTM. Let me know when you want me merge it.

map_gen/maps/safety_ores.lua
R-Nukem: Yeah, I limited it to spawning 1 per tile to prevent it being exploitable.

The problem brought up in testing was "how do we defend the pumpjacks?". You can't build walls or turrets unless they are on ores, so this was the solution.

R-Nukem:
Just had another QA session. No more changes needed that we can tell. Ready to merge. Thanks grilled!

PR #1399, opened 2024-02-17
Update manhattan.lua

Claude477:
-changed spawn point to center
-scaled map 2 to 1(easier handle monsters)
-normaly take 70hours complete map

### DIFF
--- a/map_gen/maps/crash_site/presets/manhattan.lua
+++ b/map_gen/maps/crash_site/presets/manhattan.lua
@@ -30,8 +30,8 @@ local config = {
         MGSP.enemy_none,
         MGSP.cliff_none
     },
-    grid_number_of_blocks = 65,
-    mini_grid_number_of_blocks = 123,
+    grid_number_of_blocks = 30,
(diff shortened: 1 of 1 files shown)


grilledham (commented):
If we are making the map smaller and moving the spawn to the center then I think we can reduce `grid_number_of_blocks` and `mini_grid_number_of_blocks`. Just have to check the whole map is covered in outposts after reducing the numbers.

I think we should wait for Tigress or other regular players of crashsite to confirm the new settings are better before merging.

Claude477:
Yes, i was thinking same 👍

Claude477:
20% smaller map and location changed. as Noted

Claude477:
40% smaller map. as Noted

Claude477:
reduce grid_number_of_blocks and mini_grid_number_of_blocks about 50%~

grilledham:
The outpost number of blocks have to odd else spawn doesn't work.
image
So I fixed that.

PR #1401, opened 2024-03-04
Add players' logistic requests synch between instances

RedRafe:
Just like the quick-bar module, allows saving/loading/deleting of the player's logistic requests between instances, accessing the data on our server.

Additionally, since the logi bars are not enabled on new games until logibots are researched, upon finishing the research of the logi tech, server will attempt at loading each player's data

(I reordered the color file alphabetically, but no changes there)


features/player_logistic_requests.lua
```diff
@@ -0,0 +1,181 @@
...
+        Server.try_get_data(data_set_name, p.name, set_bars_callback)
+    end
+)
+
+--- Auto loads all logistic requests for connected players when logistics gets researched
+local research_finished =
+    Token.register(
+    function(event)
+        local tech = event.research
+        if not (tech and tech.name and tech.name == 'logistic-robotics') then
```
grilledham: Minor but if `tech.name` equals 'logistic-robotics' then it can't be falsely, so no need for  `tech.name` check.

```suggestion
        if not (tech and tech.name == 'logistic-robotics') then
```

Though personally I'd write this as `if not tech or tech.name ~= 'logistic-robotics' then`

features/player_logistic_requests.lua
```diff
@@ -0,0 +1,181 @@
...
+Global.register(
+    {
+        primitives = primitives
+    },
+    function(tbl)
+        primitives = tbl.primitives
+    end
+)
+
+--- Scans all quickbar_slots into a table, then saves that table server-side
```
grilledham: Minor: You have kept 'quickbar' in some places from when you copied this code. Maybe we should change to 'logistic'?

features/player_logistic_requests.lua
```diff
@@ -0,0 +1,181 @@
...
+        for _, player in pairs(game.connected_players) do
+            if (player and player.valid and player.character and player.character.valid) then
+                load_bars(nil, player)
+            end
+        end
+    end
+)
+
+--- Registers the event only when the server is available.
+local function register_player_create()
```
grilledham: Minor: I'd change the name of this function now it does more than just register the player create event. Maybe 'register_server_start_events'?

features/player_logistic_requests.lua
```diff
@@ -0,0 +1,181 @@
...
+local research_finished =
+    Token.register(
+    function(event)
+        local tech = event.research
+        if not (tech and tech.name and tech.name == 'logistic-robotics') then
+            return
+        end
+
+        for _, player in pairs(game.connected_players) do
+            if (player and player.valid and player.character and player.character.valid) then
```
grilledham: Really minor: You don't need the brackets in lua for if statements and we seem to have standardized on that convention.
```suggestion
            if player and player.valid and player.character and player.character.valid then
```

grilledham (commented):
Looks really good 👍. I wrote some minor comments because I'm too pedantic 😄

RedRafe:
Resolved all. There's only one minor inconvenience left:

logistic requests are enabled when logistcs is researched. I do apply saved data at the research completion & for all players joining for the first time after that. But if a player had already joined & left, they will have to manually say `/logistic-requests-load` upon rejoins to apply that. Call it good, or add more logic? 

Honestly I didnt want to re-apply on player joined cuz it may apply for those who already did and override some changes, but, I could always add some more logic and keep track of players that already got initialized  to filter out those & make it automatic for everyone

grilledham:
> logistic requests are enabled when logistcs is researched. I do apply saved data at the research completion & for all players joining for the first time after that. But if a player had already joined & left, they will have to manually say /logistic-requests-load upon rejoins to apply that. Call it good, or add more logic?

That's an annoying edge case. I think it probably is worth fixing. I think keeping track of players that have been initialized as you suggest would be the way to go.

features/player_logistic_requests.lua
```diff
@@ -0,0 +1,193 @@
...
+            return
+        end
+
+        local p_name = data.key
+        local player = game.get_player(p_name)
+        if not (player and player.valid )then
+            return
+        end
+
+        if initialized_players[player.name] then
```
grilledham: This might not be the right place for the `initialized_players` check. What if a player runs the `logistic-requests-load` command? I would think we should load the logistic slots from the server even if the player has been initialized as the player might have change their settings since and wants to reset.

features/player_logistic_requests.lua
```diff
@@ -115,28 +126,29 @@ local research_finished =
...
     end
 )
 
 --- Registers the event only when the server is available.
-local function register_player_create()
+local function register_server_start_events()
     if not primitives.server_available then
-        Event.add_removable(defines.events.on_player_created, player_created)
+        Event.add_removable(defines.events.on_player_created, player_joined_or_created)
+        Event.add_removable(defines.events.on_player_joined_game, player_joined_or_created)
```
grilledham: typically you either want the joined or created event as joined is a super set of created in that any time created happens joined happens too. So I think you wanted to switch created to joined rather than add joined.

features/player_logistic_requests.lua
RedRafe: You're right, I didn't account for player resets. Might have to keep the `set = true` here still, but move the early return check into `on_player_joined`

This way:
- tech unlock still sets player initialized
- rejoins will not trigger bar overrides if already initialized
- players can manually reset endless times

Checks out?

features/player_logistic_requests.lua
RedRafe: Yes, on-created is just redundant

features/player_logistic_requests.lua
grilledham: I think I wouldn't call `Server.try_get_data...` unless logistics is unlocked. And in `set_bars_callback` I'd change `initialized_players[player.name] = success` to `initialized_players[player.name] = true`. I think we always want to set this if we have tried, even if we didn't actually set anything.

Change `research_finished` to call `Server.try_get_data...` instead of `load_bars`. Shouldn't print errors to players when logistics is researched if not on the server.

Then the `if initialized_players[player.name] then` check would go in `player_joined_or_created` and `research_finished`

With that I think `load_bars` would only get called by the command, so could add a player_print error if logistics is not researched.

RedRafe:
Updated the code, I think I got it one step further this round:

1. I've added a bit of overhead with `Force` shenanigans, based on this: 
[Force::character_logistic_requests](https://lua-api.factorio.com/latest/classes/LuaForce.html#character_logistic_requests)
> `true` if character requester logistics is enabled.

The new logic tracks initialized forces as well, and only allows setting bar based on that param:
- makes it more robust with modded runs that mess with the logistic tech
- makes it more robust with multi-forces maps

2. I've moved the `player_initialized` logic outside the simple `set_bar_callback` as you suggested, so other functions will be responsible to call a set bars only when it actually possible (based on their own conditions). but players are still able to call it manually whenever they want.

An additional message is displayed (in manual mode) if `server = true` AND `logistics = false`, so players are notified their attempt at setting bars didn't go through cuz their force hasnt unlocked logistics yet

features/player_logistic_requests.lua
```diff
@@ -125,24 +143,34 @@ local player_joined_or_created =
...
         local tech = event.research
-        if not tech or tech.name ~= 'logistic-robotics' then
+        if not (tech and tech.valid) then
             return
         end
 
-        for _, player in pairs(game.connected_players) do
+        -- Already init. or logistics not available
+        local force = tech.force
+        if force and (initialized_forces[force.name] or not force.character_logistic_requests) then
```
grilledham: Shouldn't this be?

```suggestion
        if not force or initialized_forces[force.name] or not force.character_logistic_requests then
```

Previously if `force` was `nil` you could get a nil ref error below for `for _, player in pairs(force.connected_players) do`

features/player_logistic_requests.lua
```diff
@@ -0,0 +1,221 @@
...
+    },
+    function(tbl)
+        primitives = tbl.primitives
+        initialized_players = tbl.initialized_players
+        initialized_forces = tbl.initialized_forces
+    end
+)
+
+--- Scans all player's logistic request slots into a table, then saves that table server-side
+local function save_bars(_, player)
```
grilledham, line 37: Thinking about `save_bars` again. Should we also add a check here for when the player doesn't have logistics available, else the player is actually doing a delete, which might be surprising.

features/player_logistic_requests.lua
RedRafe: Technically, `if force and ..` this will already exit in case of `force = nil`. So the parethesis doesnt get computed (accessing fields `.name` and `.character_logistic_requests`)

I guess this would be the compact version of 
`if (force and initialized_forces[force.name]) or (force and not force.character_logistic_requests) then`

It would return before getting to the `for .. in pairs(force.conn...)` loop

I can rewrite/make it more explicit if you prefer

features/player_logistic_requests.lua
RedRafe: Pulling out the [docs](https://lua-api.factorio.com/latest/classes/LuaControl.html#get_personal_logistic_slot)

`get_personal_logistic_slot(slot_index)  → LogisticParameters`
> `LogisticParameters`: The logistic parameters. If personal logistics are not researched yet, their name will be nil.

I guess.. Practically, they wont have any visual personal logi request tab displayed in game, so IDK why they would think to set it based on what they see(?), but I could add a force check here as well & return before their data gets deleted, for safety. Will add

features/player_logistic_requests.lua
grilledham: Maybe I'm being dumb but:
```lua
if (nil and initialized_forces[force.name]) or (nil and not force.character_logistic_requests) then
    return -- This would be unreachable, which I think is the problem as we should return when force is nil.
end
```

RedRafe: My god yes, you're absolutely right... I shouldn't review logics this late after work. Will change as well sorry ;)

grilledham (approved):
Looks good. Thanks for all your work!

PR #1403, opened 2024-03-17
DO: Silo Defense scenario

RedRafe:
G'day, I'd like to add this new PvE scenario to the DangerOres folder

**Objective**: Launch 666 rockets while defending your own silo against biters in a DO map

**Setting**: regular DO rules apply, with:

- enemies ON
- 20x tech cost
- all landfill tiles
- 666 launches to win
- mods: Deadlock's (ore only), RedMew-data, [Memory Storage](<https://mods.factorio.com/mod/deep-storage-unit>)

## Relevant code changes:

- `map_gen/maps/danger_ores/modules/restart_command.lua` now also prints on Discord if the players won or lose the map. Optional global `map_won_objective` may be present to set win/lose, map is considered victorious by default
- `map_gen/maps/danger_ores/modules/map.lua` added optional config param `enemy_radius` to `map(config)` of DO
- `map_gen/maps/danger_ores/modules/memory_storage_control.lua` added module for Memory Storage management (Memory Units are infinite warehouses). The module simply disables tech/recipe & spawns 4 of these special warehouses at spawn. If a player griefs them & disconnects, their inventory is immediately spilled. Can be used in other scenarios independently

Other than that, everything else is for The *Silo defense* scenario itself:

- spawn callbacks for silo/turrets placement
- listen on biters grouping to direct them towards the silo
- use `rocket_launched` module to do waves on satellites
- disabled terraforming, blue concrete, & biter drops

I'm not 100% sure on how `rocket_launched` works. I've tested on single player & I didn't manage to get the win condition/reset message.

I'm attaching a few screenshots of the map below

## Spawn area
Screenshot from 2024-03-17 14-22-33

## Biter waves
Screenshot from 2024-03-17 14-28-05

## Map lost
Screenshot from 2024-03-17 14-29-48

## Anti grief 
Screenshot from 2024-03-17 15-02-01

### DIFF
--- a/map_gen/maps/danger_ores/modules/memory_storage_control.lua
+++ b/map_gen/maps/danger_ores/modules/memory_storage_control.lua
@@ -1,0 +1,44 @@
+local Event = require 'utils.event'
+local RS = require 'map_gen.shared.redmew_surface'
+
+if not script.active_mods['deep-storage-unit'] then
+  return
+end
+
+-- Disable MU tech & drop 4 of them at spawn
+Event.on_init(function()
+  game.forces.player.technologies['memory-unit'].enabled = false
+
+  local chest = RS.get_surface().create_entity{
+    name = 'iron-chest',
+    position = {0, 0},
+    force = 'player',
+    create_build_effect_smoke = true,
+    move_stuck_players = true
+  }
+
+  chest.insert({ count = 4, name = 'memory-unit'})
+end)
+
+-- If a player disconnects with any Memory Storage in its inventory, spill all of it immediately
+Event.add(defines.events.on_pre_player_left_game, function(event)
+  local player = game.get_player(event.player_index)
+  if not (player and player.valid) then
+    return
+  end
+
+  for _, inv_type in pairs({defines.inventory.character_main, defines.inventory.character_trash}) do
+    local inv = player.get_inventory(inv_type)
+    local count = inv.get_item_count
+
+    if count('memory-unit') + count('memory-unit-with-tags') > 0 then
+      local spill_stack = player.surface and player.surface.spill_item_stack
+      local position = player.position
+
+      for i=1, #inv do
+        spill_stack(position, inv[i], true, nil, false)
+        inv[i].clear()
+      end
+    end
+  end
+end)
--- a/map_gen/maps/danger_ores/presets/danger_ore_silo_defense.lua
+++ b/map_gen/maps/danger_ores/presets/danger_ore_silo_defense.lua
@@ -1,0 +1,301 @@
+local b = require 'map_gen.shared.builders'
+local Config = require 'config'
+local Event = require 'utils.event'
+local MGSP = require 'resources.map_gen_settings'
+local RS = require 'map_gen.shared.redmew_surface'
+local Task = require 'utils.task'
+local Token = require 'utils.token'
+local Global = require 'utils.global'
+local Server = require 'features.server'
+local ShareGlobals = require 'map_gen.maps.danger_ores.modules.shared_globals'
+
+local ScenarioInfo = require 'features.gui.info'
+ScenarioInfo.set_map_name('Danger Ore Silo Defense')
+ScenarioInfo.set_map_description([[
+  Clear the ore to expand the base,
+  focus mining efforts on specific sectors to ensure
+  proper material ratios, expand the map with pollution!
+]])
+ScenarioInfo.add_map_extra_info([[
+  This map is split in three sectors [item=iron-ore] [item=copper-ore] [item=coal].
+  Each sector has a main resource and the other resources at a lower ratio.
+
+  You may not build the factory on ore patches. Exceptions:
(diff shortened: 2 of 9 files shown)


RedRafe:
Obviously I'd like to test it with actual players before adding it. The satellite module I straight up copied from [danger_ores_angelsbobs](https://github.com/Refactorio/RedMew/blob/2efc9b2604c97023622cdc6fe4ce54a80d7c58eb/map_gen/maps/danger_ores/presets/danger_bobangels_ores.lua#L148-L157) just changing the sat count

map_gen/maps/danger_ores/modules/memory_storage_control.lua
```diff
@@ -0,0 +1,44 @@
...
+  for _, inv_type in pairs({defines.inventory.character_main, defines.inventory.character_trash}) do
+    local inv = player.get_inventory(inv_type)
+    local count = inv.get_item_count
+
+    if count('memory-unit') + count('memory-unit-with-tags') > 0 then
+      local spill_stack = player.surface and player.surface.spill_item_stack
+      local position = player.position
+
+      for i=1, #inv do
+        spill_stack(position, inv[i], true, nil, false)
```
grilledham, line 39: I'm not sure if spilling items is the best idea here. Isn't this punishing the other players who now have to clean up the mess.

What if we spawned a player corpse instead. You can set the corpse to inactive to prevent it from despairing. You can look at `dump_offline_inventories.lua` for inspiration. 

Also I wonder if only dumping the memory storage items instead of the whole inventory would make more sense? Logging off with a memory storage item might be accidently rather than deliberate/griefing.

map_gen/maps/danger_ores/presets/danger_ore_silo_defense.lua
```diff
@@ -0,0 +1,301 @@
...
+    this.silo_id = script.register_on_entity_destroyed(entity)
+  end
+
+  if data.callback then
+    Task.queue_task(data.callback, entity)
+  end
+end)
+
+local set_game_lost = Token.register(function()
+  ShareGlobals.data.map_won = true
```
grilledham: Maybe we should rename `map_won` to something like `map_finished`.

map_gen/maps/danger_ores/presets/danger_ore_silo_defense.lua
```diff
@@ -0,0 +1,301 @@
...
+    target = target,
+    speed = 0.5,
+  }
+
+  Task.set_timeout(8, set_game_lost)
+end)
+
+--- Map expansion limited by biters
+-- local terraforming = require 'map_gen.maps.danger_ores.modules.terraforming'
+-- terraforming({ start_size = 8 * 32, min_pollution = 600, max_pollution = 24000, pollution_increment = 9 })
```
grilledham, line 259: Maybe we still want `terraforming` but with a much lower pollution requirements? But I'm willing to try this as is.

map_gen/maps/danger_ores/modules/memory_storage_control.lua
RedRafe: Honestly, I made it harsh on purpose since there are only 4 of the in game, & you can't get others, so it would act both as anti-grief & punishment if you log off with one of them in your inventory, since you're not supposed to have any

(Memory Storages have a scripted inventory so when picking up one Memory Storage, it appends a tag with the stored quantity to the item. So you can have 1000, 1M, 100Billions items stored in 1 inventory slot of yours)

But this may be indeed too harsh, I can tone it down by spilling just the MUs

grilledham (commented):
Haven't tested, but seems mostly fine. if I get time later, I'll try to test the win condition.

map_gen/maps/danger_ores/presets/danger_ore_silo_defense.lua
RedRafe: Yup, that's what I thought. I'll update it & all the maps using it to be more clear

map_gen/maps/danger_ores/presets/danger_ore_silo_defense.lua
RedRafe: My rationale was that both terra forming & biters would absorb pollution, going one against the other.

If terra forming was just meat to control map's byte size, maybe this map can do the same just with biters, but we'll see.
I don't want the players to clear all biters and be "chill" because all the visible chunks don't have biters anymore, so without terra forming there will always be "the next bases" that absorb pollution & send attacks 

I'll also consider reintroducing terra forming if the biters are too strong, as a form of control of biter power

RedRafe:
Updating this based on feedbacks from test runs. Todo before merging (more of a memo to me when I'll come back to this in a week or so):

- [ ] refactor MUControl to not drop full inventory but only MemUnits
- [ ] boss biters instead of big waves to reduce lag?
- [ ] add market back? [Farefang's on Discord](https://discord.com/channels/432567222481846283/821814212945903618/1221081928434978898)
- [ ] boss wave every 5?

Data side:
- [ ] increased turret range?
- [ ] copper ammos?

@ grilled - the rocket launched module worked fine (it start counting after reaching 1 evo, thats why I didn't get it on editor testing) but after 50-60 waves becomes unplayable due to the lag of biters moving towards base. So for higher wave count (>50/100) it needs some boss mechanic maybe so instead of creating 4'000 biters, only a bunch of them are created with better stats. Below a couple images from the last run

do_path
do_path_2

PR #1409, opened 2024-04-10
Add DangerOres Expanse preset

RedRafe:
Add new scenario called DangerOre Expanse

>  DangerOres meets The Expanse!
>
>  Clear the ore to expand the base,
>  focus mining efforts on specific sectors to ensure
>  proper material ratios, expand the map with pollution!
>                            -- AND --
>  Feed the hungry blue chests with the materials they require.
>  The Elder Tree and the Infinity Stone you find at spawn will
>  provide you with all the wood and ores you ever desired.

# Previews

> A (reduced scale) preview of starting area

Screenshot from 2024-04-10 16-31-35

> There are an infinite tree&rock that can be handmined at spawn to prevent resource softlocks

Screenshot from 2024-04-10 16-31-44Screenshot from 2024-04-10 16-32-22

> To unlock new chunks simply feed the chests the requested items (based on available/craftable items in game & amount is based on resource in the chest's chunk)

Screenshot from 2024-04-10 16-32-55

> Once a request is fulfilled, the new chunk is unlocked and a new requester will be created at one of the empty borders. The chest will spill excess items not absorbed and have a 20% chance of giving 1 additional coin. Coins can be used to reroll the chest requests if needed.

Screenshot from 2024-04-10 16-33-49

### DIFF
--- a/map_gen/maps/danger_ores/modules/expanse.lua
+++ b/map_gen/maps/danger_ores/modules/expanse.lua
@@ -32,10 +32,10 @@ local price_modifiers = {
 }
 
 ---@param config
----@field start_size              int, the width/height of the starting area. Will be ceiled to a multiple of a chunk
----@field chance_to_receive_token int, probability of coin rewards [0, 1]
----@field max_ore_price_modifier  int, max value for the ore_modifier after cmputing it with the distance_modifier
----@field price_distance_modifier int, how distance from {0,0} influence the request price
+---@field start_size              number, the width/height of the starting area. Will be ceiled to a multiple of a chunk
+---@field chance_to_receive_token number, probability of coin rewards [0, 1]
+---@field max_ore_price_modifier  number, max value for the ore_modifier after cmputing it with the distance_modifier
+---@field price_distance_modifier number, how distance from {0,0} influence the request price
 return function(config)
   Generate.enable_register_events = false
 
@@ -77,23 +77,6 @@ return function(config)
         end
       end
       surface.set_tiles(tiles, true)
-    end
-  end
-
-  local function reward_tokens(entity)
-    local chance = chance_to_receive_token % 1
-    local count = math.floor(chance_to_receive_token)
-
-    if chance > 0 then
-      chance = math.floor(chance * 1000)
-      if math.random(1, 1000) <= chance then
-        entity.surface.spill_item_stack(entity.position, { name = 'coin', count = 1 }, true, nil, false)
-      end
-    end
-    if count > 0 then
-      for _ = 1, count, 1 do
-        entity.surface.spill_item_stack(entity.position, { name = 'coin', count = 1 }, true, nil, false)
-      end
     end
   end
 
@@ -337,7 +320,9 @@ return function(config)
         end
       end
 
-      reward_tokens(entity)
+      if math.random() <= chance_to_receive_token then
+        entity.surface.spill_item_stack(entity.position, { name = 'coin', count = 1 }, true, nil, false)
+      end
       entity.destructible = true
       entity.die()
       return new_area
--- a/map_gen/maps/danger_ores/modules/expanse.lua
+++ b/map_gen/maps/danger_ores/modules/expanse.lua
@@ -320,7 +320,7 @@ return function(config)
         end
       end
 
-      if math.random() <= chance_to_receive_token then
+      if math.random() < chance_to_receive_token then
         entity.surface.spill_item_stack(entity.position, { name = 'coin', count = 1 }, true, nil, false)
       end
       entity.destructible = true
--- a/map_gen/maps/danger_ores/modules/expanse.lua
+++ b/map_gen/maps/danger_ores/modules/expanse.lua
@@ -1,0 +1,501 @@
+local b = require 'map_gen.shared.builders'
+local Event = require 'utils.event'
+local Generate = require 'map_gen.shared.generate'
+local Global = require 'utils.global'
+local Price_raffle = require 'map_gen.maps.danger_ores.modules.price_raffle'
+local RS = require 'map_gen.shared.redmew_surface'
+local table = require 'utils.table'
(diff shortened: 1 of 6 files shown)


map_gen/maps/danger_ores/modules/expanse.lua
```diff
@@ -0,0 +1,501 @@
...
+  ['water-mud'] = -6,
+  ['water-shallow'] = -6,
+  ['water'] = -5,
+}
+
+---@param config
+---@field start_size              int, the width/height of the starting area. Will be ceiled to a multiple of a chunk
+---@field chance_to_receive_token int, probability of coin rewards [0, 1]
+---@field max_ore_price_modifier  int, max value for the ore_modifier after cmputing it with the distance_modifier
+---@field price_distance_modifier int, how distance from {0,0} influence the request price
```
grilledham: Some of these types should be number? Use int if you are saying only whole numbers are allowed.

map_gen/maps/danger_ores/modules/expanse.lua
```diff
@@ -0,0 +1,501 @@
...
+
+  local function reward_tokens(entity)
+    local chance = chance_to_receive_token % 1
+    local count = math.floor(chance_to_receive_token)
+
+    if chance > 0 then
+      chance = math.floor(chance * 1000)
+      if math.random(1, 1000) <= chance then
+        entity.surface.spill_item_stack(entity.position, { name = 'coin', count = 1 }, true, nil, false)
+      end
```
grilledham: Just a suggestion, but wouldn't this be simpler?

```suggestion
      if math.random() < chance then
        entity.surface.spill_item_stack(entity.position, { name = 'coin', count = 1 }, true, nil, false)
      end
```

Note: `math.random()` without args is [0,1), so the change in comparison operator was deliberate.

map_gen/maps/danger_ores/presets/danger_ore_expanse.lua
```diff
@@ -0,0 +1,150 @@
...
+local Config = require 'config'
+
+local ScenarioInfo = require 'features.gui.info'
+ScenarioInfo.set_map_name('Danger Ore Expanse')
+ScenarioInfo.set_map_description([[
+  DangerOres meets The Expanse!
+
+  Clear the ore to expand the base,
+  focus mining efforts on specific sectors to ensure
+  proper material ratios, expand the map with pollution!
```
grilledham: Does the map expand with pollution?

grilledham (approved):
Nice one! I haven't tested it, but I looked through the code and looks good to me. I do have a few optional suggestion though.

Let me know when you want me to merge.

map_gen/maps/danger_ores/modules/expanse.lua
RedRafe: Ah yes, those are numbers indeed, will adjust 😉

map_gen/maps/danger_ores/presets/danger_ore_expanse.lua
RedRafe: No, indeed I need to update the description I keep recycling from 20 DangerOres maps ago 🤣

map_gen/maps/danger_ores/modules/expanse.lua
RedRafe: Yes. Honestly this part is from the batch of code I blindly copied over and never even looked at since it never caused any issue. But now that I give it a proper look, since I define the chance token to be [0,1], I don't even need to do the modulo operation, allocate count, and check for count > 0. Just `math.random() < token_chance` and get rid of all other 15 lines of bloat

RedRafe:
On my part this could all be merged as I dont have any more features/changes programmed. But if you wanna take your time to test & suggest changes, be my guest, I'm in no rush.

Only issue that makes me think is that the original expanse code

1. didn't take in account research progress (basically L366-396 of `price_raffle.lua` which I added);
2. estimates the expansion cost of each new chunk based on the resources/entities/tiles of the current chunk.

Those 2 make each expansion chest very expensive at my eyes, but it's true that EACH chunk is covered in ore (and there's some penalty/ratio in place already when accounting the cost) so maybe it's not so expensive, it only needs to get accustomed to the scale. It was still possible to get dischargers, nuclears and Mk armors in chests cost even in the starting chests, so my patch to only add recipes once unlocked with research should make it a little smoother and even reduce the need of coins to reroll costs. Will only find out after a full playthrough I guess:)

map_gen/maps/danger_ores/modules/expanse.lua
```diff
@@ -0,0 +1,486 @@
...
+            },
+            right_bottom = {
+              x = p.x - p.x % cell_size + cell_size - 1,
+              y = p.y - p.y % cell_size + cell_size - 1,
+            }
+          }
+        end
+      end
+
+      if math.random() <= chance_to_receive_token then
```
grilledham:
```suggestion
      if math.random() < chance_to_receive_token then
```
The probability is of this happening is basically zero, but `math.random()` can return 0, so without this change a `chance_to_receive_token`  of 0 could still spawn a coin. `math.random()` also return less than 1, so if the chance was set to 1 `math.random() < chance_to_receive_token` would always be true which is what we would want.

grilledham:
> Those 2 make each expansion chest very expensive at my eyes, but it's true that EACH chunk is covered in ore (and there's some penalty/ratio in place already when accounting the cost) so maybe it's not so expensive, it only needs to get accustomed to the scale. It was still possible to get dischargers, nuclears and Mk armors in chests cost even in the starting chests, so my patch to only add recipes once unlocked with research should make it a little smoother and even reduce the need of coins to reroll costs. Will only find out after a full playthrough I guess:)

Balance is hard to get right, you only really find out by testing with real players. I'm happy for players to try this, the worst that can happen is we have to restart early while we make some changes.

PR #1410, opened 2024-04-17
Fixes for incorrect chest behaviors in DO:Expanse

RedRafe:
Scenario changes:
 - `max_ore_multiplier` dropped from 0.33 to 0.20 (meaning 20% resources will be devolved into expansion)
 - `peaceful = true`

Then mainly bugfixes for incorrect chest behaviors & chunk position computation that lead into map locks during 1st run

### DIFF
--- a/map_gen/maps/danger_ores/modules/expanse.lua
+++ b/map_gen/maps/danger_ores/modules/expanse.lua
@@ -353,8 +353,8 @@ return function(config)
     surface.request_to_generate_chunks(left_top, 0)
     surface.force_generate_chunk_requests()
 
-    local max_side = cell_size - RADIUS + 1
-    local min_side = RADIUS - 1
+    local max_side = cell_size - 1 - RADIUS
+    local min_side = RADIUS
     local positions = {
       { x = left_top.x + math.random(min_side, max_side), y = left_top.y }, -- north
       { x = left_top.x, y = left_top.y + math.random(min_side, max_side) }, -- west
--- a/map_gen/maps/danger_ores/modules/expanse.lua
+++ b/map_gen/maps/danger_ores/modules/expanse.lua
@@ -89,7 +89,6 @@ return function(config)
   end
 
   local function get_empty_neighbour_left_top(entity_position)
-    local find_empty_chunk = out_of_map or (out_of_map == nil)
     local vectors = { { -RADIUS, 0 }, { RADIUS, 0 }, { 0, RADIUS }, { 0, -RADIUS } }
     --table.shuffle_table(vectors) --left, right, down, up
 
@@ -435,7 +434,7 @@ return function(config)
     end
 
     if entity.type == 'tree' then
-      for ___, corpse in pairs(surface.find_entities_filtered{
+      for _, corpse in pairs(surface.find_entities_filtered{
         position = entity.position,
         radius = 1,
         type = 'corpse'}
--- a/map_gen/maps/danger_ores/modules/expanse.lua
+++ b/map_gen/maps/danger_ores/modules/expanse.lua
@@ -11,12 +11,13 @@ local Token = require 'utils.token'
 local rendering = rendering
 local math_abs = math.abs
 
-local rock_resources = {
+local RADIUS = 3
+local ROCK_RESOURCES = {
   'iron-ore', 'iron-ore', 'iron-ore', 'iron-ore', 'iron-ore',
   'copper-ore', 'copper-ore', 'copper-ore',
   'coal'
 }
-local price_modifiers = {
+local PRICE_MODIFIERS = {
   ['cliff'] = -128,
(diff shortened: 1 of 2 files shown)


RedRafe:
On previous version, these chests spawning too close to chunk's intersection could be troublesome as they were too close to empty chunks even when invalid. So it's now not possible to spawn chests too close to the intersections, as per defined radius (`RADIUS = 3`) used to search for nearby empty chunks

Screenshot from 2024-04-17 13-36-46

map_gen/maps/danger_ores/modules/expanse.lua
```diff
@@ -406,6 +405,11 @@ return function(config)
   end
 
   local function on_tick()
+    if primitives.index ~= nil and chest_data[primitives.index] == nil then
+      primitives.index = nil
+      return
+    end
+
```
RedRafe, line 412: May need a 2nd pair of eyes on this since it was causing some issues.

While iterating the map `chest_data` during `on_tick`, it could also happen that I remove the current data I'm holding. Which is fine, since the saved `primitives.index` is the index to be used for the following iteration.

But then, the handler for `on_chunk_unlocked` may also remove some data whenever a new chunk unlocks. So if (very rarely, but happened) it removes exactly my saved `primitives.index`, at the next tick, `on_tick` calls `next` on that exactly invalid index, and breaks everything.

So I've added this additional check to see if current index is valid to pass to `next`, otherwise just reset to `nil` as if it had reached the end of the map.

I could probably not remove the data at `on_chunk_unlocked` since chests are still iterated and checked for validity 1 per tick, get rid of this additional check I just added since it shouldn't happen that the `next(.., index)` is invalid, but IDK.

This additional check seems safer now that I know indexes could be messed up, so if in the future something else modifies anything, I should not worry about this part breaking again. So I'd just leave this additional check and still `nil` the data `on_chunk_unlocked` since it's now "allowed and safe"

grilledham: Looking at how I've handled this in the past, see `chunk_list` in `terraforming.lua`, I use the table as a list, rather than a map and check the index is less than the table count. You can quickly remove items from a list in O(1) time using the helper function `table.fast_remove`.

But I think the principle is the same as what you have done here, rather than checking `index < count`, you check the value at the index is not null, before using the value. So I think this should be fine.

grilledham (commented):
I've only looked at the comment your raised. I'll see if I can do a full review tomorrow.

RedRafe:
Yes, looking at how your system is implemented, I think I could change it to that since I don't need the table to stay ordered and I'm actively traversing it rather than reacting to events.
But I would probably need to get rid of the chest removing at `on_chunk_unlocked` since I couldn't go back to the data from the `entity.unit_number` (well, I could simply iterate the list and compare IDs but at that point, I'd just wait for `on_tick` to catch up and do the cleaning itself)

RedRafe:
Yes, looking at how your system is implemented, I think I could change it to that since I don't need the table to stay ordered.
But I would probably need to get rid of the chest removing at `on_chunk_unlocked` since I couldn't go back to the data from the `entity.unit_number` (well, I could simply iterate the lost and compare IDs but at that point, I'd just wait for `on_tick` to catch up and do the cleaning itself)

map_gen/maps/danger_ores/modules/expanse.lua
```diff
@@ -406,6 +405,11 @@ return function(config)
...
     local idx, chest = next(chest_data, primitives.index)
     if not (chest and chest.entity) then
```
RedRafe, line 414: Probably I should nil this entry if it's invalid so to not get stuck iterating the table only up until this point

grilledham:
> But I would probably need to get rid of the chest removing at `on_chunk_unlocked` ...

Why would you need to remove chests when a chunk is unloaded? Unless you can have multiple chests that unlock the same chunk?

My naïve thinking of how this would work is, you would look to add chests when a new chunk is unlocked and remove completed chests on_tick and generate the chunk. (I haven't properly read through the code yet, so I might be missing something obvious.)

RedRafe:
>  Unless you can have multiple chests that unlock the same chunk?

Yes, when 2+ active chunks form an angle, the hidden chunk in between them can be unlocked by any of the adjacent chunks. So when a new chunk is unlocked, I search the nearby chunks for those "obvious invalid chests" to remove them promptly instead of waiting for the `on_tick` cleaner to loop them

RedRafe:
@grilledham can I get an update on this thread? Since this map seems to be the next one coming up on S13, would be better to integrate the 1st round of fixes & eventually leave this open for further improvements/code refactors thanks to the 2nd round of play testing 
Let me know 😄

grilledham:
Ok, let's merge it now.

PR #1414, opened 2024-07-21
New map: Frontier

RedRafe:
Added a new scenario based off the mod [Frontier Extended](https://mods.factorio.com/mod/Frontier-Extended)


grilledham (approved):
I just did a quick skim through to check for anything that would cause issues with the rest of the codebase or anything that would cause a desync and it looks fine.

I'd prefer us to use the `Global` module rather than `global` but can live with it.

We might also want to add this to scenario templates.

Happy to merge if you want me to merge now.

RedRafe:
Yes, code is just an adaptation of the mod's code to make it a scenario. I'd also prefer to port some functions to our modules style. 

I'd say dont merge rn, it's currently being tested on S19, let's see if fixes and other features are needed and I'll tidy things up before actual merge.

resources/price_raffle.lua
grilledham, line 1: I'd prefer `price_raffle.lua` to go in `features` rather than `resources` but not the end of the world if you want it here.

grilledham (commented):
Did a quick skim through and it looks ok to me. I like that you moved the globals to `Global.register`.

Do we want to add this to the scenario templates folder?

RedRafe:
# Changes
- [x] refactored all the code and simplified it a lot (50% of original file with same features supported, most important: `move_silo` now uses the `LuaEntity::clone` API instead of calculating all the different cases...)
- [x] moved DO:Expanse's `price_raffle.lua` to common `resources` since they now both use it
- [x] added loot chests when destroying biter nests
- [x] localised player messages
- [x] moved variables from `global` to our `Global` system
> Note: I've moved many variables inside the new global data table instead of leaving them locally in the file, that's intentional as I'd like to implement a restart/GUI system to manage these variables when restarting/resetting the map to tweak the diffuclty/landscape
- [x] added new map description
- [x] added scenario template for frontier 

I think this would be a good moment to merge (after review), as I plan to add GUI/soft-reset features to this map and I know I'll be changing a lot of code around in other modules as well and I fear it'll get too big to manage in 1 PR alone :) map is being playtested currently ans will be updated via our webinterface to add the new loot feature next restart.

Edit: saw now, will move price raffle to other dir ;)

PR #1415, opened 2024-07-31
Add K2 Maze scenario

RedRafe:
## Changes:
- [x] Added Krastorio2 maze scenario (based off maze scenario already present)
  - tweaks Market's items (if market is present) to add/remove things from vanilla to K2
  - uses Dedalus's mythology as theme for the scenario/description
- [x] Removed the flag `button.enabled = false` for the biter evo button in top gui because even if the button does nothing, setting it to disabled makes it look different from all other buttons
- [x] added a function to switch the score sprite based on mods present: K2 removes that achievement so any scenario with K2 and score gui would crash on load because the sprite is not available anymore.

Here's a preview:
Screenshot from 2024-07-31 10-56-27

Merge is not urgent but this PR is good to go on my part: I've already playtested the scenario multiple times

### DIFF
--- a/features/gui/score.lua
+++ b/features/gui/score.lua
@@ -75,6 +75,13 @@ local function schedule_redraw_score()
     memory.redraw_score_scheduled = true
 end
 
+local function get_modded_sprite()
+    if script.active_mods['Krastorio2'] then
+        return 'achievement/100000-matter'
+    end
+    return 'achievement/there-is-no-spoon'
+end
+
 local function player_created(event)
     local player = game.get_player(event.player_index)
     if not player then
@@ -85,7 +92,7 @@ local function player_created(event)
         {
             type = 'sprite-button',
             name = main_button_name,
-            sprite = 'achievement/there-is-no-spoon',
+            sprite = get_modded_sprite(),
             tooltip = {'score.tooltip'}
         }
     )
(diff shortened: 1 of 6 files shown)


features/gui/score.lua
```diff
@@ -75,6 +75,13 @@ local function schedule_redraw_score()
     memory.redraw_score_scheduled = true
 end
 
+local function get_modded_sprite()
```
grilledham:
```suggestion
local function get_score_sprite()
```

grilledham (approved):
LGTM. Do you want to add it to scenario templates?

RedRafe:
> Do you want to add it to scenario templates?

Mmh, IDK. I don't think so (mainly because `maze.lua` does not have its own template either) but could always add if you prefer, maybe it's useful when launching from interface if we want to play it soon? I'm good either ways

grilledham:
> > Do you want to add it to scenario templates?
> 
> Mmh, IDK. I don't think so (mainly because `maze.lua` does not have its own template either) but could always add if you prefer, maybe it's useful when launching from interface if we want to play it soon? I'm good either ways

The original maze map predates the scenario templates folder and I didn't retroactively add the old maps. There is no cost to adding maps to scenario templates, so I think it makes sense for the maps we are planning to play. Even if we think it's only going to be once, another admin might want to launch the map at a later date and it's easier if they can do that from the web interface rather than having to make a save and upload it.

RedRafe:
Well lets add all mazes now so that's done

PR #1419, opened 2024-08-04
Update Frontier scenario

RedRafe:
# Changes
- [x] added mixed ores to map gen
- [x] added silo explosion effects


grilledham (approved):
LGTM. Though I haven't play tested it, just inspection of code.

RedRafe:
There's a couple of `if player.name == 'grilledham' then ...` around the code, but for all other features they're playtesting this in loop on S19 😆

PR #1420, opened 2024-08-10
Update Frontier to V2

RedRafe:
# Changes
Another round of updates for the Frontier scenario, worth of notice changes to the codebase:
- [x] Added 4 modules in features:
  - auto stash (shortcut): sort inventory into nearby chests
  - battery recharge (shortcut): charges equipment from nearby accumulators
  - clear corpses (shortcut): does the `/corpses` command
  - enemy turrets: manages auto-refill for enemy turrets
- [x] Added `features/gui/shortcuts` (see gif below). Default in config is disabled, but the sub-modules are all "active". So each map that makes use of them will have to auto-manage which to turn off, but players using our scenarios from the github can enable it in config (of not already) and have all the (mostly QoL) features enabled already
- [x] Changed `utils/score_tracker` to have `set` and `reset` endpoint (useful for soft resets of the maps, which frontier now does) 
- [x] Updated frontier with new features
  - Added enemy turrets
  - Added soft reset
  - Added shortcuts gui
  - Added score tracker for rockets to win
  - Deaths no longer contribute to rocket to win, instead, a rng value is rolled at the beginning of the game

shortcutsv2

### DIFF
--- a/.luacheckrc
+++ b/.luacheckrc
@@ -385,6 +385,8 @@ stds.factorio_control = {
                 "reload_script",
                 "remove_offline_players",
                 "remove_path",
+                "reset_game_state",
+                "reset_time_played",
                 "save_atlas",
                 "server_save",
                 "set_game_state",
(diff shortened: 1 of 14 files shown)


grilledham (approved):
Skimmed through it, but all looks reasonable to me.

PR #1426, opened 2024-08-17
Updates/frontier v4

RedRafe:
# Changes

- [x] Tweaked map settings for pollution, resource richness/frequency
- [x] Added random markets around the map (attachment 1)
- [x] Added "Spawn shop" (attachment 2)
- [x] Battery charge module now supports modded accumulators as well

# Issues
- [ ] the soft reset mechanic does not fully work. Which means it works in single-player and on our MP server at low loads, but the server itself will crash on high loads and high player count (tested, the server just takes too much time and shuts off) - (attachment 3: even alone on our server, it goes non-responsive for almost the full bar before catching up and actually loading the map, while if there are many players, it just never catches back and crashes).

These functions are pretty heavy to execute and are the ones triggering in this order when restarting, causing the bottleneck:
> [`Main.on_game_finished()`](<https://github.com/RedRafe/RedMew/blob/b5aab643529222b39b378b09a32a50d7d55aa9d1/map_gen/maps/frontier.lua#L1759-L1769>), [`Main.on_game_started()`](<https://github.com/RedRafe/RedMew/blob/b5aab643529222b39b378b09a32a50d7d55aa9d1/map_gen/maps/frontier.lua#L1707-L1753>) and especially [`Main.reveal_spawn_area()`](<https://github.com/RedRafe/RedMew/blob/b5aab643529222b39b378b09a32a50d7d55aa9d1/map_gen/maps/frontier.lua#L1675-L1686>):

```lua
function Main.reveal_spawn_area()
  local far_left, far_right = this.kraken_distance + this.left_boundary * 32 + 1, this.right_boundary * 32 + this.wall_width
  surface.request_to_generate_chunks({ x = 0, y = 0 }, math.ceil(math_max(far_left, far_right, this.height * 32) / 32))
  surface.force_generate_chunk_requests()

  Main.compute_silo_coordinates(this.silo_starting_x + math_random(100))
  Main.move_silo()
  Terrain.create_wall(this.right_boundary * 32, this.wall_width)

  game.forces.player.chart(surface, { { -far_left - 32, -this.height * 16 }, { far_right + 32, this.height * 16 } })
end
```

I've increased the gap between the surface clear and the new map regen, but wont probably be enough. Maybe I should offload the map reveal to task manager and have it consume it over multiple ticks? Dunno if you have other suggestion on the matter.
S30 testing has a local save `frontier-zombie...` just ~30s prio map reset if you wanna take a look

> Random market with item-to-item trade offers
Screenshot from 2024-08-15 19-58-15

> Spawn shop with upgrades
Screenshot from 2024-08-16 17-22-04

> Restart issue
Screenshot from 2024-08-17 06-45-57


grilledham:
For the .png graphics, can we move them to a mod? I'm a little concerned that those assets will increase the size of all save files even for other scenarios. Maybe 100KBs doesn't seem like much now, but I can imagine that you will want to add more later.

grilledham (commented):
This is great stuff, the market gui looks really cool!

Regarding the soft restart performance issues. I would think it's the `surface.force_generate_chunk_requests()` that expensive. But I see you need to wait for the chunks to generate before you can place the walls and silo. I've found chunk generation to be expensive in the past. My solution has been to think about map generation differently. Instead of creating the chunks and entities up front, I wait for the chunks to generate and creates the entities when the chunk is ready. I also have code in https://github.com/Refactorio/RedMew/blob/develop/map_gen/shared/generate.lua that spreads the calculations over multiple ticks. 

For this though, that probably not going to be an easy change. I wonder if you can remove the `surface.force_generate_chunk_requests()`  and instead wait for the chunks to finish generating. In theory you know which chunks should generate so you can wait to see if they have been generated using the on chunk event. And then do the rest of the startup code afterwards. Or maybe you can generate smaller batches of the map at a time with a timeout in-between.

The other thing that might be expensive is `surface.clear(true)` but I'm not sure what you could do about that other than what you have done which is to have a timeout afterwards.

Also, the file is getting quite large, maybe you should think about splitting it up into different modules. One to think about for the future.

RedRafe:
Thanks! I've postponed terrain gen for so long, it is now time to finally convert it from event-react-driven to the `builder` system tools (I know you're using it for crash site outposts and my first attempts with it were miserable due to math.random stuff. Hopefully I'll get a better understanding and do a better job this time).

assets -> will move
file size -> (I know) I've already divided it multiple sub-modules so when it'll be time to split it I wont have to entangle each part. Only thing holding me back is sharing the global table but I know I'll have to expose it eventually so I'll bite the bullet on this too 😅

RedRafe:
For `surface.clear()` I can maybe offload good part of it to task manager and do `LuaSurface::delete_chunk(position)` on ticks, and then call the clear at T-X_ticks to get rid of any remaining chunks, but I guess `clear` will do some extra crazy stuff so I'll keep it distant from player's comeback on surface too

grilledham (approved):
LGTM. Just to check you want me to merge this now?

RedRafe:
Yes this can go in, I will PR other parts soon ™️

PR #1427, opened 2024-08-17
Tidy up Frontier scenario into smaller files

RedRafe:
Continuation of #1426 before (future) #1428


map_gen/maps/frontier/modules/enemy.lua
```diff
@@ -0,0 +1,402 @@
...
+    name = 'artillery-projectile',
+    position = position,
+    force = 'enemy',
+    source = position,
+    target = { x = position.x + math_random(-r, r) + math_random(), y = position.y + math_random(-r, r) + math_random() },
+    max_range = 1,
+    speed = 0.1
+  }
+end
+Enemy.artillery_explosion_token = Token.register(Enemy.artillery_explosion)
```
grilledham: Are you missing a `return Enemy` at the bottom of this file?
I see that you require this file in `scenario.lua` but not sure how that works if `enemy.lua` doesn't return anything?

grilledham (commented):
Had a quick look through. This does look better, so thanks for doing this. It's only the comment about `enemy.lua` not returning anything that I'm unsure about.

RedRafe:
Yes, I just forgot that line (didnt noticed it because I'm already on other branch with terrain optimization). Thanks!

map_gen/maps/frontier/modules/enemy.lua
RedRafe: fixed

PR #1428, opened 2024-08-18
Fix restart bottlenecks

RedRafe:
# Changes
- [x] 1. `local this = Public.get` moved inside events instead of local in file (otherwise it was nil after `on_load`)
- [x] 2. added a Queue system for initial map reveal
- [x] 3. moved some functions from to the map function itself

Altho for 3. more stuff can be moved to map function, the main in-game UPS issue is from biters pathfinding so I'll maybe come back and refactor this part when I'll add more scenery (rocks, rivers, ...) to terrain gen. 
I've tested both SP and MP on our server and can confirm .2 (+ another 10s from previous surface clear) solved all the restart issues: now the server does not even hiccup a second when restarting :)


grilledham (approved):
LGTM

PR #1430, opened 2024-08-21
Drop players armor and weapons

RedRafe:
# Changes
- [x] fixed typo in configs `offline_timout_mins` --> `offline_timeout_mins`
- [x] added config to drop player's armor and weapons alongside player's inventory when logging off

This issue was raised on Admins chat to retrieve player's armor/weapons when players are banned, and also when players disconnect hoarding a lot of stuff. So the new config has a field `startup_gear_drop_hours = 24` which means:

- ANY player logging off in the first 24h will drop their armor alongside inventory/weapons
- Rank::Regular and above will NOT drop their armor after first 24h of game (but they will still drop inventory (+weapons) as usual.

```lua
if not banned and player.connected then return end
```
I removed this check because it was not clear to me what edge case it was supposed to gate. The new conditions appeared to be always true so it seemed redundant, but happy to revert it.
Also I added a debug option to drop player's stuff in debug mode, but could be too much overhead so it can be removed if we prefer... I've tested it in all the cases (except ban) and it's unlikely I'll need to retest this often

> screenshot of the changes
Screenshot from 2024-08-21 11-14-33

### DIFF
--- a/features/dump_offline_inventories.lua
+++ b/features/dump_offline_inventories.lua
@@ -1,12 +1,18 @@
 -- This feature allows you to turn on anti-hoarding so that X minutes after a player leaves the game
 -- the resources in their inventory are returned to the teams. A corpse will spawn on the player's last
 -- position and remain until they log back in to claim it or someone else mines it.
+-- All players will drop their armors and weapons during the first 24h of the game,
+-- after this time, only regulars and above will keep their armor and just drop the inventory.
 local Event = require 'utils.event'
 local Task = require 'utils.task'
 local Token = require 'utils.token'
 local Global = require 'utils.global'
 local CorpseUtil = require 'features.corpse_util'
 local Config = require 'config'
+local Rank = require 'features.rank_system'
+local Ranks = require 'resources.ranks'
+local MINS_TO_TICKS = 60 * 60
+local HOUR_TO_TICKS = MINS_TO_TICKS * 60
 
 local set_timeout_in_ticks = Task.set_timeout_in_ticks
 local config = Config.dump_offline_inventories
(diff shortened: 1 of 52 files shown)


features/dump_offline_inventories.lua
```diff
@@ -22,59 +28,54 @@ local function spawn_player_corpse(player, banned, timeout_minutes)
     local player_index = player.index
     offline_player_queue[player_index] = nil
 
-    if not banned and player.connected then
```
grilledham: I think this check was to cover the case were the player re-connects after the timer has been started. Which we don't care about if they were banned as we don't go through the timer in that case.

If you agree this is needed, I think we should add this check back.

features/dump_offline_inventories.lua
```diff
@@ -144,7 +145,12 @@ Event.add(defines.events.on_pre_player_left_game, function(event)
         return
     end
 
-    start_timer(event, config.offline_timout_mins)
+    if _DEBUG then
```
grilledham: I think the debug code would make more sense in `start_timer`. I assume it's really the `set_timeout_in_ticks` you are trying to avoid for testing purposes. And if it's for testing it makes sense to run as much of the real code as possible.

Maybe something like this in `start_timer` would be better:
```lua
local data = {player = player, tick = tick, timeout_minutes = timeout_minutes}
if _DEBUG then
    local spawn_player_corpse_func = Token.get(spawn_player_corpse_token)
    spawn_player_corpse_func(data)
else
   set_timeout_in_ticks(timeout, spawn_player_corpse_token, data)
end
```

features/dump_offline_inventories.lua
```diff
@@ -22,59 +28,54 @@ local function spawn_player_corpse(player, banned, timeout_minutes)
...
+        defines.inventory.character_guns,
+        defines.inventory.character_ammo,
+        defines.inventory.character_vehicle,
+        defines.inventory.character_trash,
+    }
 
-    local inv_corpse_size = 0
-    if inv_main_contents then
-        inv_corpse_size = inv_corpse_size + (#inv_main - inv_main.count_empty_stacks())
+    if banned or game.tick < config.startup_gear_drop_hours * HOUR_TO_TICKS or Rank.less_than(player.name, Ranks.regular) then
```
grilledham: Just because I don't like having to think about order of operations when reading code.
```suggestion
    if banned or game.tick < (config.startup_gear_drop_hours * HOUR_TO_TICKS) or Rank.less_than(player.name, Ranks.regular) then
```

grilledham (commented):
I think this is mostly fine and a welcome improvement. As this is functionally used by multiple maps I want to play test it before I approve. And I want to wait for your thoughts on my comments before I play test.

features/dump_offline_inventories.lua
RedRafe: Np

RedRafe:
So the biggest issue for me was testing this feature, because I had to start a SP game, save it, load it to our server, startup it and then join/leave. And it's tricky because in SP when you exit the game it gets paused, while in MP it register your leave. So my SP leave would actually trigger the 1st boot of the game on server side, and then trigger regularly when joining/leaving server as usual. The issue is that when the server boots up and the leave triggers, `player.connect` would be `false`, while when joining/leaving a running server, since it runs on PRE-player leave, that check sees `player.connect = true`.

This is just a big mess to correctly debug, and that's why I left the debug code in it (which I usually dont), BUT I dont like to leave it on, because it makes the actual working logic so much more complex to maintain/read IMO. Maybe I'd prefer to leave it in code as commented (so could be temporarily de-commented if we need to refactor/debug in the future cuz it would be helpful to have it there and not figuring out this big loophole again), or maybe figure a better workflow that would work all round. I also dont want it to run on `_DEBUG = true` cuz we unlikely will want to drop our invs when testing other stuff around the maps.

Maybe was just easier to leave the timeout thing as is and just speed up to 100x from console after leaving to simulate it (or reduce time to <1m), but yeah... didnt think of it before

RedRafe:
Yes I forgot to mention in the refactor I wanted scenarios to be able to set timeout = 0 to let invs drop instantly when leaving and prev version would not allow it (I guess a workaround would be to set it to 1/60 or whatever <1 min so it just happens next tick or very shortly still which is good enough)

features/dump_offline_inventories.lua
RedRafe: > cover the case were the player re-connects after the timer has been started

Ah that might be it, thanks! Reverting

features/dump_offline_inventories.lua
RedRafe: OKay I've decided I dont care about debug mode, and If we'll have to debug, just reduce the waiting time in configs or wait/speed up game if needed. So removed all the debug junk

RedRafe:
@grilledham reverted the check for online players; and removed the debug stuff: I dont think that is necessary to keep it and would just make debugging other scenarios tedious.

grilledham (approved):
Made some changes to set a default values just in case. I've tested and looks good. Thanks for doing this.

PR #1431, opened 2024-08-26
Updates/frontier v6

RedRafe:
Round of updates for the Frontier scenario

- [x] added artillery turrets to the Enemy turrets feature (will auto target players and structures when registering such turret)
- [x] new terrain generation
- [x] fixed bugs with silo positioning on invalid tiles and filling fluid enemy turrets

### DIFF
--- a/features/enemy_turret.lua
+++ b/features/enemy_turret.lua
@@ -1,23 +1,74 @@
 local Event = require 'utils.event'
 local Global = require 'utils.global'
+local Task = require 'utils.task'
+local Token = require 'utils.token'
 local register_on_entity_destroyed = script.register_on_entity_destroyed
 
 local Public = {}
 local turrets_map = {}
 local register_map = {}
 local primitives = { index = nil }
+local Artillery = {
+  enabled = true,
+  turret_name = 'artillery-turret',
+  character_projectile = 'artillery-projectile',
+  structure_projectile = 'rocket',
+  target_force = 'player',
+  last_fire = 0,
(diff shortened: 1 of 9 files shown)


RedRafe:
There are bits commented out, especially in the debug modules for terrain gen. I'm not in a hurry to merge this but I wanted to get it out as next updates will include some refactors/additions in other modules outside Frontier map.
So maybe this could also be considered a draft before I come back to it, just because I prefer to keep the code on git updated with the one currently running on serve, or at least have a copy of it for reference.

features/enemy_turret.lua
```diff
@@ -75,6 +134,65 @@ function Public.reset()
...
+
+  if #targets == 0 then
+    return
+  end
+
+  for i = 1, #targets do
+    local target = targets[i]
+    if distance(source_position, target.position) > Artillery.min_range then
+      Task.set_timeout_in_ticks(Artillery.fire_interval, artillery_projectile_token, {
+        surface_index = surface.index,
```
grilledham: Minor, but you could pass the surface rather than the surface.index.

grilledham (approved):
I'd be happy to merge in it's current form. But I'll wait for you to say you are ready.

features/enemy_turret.lua
RedRafe: Yup, thanks

RedRafe:
Tidied up a little. This one can go in now

grilledham (approved):

PR #1432, opened 2024-08-28
Add admin panel

RedRafe:
# Changes
- [x] added config & top button for admin panel (Mk2 armor icon)
- [x] admin commands file now also exports its command functions to be used elsewhere, by providing additional param `{ actor = LuaPlayer }` if needed
- [x] updated missing handlers/rules in `utils.gui` and `luackeckr`

# Main features
Added 3 pages in admin panel: Player manager, Map manager, Lua console. Scenarios can also add more pages through the AdminPanel API, planning on adding the history of player actions to help admins moderate maps.

### Player manager
- General actions: cheat mode, create pools, delete stuff... basically shortcuts from already existing console commands for admins
- Player management: promote/demote/jail/invoke/goto players. requires to select another player from the dropdown selector (can filter for online/all/offline players)
- Players ban: helps building the ban message by providing few common reasons, the the admin can further customize the text by editi it themselves before banning.
> screenshot of the player manager panel
Screenshot from 2024-08-28 11-02-43

### Map manager
- Just a bunch of button-slider pairs to tweak map settings/enemies/pollution/evo. Label on the right shows the current map setting, while sliders have a tooltip that shows the "selected" value on the slider. Wont be applied until the user presses the button.
> screenshot of the map manager
Screenshot from 2024-08-28 11-02-49

### Lua console
- I dont expect many admins to use this, but writing commands in the console is shit because it's just 1 line high, so this is more a QoL gift for us devs :) (will probably remain at bottom of all future pages added)
> screenshot of the lua console panel
Screenshot from 2024-08-28 11-02-56

# Future development:
- maybe add some tooltips here & there for buttons based on feedbacks. Could also have each button tooltips that is a "shortcut" for a command, be the description of the command itself and all the parameters that go into it (if an admin needs to manually run the command with specific parameters, idk...)
- add the history of player actions panel. Still need to think on how to actually make it. Tracking players actions is straightforward, displaying the information in a useful/organized way instead of a giant wall of text/giant table of data is not

### DIFF
--- a/features/admin_commands.lua
+++ b/features/admin_commands.lua
@@ -24,8 +24,8 @@ Global.register(
 )
 
 --- Informs the actor that there is no target. Acts as a central place where this message can be changed.
-local function print_no_target(target_name)
-    Game.player_print({'common.fail_no_target', target_name}, Color.fail)
+local function print_no_target(target_name, player)
+    Game.player_print({'common.fail_no_target', target_name}, Color.fail, player)
 end
 
 local function know_player_or_rerun(player_name, actor, command_name)
@@ -38,7 +38,7 @@ local function know_player_or_rerun(player_name, actor, command_name)
         return true;
     end
 
-    Game.player_print({'common.rerun_no_target', player_name}, Color.fail)
+    Game.player_print({'common.rerun_no_target', player_name}, Color.fail, actor)
     return false
 end
 
@@ -57,11 +57,11 @@ end
 --- Toggles cheat mode for a player
 local function toggle_cheat_mode(_, player)
     player.cheat_mode = not player.cheat_mode
-    Game.player_print({'admin_commands.toggle_cheat_mode', tostring(player.cheat_mode)})
+    Game.player_print({'admin_commands.toggle_cheat_mode', tostring(player.cheat_mode)}, nil, player)
 end
 
 --- Promote someone to regular
-local function add_regular(args)
+local function add_regular(args, player)
     local target_name = args.player
     local maybe_target_player = game.get_player(target_name)
     local actor = args.actor or Utils.get_actor()
@@ -71,7 +71,7 @@ local function add_regular(args)
     end
 
     if Rank.less_than(target_name, Ranks.guest) then
-        Game.player_print({'admin_commands.regular_add_fail_probation'}, Color.fail)
+        Game.player_print({'admin_commands.regular_add_fail_probation'}, Color.fail, player)
         return
     end
 
@@ -82,12 +82,12 @@ local function add_regular(args)
             maybe_target_player.print({'admin_commands.regular_add_notify_target'}, Color.warning)
         end
     else
-        Game.player_print({'admin_commands.regular_add_fail', target_name, Rank.get_player_rank_name(target_name)}, Color.fail)
+        Game.player_print({'admin_commands.regular_add_fail', target_name, Rank.get_player_rank_name(target_name)}, Color.fail, player)
     end
 end
 
 --- Demote someone from regular
-local function remove_regular(args)
+local function remove_regular(args, player)
     local target_name = args.player
     local maybe_target_player = game.get_player(target_name)
     local actor = args.actor or Utils.get_actor()
@@ -104,12 +104,12 @@ local function remove_regular(args)
         end
     else
         local rank_name = Rank.get_player_rank_name(target_name)
-        Game.player_print({'admin_commands.regular_remove_fail', target_name, rank_name}, Color.fail)
+        Game.player_print({'admin_commands.regular_remove_fail', target_name, rank_name}, Color.fail, player)
     end
 end
 
 --- Put someone on probation
-local function probation_add(args)
+local function probation_add(args, player)
     local target_name = args.player
     local maybe_target_player = game.get_player(target_name)
     local actor = args.actor or Utils.get_actor()
@@ -119,7 +119,7 @@ local function probation_add(args)
     end
 
     if Rank.equal(target_name, Ranks.admin) then
-        Game.player_print({'admin_commands.probation_add_fail_admin'}, Color.fail)
+        Game.player_print({'admin_commands.probation_add_fail_admin'}, Color.fail, player)
         if maybe_target_player then
             maybe_target_player.print({'admin_commands.probation_warn_admin', actor}, Color.warning)
         end
@@ -133,12 +133,12 @@ local function probation_add(args)
             maybe_target_player.print({'admin_commands.probation_add_notify_target'}, Color.warning)
         end
     else
-        Game.player_print({'admin_commands.probation_add_fail', target_name}, Color.fail)
+        Game.player_print({'admin_commands.probation_add_fail', target_name}, Color.fail, player)
     end
 end
 
 --- Remove someone from probation
-local function probation_remove(args)
+local function probation_remove(args, player)
     local target_name = args.player
     local maybe_target_player = game.get_player(target_name)
     local actor = args.actor or Utils.get_actor()
@@ -154,7 +154,7 @@ local function probation_remove(args)
             maybe_target_player.print({'admin_commands.probation_remove_notify_target'}, Color.warning)
         end
     else
-        Game.player_print({'admin_commands.probation_remove_fail', target_name}, Color.fail)
+        Game.player_print({'admin_commands.probation_remove_fail', target_name}, Color.fail, player)
     end
 end
 
@@ -169,7 +169,7 @@ local function jail_player(args, player)
     local target = Utils.validate_player(target_ident)
 
     if not target then
-        print_no_target(target_ident)
+        print_no_target(target_ident, player)
         return
     end
 
@@ -182,7 +182,7 @@ local function unjail_player(args, player)
     local target = Utils.validate_player(target_ident)
 
     if not target then
-        print_no_target(target_ident)
+        print_no_target(target_ident, player)
         return
     end
     Report.unjail(target, player)
@@ -199,6 +199,7 @@ local function pool(_, player)
     end
     player.surface.set_tiles(t)
     player.surface.create_entity {name = 'fish', position = {p.x + 0.5, p.y + 5}}
+    Game.player_print({'admin_commands.create_pool'}, Color.success, player)
 end
 
 --- Takes a target and teleports them to player
@@ -207,13 +208,13 @@ local function invoke(args, player)
     local target = Utils.validate_player(target_ident)
 
     if not target then
-        print_no_target(target_ident)
+        print_no_target(target_ident, player)
         return
     end
 
     local pos = player.surface.find_non_colliding_position('character', player.position, 50, 1)
     if not pos then
-        Game.player_print({'admin_commands.invoke_fail_no_location'})
+        Game.player_print({'admin_commands.invoke_fail_no_location'}, player)
         return
     end
     target.teleport({pos.x, pos.y}, player.surface)
@@ -226,7 +227,7 @@ local function teleport_player(args, player)
     local target = Utils.validate_player(target_ident)
 
     if not target then
-        print_no_target(target_ident)
+        print_no_target(target_ident, player)
         return
     end
 
@@ -234,27 +235,27 @@ local function teleport_player(args, player)
     local surface = target.surface
     local pos = surface.find_non_colliding_position('character', target.position, 50, 1)
     if not pos then
-        Game.player_print({'admin_commands.tp_fail_no_location'})
+        Game.player_print({'admin_commands.tp_fail_no_location'}, Color.fail, player)
         return
     end
     player.teleport(pos, surface)
     game.print({'admin_commands.tp_player_announce', target_name})
-    Game.player_print({'admin_commands.tp_player_success', target_name})
+    Game.player_print({'admin_commands.tp_player_success', target_name}, Color.success, player)
 end
 
 --- Takes a selected entity and teleports player to it
 local function teleport_location(_, player)
     if not player.selected then
-        Game.player_print({'admin_commands.tp_ent_fail_no_ent'})
+        Game.player_print({'admin_commands.tp_ent_fail_no_ent'}, Color.fail, player)
         return
     end
     local pos = player.surface.find_non_colliding_position('character', player.selected.position, 50, 1)
     if not pos then
-        Game.player_print({'admin_commands.tp_fail_no_location'})
+        Game.player_print({'admin_commands.tp_fail_no_location'}, Color.fail, player)
         return
     end
     player.teleport(pos)
-    Game.player_print({'admin_commands.tp_end_success'})
+    Game.player_print({'admin_commands.tp_end_success'}, Color.success, player)
 end
 
 --- If a player is in the tp_players list, remove ghosts they place and teleport them to that position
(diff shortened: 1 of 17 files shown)


features/gui/admin_panel/player_manager.lua
```diff
@@ -0,0 +1,323 @@
...
+end)
+
+Gui.on_click(on_add_regular, function(event)
+  local target_name = get_selected_player(event.player)
+  Actions.regular_add({ player = target_name, actor = event.player.name })
+end)
+
+Gui.on_click(on_add_probation, function(event)
+  local target_name = get_selected_player(event.player)
+  Actions.probation_add({ player = target_name, actor = event.player.actor })
```
RedRafe: I need to change this to `event.player.name`

features/gui/admin_panel/map_manager.lua
```diff
@@ -0,0 +1,334 @@
...
+local on_evolution_destroy_factor= Gui.uid_name()
+local on_evolution_time_factor = Gui.uid_name()
+local on_map_chart = Gui.uid_name()
+local on_map_hide = Gui.uid_name()
+local on_map_reveal = Gui.uid_name()
+local on_map_rechart = Gui.uid_name()
+
+local this = {}
+
+Global.register(this, function(tbl) this = tbl end)
```
RedRafe: Those may be never used so I could remove

features/gui/admin_panel/core.lua
```diff
@@ -0,0 +1,202 @@
...
+  end
+end
+
+Event.add(defines.events.on_player_created, function(event)
+  local player = game.get_player(event.player_index)
+  if not (player and player.valid) then
+    return
+  end
+  Public.update_top_button(player)
+end)
```
grilledham, line 170: Just my paranoia, but could we add an `on_player_joined` event that:
- Sets the main_button's visibility based on if the player is an admin.
- Closes the main frame if it is open.

features/gui/admin_panel/functions.lua
```diff
@@ -0,0 +1,160 @@
...
+  local msg = 'Saving map: _autosave-' .. filename ..'.zip'
+  if player and player.valid then
+    player.print(msg)
+  else
+    game.print(msg)
+  end
+end
+
+---@param player? LuaPlayer
+function Actions.delete_all_blueprints(player)
```
grilledham: Delete ghosts might be a better name and for the button caption too. At first based on the name I thought this deleted blueprints from the library.

features/gui/admin_panel/functions.lua
```diff
@@ -0,0 +1,160 @@
...
+  local msg = count .. ' speakers removed from all surfaces.'
+  if player and player.valid then
+    player.print(msg)
+  else
+    game.print(msg)
+  end
+end
+
+---@param player? LuaPlayer
+function Actions.kill_all_units(player)
```
grilledham:
```suggestion
function Actions.kill_all_enemy_units(player)
```
The button captions are fine.

features/gui/admin_panel/lua_console.lua
```diff
@@ -0,0 +1,116 @@
...
+end
+
+local function process_command(event)
+  local player = event.player
+  local data = Gui.get_data(event.element)
+  local input, output = data.input, data.output
+
+  local cmd = input.text
+  this.last_lua_input[player.index] = cmd
+  cmd = cmd:gsub('game%.player%.', 'game.players[' .. player.index .. '].')
```
grilledham: In the past I solve this problem using `rawset(game, 'player', player)` see https://github.com/Refactorio/RedMew/blob/develop/features/gui/debug/model.lua#L113
Not sure if my or your hack is better though, so I'll leave it up to you.

features/gui/admin_panel/lua_console.lua
```diff
@@ -0,0 +1,116 @@
...
+  type = 'sprite-button',
+  sprite = 'utility/scripting_editor_icon',
+  tooltip = '[font=default-bold]Lua console[/font]',
+  name = main_button_name,
+  auto_toggle = true,
+}
+
+local function draw_gui(player)
+  local canvas = AdminPanel.get_canvas(player)
+  canvas.clear()
```
grilledham:
```suggestion
  Gui.clear(canvas)
```
Basically whenever you use `Gui.set_data` you need to use `Gui.destroy` and `Gui.clear` instead of `destory` and `clear` else you get a memory leak.

grilledham: I see there are other places where you have used `.clear()` so will need to change those too.

features/gui/admin_panel/map_manager.lua
```diff
@@ -0,0 +1,329 @@
...
+  end
+
+  local data = Gui.get_data(element)
+  data.slider.tooltip = string.format(data.format, data.slider.slider_value)
+end)
+
+Gui.on_click(on_slow_down, function(event)
+  local element = event.element
+  local value = update_slider(element)
+  game.speed = value
```
grilledham, line 287: We could consider using https://github.com/Refactorio/RedMew/blob/d109a6606b9d4168dc5d56cbeb71912c2ceaed7a/features/performance.lua#L38 instead. `Performance.set_time_scale` also modifies character stats so the game doesn't feel as slow. 

Though an argument could be made that you wouldn't want that behaviour so it's not so clear cut. My concern is that  if we offer an easy way to alter the game speed through the admin panel, admins might preferer that to using the command `performance-scale-set`.

features/gui/admin_panel/player_manager.lua
```diff
@@ -0,0 +1,323 @@
...
+-- local on_teleport = Gui.uid_name()
+-- local on_destroy_selected = Gui.uid_name()
+
+local this = {
+  ---@type table<number, table< index: number, name: string >>
+  player_selection = {},
+  ---@type table<number, string>
+  selection_switch = {},
+  ---@type table<number, table<string>>
+  ban_items = {},
```
grilledham:
```suggestion
  player_ban_items = {},
```
A bit confusing that this has the same name as `local ban_items` below.

features/gui/admin_panel/lua_console.lua
grilledham: By the way you can check for memory leaks, by seeing what gui data is active for each player using the `/debug` command and then under Gui Data.

features/gui/admin_panel/core.lua
```diff
@@ -0,0 +1,202 @@
...
+function Public.get_pages()
+  return pages
+end
+
+function Public.get_canvas(player)
+  return Gui.get_data(Public.get_main_frame(player)).right
+end
+
+function Public.get_main_frame(player)
+  local frame = player.gui.screen[main_frame_name]
```
grilledham, line 48: I think we need to add screen here: https://github.com/Refactorio/RedMew/blob/develop/utils/gui_tests.lua#L11

You can run the tests with `/test-runner` if _DEBUG mode is on.

grilledham (commented):
Had a play around, looks good. Just some comments to address.

features/gui/admin_panel/lua_console.lua
RedRafe: Whoa 😮 I didnt know we had such an amazing tool! Definitely need to learn more about all our debug options 
Screenshot from 2024-09-01 10-54-16

features/gui/admin_panel/map_manager.lua
RedRafe: Sure, performace_set is just a fancy game.speed. 

Personally I use the perf command a lot since I've discovered it because the speed boost is cool VS just slowing down the game. I had in mind to use perf_set when tweaking game speed, but then I forgot. 

One thing I'll say about the perf command (that maybe needs a separate issue/PR to discuss), is that the perf command is generally run lategame when base expands a lot and there's a ton of combat happening. Not much in the early game. So by the time the speed change is needed, players already have a full armor with tons of legs, so the speed boost gained back from the perf. command is almost irrelevant (i.e. given a +500% speed bonus from exos, setting perf speed at 80% will give back only 12% of the speed players lost, so at 80% perf speed players will run as 82.5% speed effectively). And craft/mining modifiers are also quite irrelevant after bots... So going through performance module means remembering to set/reset game speed through the module before changing speeds anywhere, and sometimes if game.speed is changed outside the perf module the modifiers artifact of the module persist because there's no event notifying it.

Couple solutions for this could be:
- have separated buttons in admin panel for: increase/decrease just game speed, AND button/slider to it through performance (they also have different range so I'd rather have a set slider for it that only goes 0.5-1.0 speed)
- maybe add a `on_nth_tick` handler to the performance module, that once every like 5 mins polls the game speeds and re-applies the force modifiers so if anything changes game speed (even other mods), we're able to patch those events

Bonus: maybe disable the speed garbage collector or add extra command for admins/debug mode so I can have game speed and walk speed x20 when debugging?

RedRafe:
The map manager is already quite useful to tweak game speed/map reveals/enemy settings while debugging. So we can also think to implement more debugging features, or even adding a separate page (only visible to admin panel in debug mode) with shortcuts of debug tools/scripts that we usually run.

RedRafe:
Made it a separate slider with its command tooltip as well to make it clear
Screenshot from 2024-09-01 12-03-16

features/gui/admin_panel/functions.lua
```diff
@@ -118,6 +119,17 @@ end
...
 
+---@param scale number
+function Surface.performance_scale_set(scale)
+  Performance.set_time_scale(scale)
+  local p = game.print
+  local stat_mod = Performance.get_player_stat_modifier()
+  p({'performance.stat_preamble'})
+  p({'performance.generic_stat', {'performance.game_speed'}, string.format('%.2f', Performance.get_time_scale())})
+  local stat_string = string.format('%.2f', stat_mod)
+  p({'performance.output_formatter', {'performance.game_speed'}, stat_string, {'performance.manual_mining_speed'}, stat_string, {'performance.manual_crafting_speed'}, stat_string})
```
grilledham: I assume this should be running_speed.
```suggestion
  p({'performance.output_formatter', {'performance.running_speed'}, stat_string, {'performance.manual_mining_speed'}, stat_string, {'performance.manual_crafting_speed'}, stat_string})
```

RedRafe: Fixed in the module as well

grilledham:
Testing again, I notice that when you ban a player, it says that the player was banned by script.
image
This concerns me as we keep track of which admin did the ban at https://refactorio.de/admin/bans and I wouldn't like to lose that. I'm not sure what the solution to this is.

grilledham:
Actually this might jus be coming from `text[#text + 1] = ' was banned by <script>**\\n'` in `Report.ban_player`. I'll investigate further.

RedRafe:
Yeah I thought it was strange too when testing. I looked at https://github.com/Refactorio/RedMew/blob/d109a6606b9d4168dc5d56cbeb71912c2ceaed7a/features/report.lua#L436 and there was no "admin" player in the code so I just assumed bans always had `<server>` as actor. I actually never banned myself a player in game, only through our interface, so I dont know if there's another route that shows the actor of the ban? Actor of the ban should be displayed if done by command/buttons (probably need to add it as 3rd param in the Report module and if not available, default to `<script>` just like other commands)

grilledham:
In game we normally ban using the `/ban` command that picks up the admin automatically. The problem is that https://lua-api.factorio.com/latest/classes/LuaGameScript.html#ban_player doesn't let you pass in the admin. I'm not sure how we can work around this. I tried the `rawset(game, 'player', admin)` trick, but it didn't work in this case 🙁.

image

features/admin_commands.lua
```diff
@@ -472,3 +472,19 @@ Command.add(
...
+    invoke_player = invoke,
+    jail_player = jail_player,
+    probation_add = probation_add,
+    probation_remove = probation_remove,
+    regular_add = add_regular,
+    regular_remove = remove_regular,
+    revive_ghosts = revive_ghosts,
+    show_reports = show_reports,
+    teleport_command = teleport_command,
+    toggle_cheat_mode = toggle_cheat_mode,
```
grilledham, line 492: Any of these commands that use `Game.player_print` are slightly problematic. `toggle_cheat_mode` is a good example. `Game.player_print` assumes the player is running a command and is able to source the player form `game.player`. We might have to change `Game.player_print` to have a player arg. Then the useful thing `Game.player_print` does is if player is `nil` it uses `print` rather than `player.print`.

features/gui/admin_panel/map_manager.lua
grilledham: I think what you have done with the performance toggle is good for now. The wider problems with the perf command and scenario/mod code that also changes those modifies is complicated and I'm not sure if really solvable. By making it an admin command we put the responsibility on the admin to use responsibly 😄 ... but they never consider it anyway and just assumes it works. 

In `player_create.lua` we do some useful stuff for testing when `_CHEATS` is `true`. Though it's not the same as being able to toggle it at will.

features/admin_commands.lua
RedRafe: ~~@grilledham  Maybe we could do `raise_event(defines.events.on_console_command, ...)` for all the button-commands? Maybe this would solve the ban command `by <script>` issue AND spare us to add a 3rd parameter in many parts of the code.~~

Edit: can't be done

RedRafe: Okay I've refactored those functions and all nearby function calls to take optional parameter `player <LuaPlayer?>` in case their method is not called by command but by function. I've tested both in admin panel buttons and console commands, but Would appreciate a second review on this last commit because it touches different parts of the codebase.

features/report.lua
```diff
@@ -448,10 +449,16 @@ function Module.ban_player(player, reason)
 
     local server_id = Server.get_server_id()
     local server_name = Server.get_server_name()
+    local actor_name = '<server>'
```
grilledham: I would prefer to keep this `'<script>'`. Factorio uses server when the ban comes from the console, and script when the ban comes from scripts. As we have auto ban script, scripts makes more sense.

Same for `kick_player` below.

features/report.lua
```diff
@@ -476,7 +483,8 @@ end
 --- kicks the player and reports the kick to moderation log channel.
 -- @param  player<LuaPlayer>
 -- @param  reason<string?> defaults to empty string.
-function Module.kick_player(player, reason)
+-- @param actor <LuaPlayer?> the player carrying the ban action, if any
```
grilledham:
```suggestion
-- @param actor <LuaPlayer?> the player performing the kick action, if any
```

features/report.lua
```diff
@@ -426,14 +426,15 @@ function Module.unjail(target_player, player)
...
     end
 end
 
 --- Bans the player and reports the ban to moderation log channel.
--- @param  player<LuaPlayer>
--- @param  reason<string?> defaults to empty string.
-function Module.ban_player(player, reason)
+-- @param player <LuaPlayer>
+-- @param reason <string?> defaults to empty string.
+-- @param actor <LuaPlayer?> the player carrying the ban action, if any
```
grilledham:
```suggestion
-- @param actor <LuaPlayer?> the player performing the ban action, if any
```

grilledham (commented):
I think these are the last few comments.

grilledham (approved):

PR #1434, opened 2024-09-04
Updates/frontier v8

RedRafe:
# Changes
- [x] fixed players' inventory not being wiped out completely on restarts
- [x] added Restart module to frontier: scenario manager in admin panel for `"global"` scenario settings, scenario restart/abort, and to discord notifications

 > in-game admin gui for scenario settings (tweak map difficulty and such...)
Screenshot from 2024-09-04 12-51-14

> example of the discord messages using the testing channel
Screenshot from 2024-09-04 17-54-18

### DIFF
--- a/map_gen/maps/frontier/shared/core.lua
+++ b/map_gen/maps/frontier/shared/core.lua
@@ -20,14 +20,27 @@ Public.scores = {
   shop_funds      = { name = 'shop-funds-frontier',       tooltip = {'frontier.shop_funds'},        sprite = '[img=item.coin]' },
 }
 
+Public.server_commands = {
+  none = 1,
+  reset = 2,
+  restart = 3,
+  switch = 4,
+}
+
 Public.ESCAPE_PLAYER = false
 Public.VALUE_7_PACKS = 451
 Public.PROD_PENALTY = 1.2 * 1.4^5
 
 local this = {
-  abort = false,
   rounds = 0,
-
+  server_commands = {
+    restarting = false,
+    mode = Public.server_commands.reset,
+    switch_map = {
+      name = nil,
+      mod_pack = nil,
+    }
+  },
   -- Map gen
   chart_queue = Queue.new(),
   height = 36,              -- in chunks, height of the ribbon world
@@ -58,6 +71,7 @@ local this = {
   scenario_finished = false,
 
   -- Enemy data
+  spawn_enemy_outpost = false,
   spawn_enemy_wave = false,
   invincible = {},
   target_entities = {},
--- a/map_gen/maps/frontier/shared/core.lua
+++ b/map_gen/maps/frontier/shared/core.lua
@@ -20,7 +20,7 @@ Public.scores = {
   shop_funds      = { name = 'shop-funds-frontier',       tooltip = {'frontier.shop_funds'},        sprite = '[img=item.coin]' },
 }
 
-Public.server_commands = {
+Public.restart_mode = {
   none = 1,
   reset = 2,
   restart = 3,
@@ -35,7 +35,7 @@ local this = {
   rounds = 0,
   server_commands = {
     restarting = false,
-    mode = Public.server_commands.reset,
+    mode = Public.restart_mode.reset,
     switch_map = {
       name = nil,
       mod_pack = nil,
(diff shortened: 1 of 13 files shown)


grilledham (approved):
I did a quick play test and looks good to me. Only minor thing I noticed was you can abort the restart when no restart is in progress, but I'll leave that up to you if you want to fix.

Also I added the frontier discord channel to the list of known channels. You can see how I did that in the channel: https://discord.com/channels/432567222481846283/1274493726647124071/1280927420999930050

RedRafe:
@grilledham thanks! It was a bug, but was kinda a nice feature (the possibility to prevent next restart to happen in async mode). So I'm refactoring the code to

1. add Server.restart to reboot server from scratch via command (like CS / DO(
2. add possibility for admins to asyncronously prevent/set how the next restart will happen, even if they wont be online.

Just to be clear, I haven't tested but I believe in CS and DO, when a `/restart` is initiated, if map is hosted in single player then the command does simply nothing? Or maybe in SP that command is not even available. I'm not planning any fallback to that command but maybe I can simply make the map soft-reset instead of rebooting a "fake" server. Neither of those options really make sense tbh...

grilledham:
@RedRafe

> Just to be clear, I haven't tested but I believe in CS and DO, when a `/restart` is initiated, if map is hosted in single player then the command does simply nothing?

It prints all the restart messages, but yes it doesn't actually restart the game. The way restarting works is it prints a special message that the server looks for and restarts.

> I'm not planning any fallback to that command but maybe I can simply make the map soft-reset instead of rebooting a "fake" server. Neither of those options really make sense tbh...

I would think if you are planning to do a hard reset it doesn't need to fall back to a soft reset. It would only be an issue in single player and that case doesn't matter for restarting.

RedRafe:
Okay I've reworked the additional scenario's admin tab, especially the restart part, to handle more cases: under **Restart settings** admins can:

1. set the restart mode to use when game will finish (None - will not restart, Reset - will soft reset the map for another round, Restart - will cause the server to restart and grab latest version from gh, Load - will cause the server to restart and use a savefile or a modpack IF specified below). This works mainly as default settings for the scenario, or as async method for admins to manage the map.
2. Load settings will be ignored if not using that mode

Additionally:

A. admins can instantly Abort/Restart the restart procedure using the Abort/Restart buttons (restart button is hidden under the dropdown).
B. if the game is hosted on our servers, it will rabase scenario code every 5 resets with a server restart. Otherwise it will simply fallback to a map reset, if not set to `None` (meaning restart got disabled)

Screenshot from 2024-09-13 17-17-25

This is the workflow in a nutshell:
```lua
function Restart.execute_server_command()
  local cmd = Public.get('server_commands')
  if not cmd.restarting then
    return
  end
  local is_hosted = Server.get_current_time() ~= nil

  if is_hosted and cmd.mode == Public.server_commands.switch then
    Server.start_game({
      type = (cmd.switch_map.name and 'save') or 'scenario',
      name = cmd.switch_map.name or 'frontier',
      mod_pack = cmd.switch_map.mod_pack,
    })
  elseif is_hosted and cmd.mode == Public.server_commands.restart then
    Server.start_game({
      type = 'scenario',
      name = 'frontier',
      mod_pack = nil,
    })
  elseif cmd.mode ~= Public.server_commands.none then
    Restart.queue_restart_event()
  end
  cmd.restarting = false
end
```

map_gen/maps/frontier/shared/core.lua
```diff
@@ -20,14 +20,27 @@ Public.scores = {
   shop_funds      = { name = 'shop-funds-frontier',       tooltip = {'frontier.shop_funds'},        sprite = '[img=item.coin]' },
 }
 
+Public.server_commands = {
```
grilledham: I'd probably call this something like `restart_type` at the very least it would avoid confusion with `this.server_commands`

grilledham: Maybe `restart_mode`.

grilledham (approved):
LGTM. I have one suggestion about renaming `Public.server_commands` that I'll leave up to you if you want to change.

Let me know when you want me to merge.

map_gen/maps/frontier/shared/core.lua
RedRafe: Seems totally reasonable, thanks!

PR #1436, opened 2024-10-18
Factorio 2.0 update

RedRafe:
Updated code to Factorio 2.0 API

### Relevant changes
  - [2.0] Updated `.luackeckr` with new classes, defines, and methods. `game` may be a bit out-of-date due to the very long list of methods it still holds
  - [2.0] Type of LuaObjects is now "userdata" instead of "table". Removed __self from the LuaObjects. Intended way of checking if an object is a lua object is to check type is userdata.
  - [2.0]  Added new controller type (remote), which is to build space platforms, so it allows ghost building but not any physical manipulation. Added `LuaPlayer::physical_surface, physical_surface_index, physical_vehicle and physical_position` read. Accessing `LuaPlayer::position` will return player's remote position, if available. Need to keep an eye on `.position` checks. Suggest to use `LuaPlayer::character::position` for future works to disambiguate any doubt about wanting to access player's "virtual" position or "physical" position. 
  - [2.0] Changed logistic networks to use `LuaLogisticPoint`. Added util `logistic_point_util.lua`
  - [2.0] Changed `on_cutscene_waypoint_reached` event's parameter "waypoint_index" to not be zero indexed. Need to test the ported Cutscene module.
  - [2.0] Added LuaRenderObject. All LuaRendering methods for manipulating object selected by id were moved to LuaRenderObject.
  - [2.0] `LuaBootstrap::raise_event()`'s and `LuaBootstrap::on_event()`'s "event" now also accepts event names as string as alternative to their numerical IDs.
  - [2.0] Removed the "flying-text" entity type. Use `LuaPlayer::create_local_flying_text` or `LuaRendering::draw_text` instead. `utila.game.lua` has been updated to replace all instances of it.
  - [space-age] Removed satellite item & recipe.
  - [space-age] Quality may break many inventory/item manipulations with additional param `.quality` added to old LuaItemStack. It's generally safe to use ItemStacks in vanilla, as `quality = normal` will be set as default.


RedRafe:
I managed to fix the test suite which was broken in the new API (@grilledham you'll have to take a look at `utils.test.discovery.lua` because I just plugged in whatever worked but I understand very little of it).

Now we can run tests and 50% of tests are broken. However, all the utils/gui tests pass while the features dont. I think it may be cause by this function used to setup the tests:

```lua
function Public.modify_lua_object(context, object, key, value)
    local old_value = object[key]
    rawset(object, key, value)

    context:add_teardown(function()
        rawset(object, key, old_value)
    end)
end
```
`rawset` may not work with new `LuaObjects` of type `userdata`.

Screenshot from 2024-10-18 16-35-55

features/gui/evolution_progress.lua
```diff
@@ -34,7 +35,7 @@ local function get_evolution_percentage()
         return 0
     end
 
-    local value = round(game.forces.enemy.evolution_factor * 1000) * 0.001
+    local value = round(game.forces.enemy.get_evolution_factor(RS.get_surface_name()) * 1000) * 0.001
```
grilledham, line 38: So now there is separate evolution for each surface? This is fine for now. But I wonder if we need to show the evolution for the current surface or for each surface? Something for the future.

features/player_stats.lua
```diff
@@ -46,7 +48,10 @@ ScoreTracker.register(player_deaths_name, {'player_stats.player_deaths'})
 ScoreTracker.register(player_console_chats_name, {'player_stats.player_console_chats'})
 ScoreTracker.register(player_items_crafted_name, {'player_stats.player_items_crafted'})
 ScoreTracker.register(player_distance_walked_name, {'player_stats.player_distance_walked'})
-ScoreTracker.register(satellites_launched_name, {'player_stats.satellites_launched'}, '[img=item.satellite]')
+if not SA.enabled() then
+    ScoreTracker.register(satellites_launched_name, {'player_stats.satellites_launched'}, '[img=item.satellite]')
+end
+ScoreTracker.register(rockets_launched_name, {'player_stats.rockets_launched'}, '[img=item.rocket-silo]')
```
grilledham, line 54: Maybe this should be an else. If SA isn't enabled tracking rocket launches without satellites probably isn't that interesting?

Also this is a good compromise to show rocket launches instead of satellites for SA. I was going to suggest just removing it, but I think your solution is better.

map_gen/maps/crash_site/scenario.lua
```diff
@@ -896,7 +896,7 @@ local function init(config)
             {name = 'utility-science-pack', price = 125},
             {
                 price = 100,
-                name = 'player-port',
+                name = 'simple-entity-with-owner',
                 name_label = 'Train Immunity (1x use)',
                 description = 'Each player port in your inventory will save you from being killed by a train once.'
```
grilledham:
```suggestion
                description = 'Each train immunity item in your inventory will save you from being killed by a train once.'
```

features/gui/evolution_progress.lua
RedRafe: Yes, evo is per-surface-per-force now. in 2.0 this button can be changed to show as main number the `LuaPlayer::surface` (or `physical_surface`), and have in tooltip an easy table.concat of all surfaces and enemies evos listed. Although this wouldnt work well in PvP scenarios, but I guess those would disable the button and do one custom for their case. There are too many possibilities and ifs to take in account otherwise

map_gen/maps/crash_site/scenario.lua
RedRafe: Will update

features/player_stats.lua
RedRafe: i'm not sure how we're going end up with SA compatibility, but I guess SA module will get populated with time. I think it can collect all the edge cases of SA for now. 
In this particular case, I think Factorio itself it's handling badly the issue because SA is straight up casting to nil some things it doesnt use (I would have preferred they simply removed the recipe and kept the item for editor mode, but whatever). I just dont want to have hundreds of ifs around the code for every little change. I could if else this, but then I'll have to check for the existence of other stat where that is used as well... Like Frontier scenario cares about launches rather than satellites, so I thought it was not a big deal to have 1 additional stat even if it's used less frequently

map_gen/shared/entity_placement_restriction.lua
```diff
@@ -214,7 +214,7 @@ local on_built_token =
 
         local index = event.player_index
 
-        local stack = event.stack
+        local stack = event.consumed_items.get_contents()[1]
```
grilledham: Can `consumed_items.get_contents()` return multiple stacks? If so wouldn't we need to loop over all stacks. This could also be a todo to solve later.

features/player_stats.lua
grilledham: Ah, if there are cases where we care about rocket launches, then we can keep this. It's not the worst to have an extra stat.

utils/global.lua
```diff
@@ -1,6 +1,9 @@
 local Event = require 'utils.event_core'
 local Token = require 'utils.token'
 
+--local matching_path = '^.+/currently%-playing/(.+)$'
```
grilledham: Presumably we can delete this commented out code. Or are you not sure what it should be?

utils/global.lua
```diff
@@ -47,8 +50,10 @@ if _DEBUG then
     Global.names = names
 
     function Global.register(tbl, callback)
-        local filepath = debug.getinfo(2, 'S').source:match('^.+/currently%-playing/(.+)$'):sub(1, -5)
+        log(debug.getinfo(2, 'S').source)
+        local filepath = debug.getinfo(2, 'S').source:match(matching_path):sub(1, -5)
         local token = Token.register_global(tbl)
+        log(filepath)
```
grilledham: I would probably also get rid of the logs. Once this is working we don't need them. But not the end of the world to keep them.

utils/global.lua
grilledham: Same again for utils/gui.lua

grilledham:
> I managed to fix the test suite which was broken in the new API (@grilledham you'll have to take a look at `utils.test.discovery.lua` because I just plugged in whatever worked but I understand very little of it).
> 
> Now we can run tests and 50% of tests are broken. However, all the utils/gui tests pass while the features dont. I think it may be cause by this function used to setup the tests:
> 
> ```lua
> function Public.modify_lua_object(context, object, key, value)
>     local old_value = object[key]
>     rawset(object, key, value)
> 
>     context:add_teardown(function()
>         rawset(object, key, old_value)
>     end)
> end
> ```
> 
> `rawset` may not work with new `LuaObjects` of type `userdata`.
> 
> Screenshot from 2024-10-18 16-35-55

I'll look at it later. I think for now it's fine for the tests to not be passing.

map_gen/shared/entity_placement_restriction.lua
RedRafe: `consumed_items` is a `LuaInventory` (the devs were not clear what or IF it is possible for player to build an entity using more than 1 item(s). So I'm defaulting to picking 1st element as I assume 99.99% of the time that'll be it, but definitely marked for TODO and see if it gets changed down the line / the devs will be more clear about it. The respective bot even is still using `LuaItemStack` instead

utils/global.lua
RedRafe: Yeah sorry, logs were for me debugging this part cuz it gave me plenty of headaches. Will of course remove

grilledham (approved):
This is amazing work, thanks for taking the time to do this, very much appreciated.

I can't really test this until after SA launches, but the changes look fine to me.

You have my approval to merge this after we update the server to 2.0/SA.

map_gen/shared/entity_placement_restriction.lua
grilledham: Cool. I think a todo/fixme in the code is all that is needed for now.

map_gen/shared/entity_placement_restriction.lua
```diff
@@ -225,7 +225,7 @@ local on_built_token =
             }
         )
 
-        local player = game.get_player(index)
+        local player = game.get_player(index or 'none')
```
grilledham: What is the `or 'none'` for? What happens if there is a player called 'none'.

grilledham:
Regarding your comments in the other PR, this one:

>There are some of our features that are built-in in 2.0, i.e.
lanfill remover: landfill tiles act like other tiles and can be deconstructed after placed with default decon planner (+ there's super-force-build and undo mechanics that already mess with player actions so I'd just keep new base game and deprecate ours)
spidertron control: now possible to drag and select spidetrons to remote control without the remote item (deprecate module as well)
disable loaders?

Do you want to delete the landfill remover, spidertron control and loader code as part of this PR?

map_gen/shared/entity_placement_restriction.lua
RedRafe: Yeah, if there's a player called `none` this would fail. Maybe I can change the null token to be empty string `''`. I tested and it would work.
Issue is, `event.player_index` may be nil so `game.get_player()` would crash when `index = nil`

RedRafe:
> Regarding your comments in the other PR, this one:
> 
> > There are some of our features that are built-in in 2.0, i.e.
> > lanfill remover: landfill tiles act like other tiles and can be deconstructed after placed with default decon planner (+ there's super-force-build and undo mechanics that already mess with player actions so I'd just keep new base game and deprecate ours)
> > spidertron control: now possible to drag and select spidetrons to remote control without the remote item (deprecate module as well)
> > disable loaders?
> 
> Do you want to delete the landfill remover, spidertron control and loader code as part of this PR?

Honestly I didnt test this part at all. I'm fine keeping it for now (modules may be disabled by default) and come back to check when testing the scenarios to see why it was needed and if new behavior can substitute it or just the entire feature as QoL

map_gen/shared/entity_placement_restriction.lua
RedRafe: Rewind. I checkd again and this token only gets used by `on_built_entity` (not `on_robot_built_entity`) so it would be fine not handling the `event.player_index = nil` case as it should never react to that event - I think this was an oversight on my part becase when tweaking the `get_contents()` above I may have moved parts of the code and this happened to pop out and stay, when it should have never been here sorry

grilledham:
> > Regarding your comments in the other PR, this one:
> > > There are some of our features that are built-in in 2.0, i.e.
> > > lanfill remover: landfill tiles act like other tiles and can be deconstructed after placed with default decon planner (+ there's super-force-build and undo mechanics that already mess with player actions so I'd just keep new base game and deprecate ours)
> > > spidertron control: now possible to drag and select spidetrons to remote control without the remote item (deprecate module as well)
> > > disable loaders?
> > 
> > 
> > Do you want to delete the landfill remover, spidertron control and loader code as part of this PR?
> 
> Honestly I didnt test this part at all. I'm fine keeping it for now (modules may be disabled by default) and come back to check when testing the scenarios to see why it was needed and if new behavior can substitute it or just the entire feature as QoL

Ok, so do we want to just disable these modules by default in the minimal scenario I made? I can do that if you would prefer?

RedRafe:
> > > Regarding your comments in the other PR, this one:
> > > > There are some of our features that are built-in in 2.0, i.e.
> > > > lanfill remover: landfill tiles act like other tiles and can be deconstructed after placed with default decon planner (+ there's super-force-build and undo mechanics that already mess with player actions so I'd just keep new base game and deprecate ours)
> > > > spidertron control: now possible to drag and select spidetrons to remote control without the remote item (deprecate module as well)
> > > > disable loaders?
> > > 
> > > 
> > > Do you want to delete the landfill remover, spidertron control and loader code as part of this PR?
> > 
> > 
> > Honestly I didnt test this part at all. I'm fine keeping it for now (modules may be disabled by default) and come back to check when testing the scenarios to see why it was needed and if new behavior can substitute it or just the entire feature as QoL
> 
> Ok, so do we want to just disable these modules by default in the minimal scenario I made? I can do that if you would prefer?

I wouldn't remove them without testing. Having them still in 2.0 will not cause any issue, just redundancy if built-in feature will cover 100% of the cases we use them for. Let's just say it'll be a good "first issue" to port or remove code in 2.0 if checking with the scenarios reveals those modules are not needed anymore

Edit. There'll be probably many other modules and tasks that havent been updated in years even in recent version of 1.0-1.1 that I didn't catch with the eye at first glance but could now be optimized or removed in 2.0 other than these 2 examples I believe

map_gen/shared/entity_placement_restriction.lua
grilledham: > Issue is, event.player_index may be nil so game.get_player() would crash when index = nil

So why not `local player = index and game.get_player(index)`

RedRafe: That would bee too clever and right to do (fixed, thanks!)

RedRafe:
Small batch of one-line hotfixes in  our current SA maps. May not be the best solution, but im just too tired and I'll have a better look into it the following days

grilledham (approved):

PR #1437, opened 2024-10-19
Add minimal scenario

grilledham:
I wanted a scenario we can use that has the non-essential features turned off. I think this would be an better scenario to use for SA compared to develop. There is also the possibly that after we merge this for 1.1, we change it again in your PR for SA.

If there are any other features you think we should turn off I'm open to suggestions. landfill_remover, autofill and turret_active_delay could be controversial. I'm also not sure what to use for `player_create.starting_items` but I figured in a multiplayer map it doesn't make much difference.

### DIFF
--- a/scenario_templates/minimal/map_selction.lua
+++ b/scenario_templates/minimal/map_selction.lua
@@ -1,0 +1,7 @@
+local config = require 'config'
+config.redmew_surface.enabled = false
+config.market.enabled = false
+config.reactor_meltdown.enabled = false
+config.player_create.starting_items = {}
+config.player_rewards.enabled = false
+config.apocalypse.enabled = false


grilledham:
@RedRafe could you review please. I don't think I can assign you are a reviewer unless you are a member.

RedRafe:
There are some of our features that are built-in in 2.0, i.e.

- lanfill remover: landfill tiles act like other tiles and can be deconstructed after placed with default decon planner (+ there's super-force-build and undo mechanics that already mess with player actions so I'd just keep new base game and deprecate ours)
- spidertron control: now possible to drag and select spidetrons to remote control without the remote item (deprecate module as well)
- disable loaders?

grilledham:
> There are some of our features that are built-in in 2.0, i.e.
> 
> * lanfill remover: landfill tiles act like other tiles and can be deconstructed after placed with default decon planner (+ there's super-force-build and undo mechanics that already mess with player actions so I'd just keep new base game and deprecate ours)
> * spidertron control: now possible to drag and select spidetrons to remote control without the remote item (deprecate module as well)
> * disable loaders?

Let's keep those for this PR, as this will go into 1.0. But remove all that stuff from the Redmew scenario for 2.0/SA.

RedRafe:
These I'm not sure what to do:

- turret_active_delay:
- research_printer -- even 1.1 has this autoprinter now but it's not fancy

grilledham:
> These I'm not sure what to do:
> 
> * turret_active_delay:
> * research_printer -- even 1.1 has this autoprinter now but it's not fancy

research_printer is at least useful for printing to discord.
I'm leaning towards keeping turret_active_delay especially if we keep autofill. Even if it's not updated for the turret types. Worse case scenario we just turn it off later.

RedRafe:
Works for me. Edit: new turret types are expansion-only

RedRafe:
@grilledham  Edit. built-in research printer has been improved with rich text so we can deprecate our module because it's just a duplicate

Screenshot from 2024-10-21 15-09-59

grilledham:
> @grilledham Edit. built-in research printer has been improved with rich text so we can deprecate our module because it's just a duplicate
> 
> Screenshot from 2024-10-21 15-09-59

@RedRafe We can remove the code that prints to game, but we'll keep the part that prints to discord.

PR #1439, opened 2024-10-26
Do quality

RedRafe:
- Added a 2.0 Danger Ore x Quality preset
- Small fixes here & there from 1st week of hosting


map_gen/maps/danger_ores/modules/ban_quality_mining.lua
```diff
@@ -0,0 +1,30 @@
+-- This module prevents all but the normal quality drills from being built on top of resources
```
grilledham, line 1: What do quality mining drills do? Are they just faster?

grilledham (approved):
LGTM

map_gen/maps/danger_ores/modules/ban_quality_mining.lua
RedRafe: No, they're same speed but have a reduced resource drain 
normal = 100%, uncommon = 88%, .... lenegdary 20% (sort of mining productivity). SO I disabled it for DO. Just no quality miners. Everything else is ok

PR #1440, opened 2024-10-26
Updates normal cs to 2.0

RedRafe:
Updated crash site to 2.0
- Updated only core scenario code and tested `normal` preset
- Migrated internal names for collision masks, rocks, modules, inserters, chests of all repo


grilledham:
Did a quick play test and got this error:
image

Should we be using `registration_number` instead of `unit_number`?

RedRafe:
Yes this is one of the migrated behaviors. I shall have a second look at this and possibly other migrated cases of registrations

RedRafe:
> Should we be using `registration_number` instead of `unit_number`?

`register_on_object_destroyed(LuaEntity)` returns --> `registration_number, useful_id, defines.target_type`.

Same 3 values are provided as params in the raised event when object is destroyed `on_object_destroyed`. 

`useful_id` is same as `LuaEntity::unit_number` so in my opinion, keeping the entries in the dictionary sorted by key `LuaEntity::unit_number` is more convenient in 2.0 as we don't need to keep tabs on `registration_number` which was an arbitrary ID generated for the event

RedRafe:
I've added extra checks for `useful_id` in case any mod registers rocks/trees (simple entities do not have uid) just to play it safe

ref:
> The registration is global across all mods, meaning once one mod registers an object, all mods listening to [on_object_destroyed](https://lua-api.factorio.com/latest/events.html#on_object_destroyed) will receive the event when it is destroyed.

grilledham:
Did another play test, ran into two issues:

1. Outposts can't be captured. I think it's because in `outpost_builder.lua`  `turret_died` function line 1673 we need to change `local number = event.unit_number` to `local number = event.useful_id`.

2.
image

RedRafe:
These have been fixed. Debug mode on this map is still broken, probably on cutscene. It may have to do with renders but I'll need to check

grilledham:
> These have been fixed. Debug mode on this map is still broken, probably on cutscene. It may have to do with renders but I'll need to check

Do you want to fix that before we merge? We could merge without fixing if this blocks other PRs.

RedRafe:
I'm in no rush to merge. I'll take the time to review and fix what's missing. It just has been a tough weeks and I've spent more time just playing tbh 😆

grilledham:
> I'm in no rush to merge. I'll take the time to review and fix what's missing. It just has been a tough weeks and I've spent more time just playing tbh 😆

That's cool. No worries in taking the time you need, there is no rush. I just wanted to make sure I wasn't blocking any of your work.

RedRafe:
@grilledham fixed the polygon function of renders which was changes and broke some of CS renders. The cutscene is a delicate piece of the scenario during the migration so I'd like a second feedback if everything is working as supposed

map_gen/maps/crash_site/events.lua
```diff
@@ -245,7 +245,7 @@ local function do_pole(entity)
         return
     end
 
-    local supply_area_distance = entity.prototype.supply_area_distance
+    local supply_area_distance = entity.get_supply_area_distance(entity.quality)
```
grilledham: I think this needs to be.
```suggestion
    local supply_area_distance = entity.prototype.get_supply_area_distance(entity.quality)
```

grilledham:
The cutscene doesn't error in debug mode. But I don't think it's displaying all the text that it should. It shows the first couple of text messages, but after that I only get the clicking sound that is played when rendering text but no actual text.

I'm not sure what is wrong. But we could leave fixing the cutscene to later if you don't want to fix it now.

grilledham:
I tested this again and I think it is good expect for the cutscenes. Do we want to merge it now and fix the cutscenes later?

RedRafe:
Yes I guess we could merge this and maybe have it run on a server. Usually letting ppl play test it makes it easier to spot missing bugs since the have more time than me to fully test the scenario from beginning to end

PR #1442, opened 2024-11-10
Update Frontier to 2.0

RedRafe:
Updated Frontier scenario to 2.0. This scenario now works on `nauvis` surface instead of redmew's


map_gen/maps/frontier/modules/enemy.lua
```diff
@@ -439,7 +440,11 @@ function Enemy.stop_tracking(entity)
 end
 
 function Enemy.get_target()
-  return Table.get_random_dictionary_entry(Public.get().target_entities, false)
+  local _list = Public.get().target_entities
+  if table_size(_list) == 0 then
```
grilledham: Just be aware that `table_size` is O(n) so might not be the best when you expect `_list` to contains lots of entries. In the past I've worked around this by having a count field in the table.

grilledham: On second look, why did you make this change? Specifically the `if table_size(_list) == 0` part, doesn't `get_random_dictionary_entry` return `nil` on empty tables?

grilledham (approved):
LGTM, you have my approval to merge.

map_gen/maps/frontier/modules/enemy.lua
grilledham: Oh is it because `math.random(1, 0)` fails? In which case this is reasonable.

Or we could change `get_random_dictionary_entry` to return `nil` for empty tables. Up to you.

RedRafe: Yeah I've just looked at it now and it was the `math.random(1,0)` inside `table::get_random_dictionary_entry` that caused the issue. On a second thought, I should probably fix it in the util as that function is used in other parts of the repo

On a side note, this function is only called <30x for a whole run so I was not too concerned with performance (edit. i.e. when silo explodes. also, they voted out this feature as it was deemed "too powerful" so it's actually not called rn)

grilledham (approved):

PR #1445, opened 2024-11-12
Add inception scenario

RedRafe:
Introduces a new scenario called "Inception".

Currently only features a nice maze with mixed patches and fancy terrain gen. Supports Nauvis. Will add more features and (likely) support for SA's planets in the future


grilledham (approved):
Havn't tested it, but the code looks good.

PR #1447, opened 2024-11-20
Danger ores refactor

RedRafe:
# Changes

- [x] created new template `danger_ore/scenario.lua` to collect and boostrap DO scenarios
- [x] created new config `danger_ore/configuration.lua` to collect and enable/disable/tweak common features shared between DO scenarios
- [x] updated all scenarios to use the shared scenarion boostrapper and configuration, only editing what the scenario is changing from the default template instead of copy/pasting all modules
- [x] removed duplicates `..._deadlock_beltboxes.lua`, now all mod compatibilities are listed in `danger_ores/compatibility/` folder, and listed/enabled in `configuration.lua` if supposed to work regardless of the modpack played (deadlock, aai-loaders, early construction...)
- [x] updated map poll and scenario templates

Modded scenarios have been ported but as usual, they'll need to be updated for real when/if they'll be updated and released to 2.0 and then enabled from map poll.

Overall I maintained pretty much the same scenarios as before, but I took the liberty to tweak some settings to make them global (some scenarios had different terraforming setting, or different tech multipliers). I tried to merge all these settings into the configuration, so some scenarios may have higher multiplier than before, or different terraforming settings.

I believe most settings them were slightly different because they were copy-pasted around and then got changed in the new scenarios but not in the old one. My goal would be to have all scenario share as much as possible form the configuration file, and maybe tweak the global settings to find the sweet spot, rather than having each scenario fine-tune to the last digit its config. So it would be easier to maintain and turn on/off/tweak settings for all.

I maintained custom settings for those that I believed needed them (modded scenarios, or ribbon maps with terraforming for example). But happy to revert or further tweak the settings where needed :)


map_gen/maps/danger_ores/presets/danger_ore_3way.lua
```diff
@@ -0,0 +1,66 @@
...
+local Scenario = require 'map_gen.maps.danger_ores.scenario'
+local ScenarioInfo = require 'features.gui.info'
+
+ScenarioInfo.set_map_name('Danger Ores - 3way')
+ScenarioInfo.add_map_extra_info([[
+  This map is split in three sectors [item=iron-ore] [item=copper-ore] [item=coal].
+  Each sector has a main resource and the other resources at a lower ratio.
+]])
+
+ScenarioInfo.set_new_info([[
```
grilledham, line 13: We could potentailly get rid of the `ScenarioInfo.set_new_info` from all the presets it's not like we are updating it. Unless there is anyhting you want to keep.

grilledham (commented):
This is fantastic work! Really something that should have been done a long time ago, but I'm too lazy 😅

From what I've seen this all looks good, but I'll need more time to check it all.

map_gen/maps/danger_ores/presets/danger_ore_3way.lua
RedRafe: Yeah we could delete them. Mainly because 
1. I don't think anyone reads them
2. After updating the thirs preset I'm quite sure it's the same text copied over and over between maps... but I didn't read it. Commentary on changes is nice, but then... Git does the same.

grilledham: It was meant for the players. But it's too much effort to update and as you say players probably don't read them anyway.

RedRafe: We could keep just 1 changelog that gets injected in the template/base scenario.
Would be shared among all maps, but could be a fair tradeoff between the hassle of updating all maps and keeping the players informed.
I may have an idea for this.

Will do in a separate pr tho

grilledham: OK let's do this in another PR then, no rush though.

grilledham (approved):

PR #1450, opened 2024-11-23
Set hidden tile for landfill tiles

RedRafe:
In Factorio 2.0, landfill tile are now minable/refundable.
For scenarios that use `landfill` tiles in terrain generation now a `shallow-water` tile will be placed beneath landfill as `LuaTile::hidden_tile`, so when mining landfill, water is discovered.
Scenarios that do not want player to find water beneath need pass as hidden tile another valid tile in the shape function.

Screenshot from 2024-11-23 04-46-47

### DIFF
--- a/map_gen/shared/generate.lua
+++ b/map_gen/shared/generate.lua
@@ -30,6 +30,12 @@ local function do_tile_inner(tiles, tile, pos)
     end
 end
 
+local function do_landfill_hidden_tile(tiles, tile, pos)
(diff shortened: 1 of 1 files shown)


grilledham (approved):

grilledham:
I think this is causing errors in crashsite:
`
2424.871 Script @__level__/utils/task.lua:77: __level__/map_gen/shared/generate.lua:34: attempt to index local 'tile' (a boolean value)
`

PR #1451, opened 2024-11-23
Organize danger ores changelog

RedRafe:
Moves per-scenarios changelog into a shared entry list under `danger_ores/changelog.lua`.
Presets can list custom changes using their own tag `[DO:<preset_name>]` or simply `[DO]` if it involves more/all.
Entries are listed in descending chronological order because the info panel automatically scrolls to the end when opening it, so players will immediately see the latest changes.

Screenshot from 2024-11-23 05-18-39


grilledham (approved):
This is a great idea, thanks for doing this.

PR #1452, opened 2024-11-23
Add startup techs to DO

RedRafe:
Unlocks some techs (uranium, oil) at startup if they require actions to be unlocked, to allow players to still unlock the recipes and prevent any softlock. This also helps with the dats stage that may add early coal liq or stone->uranium

Screenshot from 2024-11-23 05-51-26

### DIFF
--- a/map_gen/maps/danger_ores/modules/technologies.lua
+++ b/map_gen/maps/danger_ores/modules/technologies.lua
@@ -1,0 +1,30 @@
+local Event = require 'utils.event'
+
+local trigger_names = {
+  uranium = {
+    'uranium-mining',
+    'uranium-processing',
+  },
+  liquefaction = {
+    'oil-processing',
+  },
+}
+
+return function(config)
+  Event.on_init(function()
+    for _, force in pairs(game.forces) do
+      for key, enabled in pairs(config) do
+
+        local names = enabled and trigger_names[key] or {}
+
+        for _, name in pairs(names) do
+          local tech = force.technologies[name]
+          if tech and tech.prototype.research_trigger then
+            tech.researched = true
+          end
+        end
+
+      end
+    end
+  end)
+end
--- a/map_gen/maps/danger_ores/modules/technologies.lua
+++ b/map_gen/maps/danger_ores/modules/technologies.lua
@@ -1,29 +1,19 @@
 local Event = require 'utils.event'
 
-local trigger_names = {
-  uranium = {
-    'uranium-mining',
-    'uranium-processing',
-  },
-  liquefaction = {
-    'oil-processing',
-  },
-}
-
 return function(config)
-  Event.on_init(function()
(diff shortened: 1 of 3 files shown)


grilledham:
This means you can build oil refineries before you have researched oil gathering. Would it be better to unlock the tech when oil gathering is researched?

RedRafe:
Changed to unlock them when the previous one is researched

RedRafe:
I'm not 100% sure if the check for triggers is a good thing.
Technically, making it able to unlock every tech regardless of its research cost would be more useful, on the other hand, it is not universal and it's a bit OP to give out techs for free when they can just be researched.

This is true for coal liquefactionm which is listed as "possible unlock" bu its not unlocked by script as it can be researched with packs. However, current map got me thinking as uranium and oil did not spawn before chem science, so players were stuck with no oil regardless of the tech unlocks. I think the issue to fix is at root: consistently spawn uranium and oil withing a reasonable distance (especially with terraforming enabled) so players dont soflock

grilledham:
> I'm not 100% sure if the check for triggers is a good thing.

Yeah I'm not sure why it would be needed. We are specifing the techs to unlock for a given tech, that is all the control we need.

> However, current map got me thinking as uranium and oil did not spawn before chem science, so players were stuck with no oil regardless of the tech unlocks. I think the issue to fix is at root: consistently spawn uranium and oil withing a reasonable distance (especially with terraforming enabled) so players dont soflock

I think this is working as intended, with the redmew_data mod you can turn coal in to oil products and stone in to uranium, so no need to spawn them for normal maps.

RedRafe:
With the refactor, I've coded so

if redmew_data is present, do not spawn oil/uranium as they're processed from coal/stone
otherwise, make crude and oil patches

This is injected as part of the redmew compatibility, so the scenarios would always work whether users play it vanilla or with the data mod.

With mod it's always safe as no special patches are needed and coal liq effects are duplicated in oil-processing.
Without mod, it's a matter of "luck" and fine tuning the radius where special resource spawn

map_gen/maps/danger_ores/modules/technologies.lua
```diff
@@ -0,0 +1,20 @@
...
+    return
+  end
+
+  Event.add(defines.events.on_research_finished, function(event)
+    local research = event.research
+    local techs = research.force.technologies
+
+    for _, name in pairs(unlocks[research.name] or {}) do
+      local tech = techs[name]
+      if tech and tech.prototype.research_trigger then
```
grilledham: Just for clarity I was expecting `and tech.prototype.research_trigger` to be removed. Apologies if you just havn't got round to it yet.

RedRafe: Yes, updated, thanks :)

grilledham (approved):

Issue #1398, opened 2024-02-12
Entity Placement Restrictions by type


RedRafe:
Looking at DangerOres & similar scenarios, entities are whitelisted/blacklisted by name lists from static files.

This results in a lot similar files & setups that are not very mod-friendly.

I suggest to switch to ban/allow by type instead (where possible).

This could be done by reviewing the scenarios & change the code to generate the list of entities by reverse-engeneering the names from the available prototypes in game at runtime, given a listed of allowed/banned entity types, so to be more compatible. 

Also, 2 more APIs to whitelist/blacklist directly by a type list could be added to the module's logic. Not necessary, but still useful for future projects.

PR #1456, opened 2024-11-26
Implement entity restriction by type for Danger Ores

RedRafe:
Implements https://github.com/Refactorio/RedMew/issues/1398

# Changes

- Implemented a module for DangerOres that uses entity restrictions by type instead of by name.
Does not change much for vanilla, but ensures broad compatibility with any mod (no need to whitelist higher tier belts or miners).
Module accepts, in order of precedence:
  - `banned_entities` (by name): will always be removed when built
  - `allowed_entities` (by name): will always be allowed, if not already blacklisted, even if type is not listed
  - `types` (by type): will be allowed

- FX: added funny messages & effects when destroying a banned entity:
  - `water-splash` + message if ghost
  - random explosion + message if solid entity

- Compatibility: removed all custom entity lists from modded scenarios as they're no longer require.


RedRafe:
Resources have been added to the list of placeable entities to mirror behavior of PR https://github.com/Refactorio/RedMew/pull/1458 and allow player in editor mode to place resources (I dont think there's any way in game foe a character to "build" resources, but useful for admins to hit `/editor` and spawn whatever may be needed to fix the scenario).

grilledham:
This is great work to simplify the configs for the maps.

But two things concern me:
1. The items placed on ores are lost. This feels too harsh. It's a painful way to learn the rules and given you are not supposed to build on the ore anyway it isn't a useful game mechanic to destroy items. It would also be really annoying to experienced players when new players lose expensive items trying to build on the ores.
2. We don't tell the players what they can build on the ores. I'm OK with the new messages as long as we also show a message with what can be built/explaining the rules.

RedRafe:
>     1. The items placed on ores are lost. This feels too harsh. It's a painful way to learn the rules and given you are not supposed to build on the ore anyway it isn't a useful game mechanic to destroy items. It would also be really annoying to experienced players when new players lose expensive items trying to build on the ores.

From playing it, I dont think there's anything in DO that is valuable enough to feel sorry if lost. Even the cost of a nuclear reactor which is the most expensive entity in game, it's merely peanuts with tech multiplier of 20+ and infinite ore to clear around. It is true that newbs could spam entities around and annoy vets with losing a lot of entities, maybe this should be accounted and controlled as a form of griefing?


>     2. We don't tell the players what they can build on the ores. I'm OK with the new messages as long as we also show a message with what can be built/explaining the rules.

This may be true. There is a general message displayed in the info map:
```txt
  You may not build the factory on ore patches. Exceptions:
      - pipes, underground pipes & pumps
      - belts & underground belts
      - rails, stations, trains and vehicles
      - power poles
      - mining drills and pumpjacks
```
But I doubt many ppl read it. There's too much text to read when joining a map, that I dont think any of it gets even seen. I could add a message that pops out when a player builds 1st item on ore (or just builds any item). Sort of cutscene. I know diggy has some introduction to players as well to onboard them to the features of the scenario. Maybe DO could have something similar

RedRafe:
A counter-proposal could be:

- for the destroy event: track C, the number of first 100 items "lost" by player 1-100. Each player has a chance of C% to have that entity lost when built on ore, otherwise refounded (overall, a 50% chance of losing first 100 items, but it starts very slow at 1-10%). After 100, the probability of it getting destroyed is 100%

- for the display of items allowed: first 100 entities lost will also print the full list of items allowed on ore. Then, list will only be available in the info map and not re-printed each time.

I'm also planning of adding a gui specific for Danger Ores with a few info (i.e. quantity of ore mined, n. of rockets to win, entities allowed, entities lost(?), top builder/fighter...) Something alike what we print at the end of the game, but available at any moment. The allowed list could be displayed in this info page. This is of course just a future project

grilledham:
I see placing banned items on ore in the same way as trying to place an item out of reach or on top of another entity. No action is taken and you are given a message explaining why. I don't see a situation where players would want to lose their items for what could just be a misclick. Personally I don’t like this behavior, but you can poll the players to get their opinions if you want.

RedRafe:
Implemented entity refund & content spillage (default option), as well as info message with allowed entities to be built (in the screenshot may be very small, but the cursor shows 4 stone furnaces in hand even if I tried to build them several times on ore, as they got refunded).

Turning the `.refund = false` will instead enabled the harsh version with just entity destroy and ominous message from scenario. I've kept it as an option.

Screenshot from 2024-11-27 10-46-44

grilledham (approved):
Thanks for implementing the refund and the allowed entities message.

The explosions are funny. They might surprise players the first time they see them. Hopefully they won't get annoying.

Once you fix the merge conflict you can merge.

RedRafe:
Yeah I most glad for ghosts it only uses `water-splash` instead of drawing randomly from the explosion list, so only way for players to spam explosion would be by doing it manually which shouldn't bother server's performance :)

As always, thanks for reviewing @grilledham !

PR #1460, opened 2024-11-29
Add DO:X Cross map

RedRafe:
Adds a X Cross map (very similar to the 3Way T shaped one).
Each arm has a main ore among iron, copper, coal, stone.
Map is rotated 45 degrees

x_cross
x_cross_zoom


RedRafe:
For reference, `main_ores_builder` is same of the 3-way (T cross) just without the tip of one arm cropped, and rotated 45 degs. Same for the scenario code, while ore config is default ore config, + stone not commented out

grilledham:
I don't think the map gen is working, I just get this:
image

I don't get any errors in Debug mode. Do you know what is going on?

grilledham:
My bad, I messed up the require in my map selection.

grilledham (approved):
Looks good. Stone might be a little high but we can adjust after a playthrough.

PR #1461, opened 2024-11-29
Add DO:Permanence map

RedRafe:
Permanence is a module that makes it so when mining an entity (if it's not one of the entities whitelisted to be built on ore like miners, poles, belts, rails...) it will spawn ore beneath it.

This evil map makes planning the base in advance much more important, as even upgrading machines will cause ore to spawn beneath old machine mined!

Map description:

```
  Mined entities will create ore drops beneath them, encouraging players to strategize their factory layouts.

  Only specific machines, such as assembling machines, furnaces, roboports, chemical plants... can generate ore drops,
  while entities already allowed on ore like miners, belts, and power poles will not.

  Additionally, upgrading your base will also spawn resources, so plan and upgrade carefully!
  Transform your gameplay and embrace the permanence of your factory's footprint!
```

`danger_ores/modules/permanence.lua` has additional info on parameters it accepts to customize it.
By default, iron > copper > coal > stone are spawned (only 1 type of ore per entity)m and total amount of ore spawned = entity's health.

Here's an example of the ore footprint left behind when mining assemblers or upgrading inserters
> permanence

### DIFF
--- a/map_gen/maps/danger_ores/modules/allowed_entities.lua
+++ b/map_gen/maps/danger_ores/modules/allowed_entities.lua
@@ -191,6 +191,16 @@ end
 
 local Public = {}
 
+---@param entity LuaEntity
+---@return bool
+Public.is_allowed = function(entity)
+  local e = get_entity_info(entity)
+  if not banned_entities[e.name] and (allowed_entities[e.name] or types[e.type]) then
+    return true
+  end
+  return false
+end
+
 ---@param config
 ---@field types? table<string, bool>
 ---@field allowed_entities? table<string, bool>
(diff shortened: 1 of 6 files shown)


grilledham (approved):
Very evil 😄.

I get why it's happend, but spliters spawning ore and underground belts not feels odd.

If you don't want to change it, feel free to merge.

RedRafe:
> I get why it's happend

Haha why? It actually came up in a chat between me & Jay.

Regarding the mechanics, I'm open to suggestions. :) I've not yet played it, other than testing the feature. And I just realized players can just shoot dead the machines instead of mining them, so I think I'll add an handler for `on_entity_died` as well

grilledham:
> Haha why? It actually came up in a chat between me & Jay.

When building I think spliters and underground belts get used in similar ways. But I'm happy to wait for playtesting to see what is right.

> And I just realized players can just shoot dead the machines instead of mining them, so I think I'll add an handler for on_entity_died as well

That would be good to add.

PR #1463, opened 2024-12-02
Update water settings in map gen settings

RedRafe:
# Changes

- [x] Updated `resources/map_gen_settings` presets to 2.0
- [x] Updated scenarios with new water format
- [x] Added `map_gen_size` in resources for future reference

Basically `MapGenSettings::water` and `MapGenSettings::terrain_segmentation` have been moved inside `MapGenSettings::autoplace_controls->water` in 2.0, with default params `frequency/richness/size` just like ores instead of being stand alone in root mgs.

### DIFF
--- a/features/antigrief.lua
+++ b/features/antigrief.lua
@@ -6,7 +6,30 @@ storage.original_last_users_by_ent_pos = {}
 
 Event.on_init(
     function()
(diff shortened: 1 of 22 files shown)


RedRafe:
Disclaimer: even if code updated to 2.0 looks exactly like 1.1's, I expect some maps to not behave exactly like 1.1 due to the noise changes introduced in 2.0. Especially water and cliffs, Earendel has tweaked their algorithm so same settings may produce 2 completely different maps in 1.1 and 2.0

grilledham (approved):

PR #1472, opened 2024-12-12
Experience revamp II

RedRafe:
Followup pr of #1470 

---

# Changes
- [x] refactored Gui appearance for Experience module
- [x] increased reward for character's manual mining speed formula (`24*0.9^L` > `24*0.9^sqrt(L)`)
- [x] added small compatibility with `redmew-data` mod:
  - if available from mod, play sound on each level reached
  - if available from mod, do not replace mining prod with bot capacity (the mod add the new tech instead)
  
# Preview

### New gui

By default, top container shows XP gaines so far + progressbar (similar as top part of old gui). Then, **Bonuses** and **Rewards** expandable containers are available to toggle (former is by default visible, and latter is hidden due to their relative lengths). Settings for what is visible and not are saved per-player (so no need to re-toggle each time)

> Bonuses
- Bonuses have been moved up before the market rewards, and now show the *actual* (total) bonus for the force (derived from vanilla research + bonus applied) instead of the only bonus added by the module (which in old system would show up as "+90 slots (up to 100)" but would never reach 100 in the gui cuz 10 of those slots were from toolbelt research.
- Formulas to show/compute bonuses have had small improvements to be more clear about what they represent
Screenshot from 2024-12-13 00-05-35

> Rewards
- The rewards table has been redesigned to use localised strings for item names, and color-coded buttons with coin price on the corner. A tooltip would show all the extra info (such as XP at which is unlocked)
Screenshot from 2024-12-13 00-05-51

Overall, I expect the performance of the module to be improved from previous version as the gui is not cleared & rebuilt on_tick, using a lot of LuaGuiElement.add calls to the api and re-drawing all the styles for the elements created. Only a few captions/tooltip texts are updated at runtime, while the styles are set only on creation and never touched again.

### DIFF
--- a/features/gui/experience.lua
+++ b/features/gui/experience.lua
@@ -36,6 +36,7 @@ local round_sig = math.round_sig
 local gain_xp_color = Color.light_sky_blue
 local lose_xp_color = Color.red
 
+local notify_name = 'notify_experience_level_up'
 local experience_lost_name = 'experience-lost'
 local on_bonuses, off_bonuses = '▼  Bonuses', '▲  Bonuses'
 local on_rewards, off_rewards = '▼  Rewards', '▲  Rewards'
@@ -69,6 +70,7 @@ Global.register(
     end
 )
 
+Settings.register(notify_name, Settings.types.boolean, true, 'experience.notify_caption_short')
 ScoreTracker.register(experience_lost_name, { 'experience.score_experience_lost' }, '[img=item.artillery-targeting-remote]')
 
 local global_to_show = storage.config.score.global_to_show
@@ -141,7 +143,7 @@ local function play_level_up_sound(force)
     end
     force_sounds[force.index] = game.tick + (config.sound.duration or 20 * 60)
     for _, player in pairs(force.connected_players) do
-        if Settings.get(player.index, 'notify_task') then
+        if Settings.get(player.index, notify_name) then
             player.play_sound(config.sound)
         end
     end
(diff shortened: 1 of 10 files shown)


features/gui/experience.lua
```diff
@@ -136,8 +137,12 @@ local function play_level_up_sound(force)
     if force_sounds[force.index] and game.tick < force_sounds[force.index] then
         return
     end
-    force.play_sound(config.sound)
     force_sounds[force.index] = game.tick + (config.sound.duration or 20 * 60)
+    for _, player in pairs(force.connected_players) do
+        if Settings.get(player.index, 'notify_task') then
```
grilledham: I think you should register your own setting for this. See `Settings.register(auto_play_cutscene_setting_name, Settings.types.boolean, true, auto_play_cutscene_setting_name)` as an example.

grilledham (commented):
I think this is good. So it's just I think there should be a new player setting for the diggy level up sound.

RedRafe:
The diggy cutscene caused 3 crashes right in the first few minutes of MP tonight on S13. At least, I suspect it was the cutscene, cuz no error was logged. It just straight up made the server crash and after 3 reloads, disabling the cutscene module let us play past 1st minute. Needs some debugging, possibly in MP too. SP I didnt notice anything weird even with debug mode enabled and I tested all kind of follow/skip cuscenes

features/gui/experience.lua
RedRafe: Thanks, added

grilledham (approved):

PR #1475, opened 2024-12-30
Fix map restart from poll

RedRafe:
Issue: DO would not load next modpack because I changed the keys in the available mods dictionary with the DO refactor and so none got selected when it needed to switch modpack.

This PR makes it so the keys and values of the modpack list match, so the restart module can index it correctly when picking next modpack.


map_gen/maps/danger_ores/modules/map_poll.lua
```diff
@@ -16,53 +16,53 @@ Global.register(data, function(tbl)
...
+    --{ name = 'danger-ore-industrial-revolution-3', display_name = 'Industrial Revolution 3 (default)', mod_pack = mod_packs.danger_ore_ir3 },
+    --{ name = 'danger-ore-industrial-revolution-3-grid-factory', display_name = 'Industrial Revolution 3 Grid Factory (squares)', mod_pack = mod_packs.danger_ore_ir3 },
+    --{ name = 'danger-ore-krastorio2', display_name = 'Krastorio2 (landfill)', mod_pack = mod_packs.danger_ore_krastorio2 },
+    { name = 'danger-ore-landfill', display_name = 'Landfill (all tiles)', mod_pack = mod_packs.danger_ore_normal },
+    { name = 'danger-ore-lazy-one', display_name = 'Lazy One (no handcraft)', mod_pack = mod_packs.danger_ore_normal },
+    { name = 'danger-ore-normal-science', display_name = 'Normal science (+ biters)', mod_pack = mod_packs.danger_ore_normal },
+    { name = 'danger-ore-omnimatter', display_name = 'Omnimatter (1 ore)', mod_pack = mod_packs.danger_ore_omnimatter },
+    { name = 'danger-ore-danger_ore_omnimattermatter-cages', display_name = 'Omnimatter Cages (1 ore + frames)', mod_pack = mod_packs.danger_ore_omnimatter },
```
grilledham: `danger-ore-danger_ore_omnimattermatter-cages` seems like a find+replace error.

RedRafe: Yup, ty

grilledham (approved):

PR #1482, opened 2025-01-03
Add rocket, inventory & quality utils

RedRafe:
# Changes

A bit of lib work to support quality around the codebase. Replaced all code repeated for `on_rocket_launched` with rocket utils to get sat count & ability to quickly filter by quality.

- [x] added `Inventory` utils
- [x] added `Quality` utils
- [x] added `Rocket` utils

### DIFF
--- a/utils/quality.lua
+++ b/utils/quality.lua
@@ -25,7 +25,7 @@ end
 ---@param q1 QualityID First quality ID
 ---@param q2 QualityID Second quality ID
 ---@param comparator string Comparison operator
----@return boolean|nil Result of comparison
+---@return boolean Result of comparison
 local compare = function(q1, q2, comparator)
   if type(comparator) ~= 'string' then
     error('Invalid comparator: expected string, got ' .. type(comparator))
--- a/utils/quality.lua
+++ b/utils/quality.lua
@@ -25,13 +25,21 @@ end
 ---@param q1 QualityID First quality ID
 ---@param q2 QualityID Second quality ID
 ---@param comparator string Comparison operator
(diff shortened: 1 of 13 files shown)


utils/quality.lua
```diff
@@ -0,0 +1,52 @@
...
+-- Compare two quality identifiers based on the specified comparator
+---@param q1 QualityID First quality ID
+---@param q2 QualityID Second quality ID
+---@param comparator string Comparison operator
+---@return boolean|nil Result of comparison or nil on error
+local compare = function(q1, q2, comparator)
+  local l1 = level(q1)
+  local l2 = level(q2)
+
+  local success, result = pcall(OPS[comparator], l1, l2)
```
grilledham: What is `pcall` protecting against? It seems like all the code that could fail at this point is the code that is contained in this file and would therefore be a bug we should fix rather than ignore. Unless I'm missing something?

grilledham (approved):

utils/quality.lua
RedRafe: A wrongful comparator I guess, but I did implement pcall before refactoring `level` function & making it a bit fail safer so calling compare with invalid comparator seems just the last usecase left. I could just return if no function is found

grilledham: Wouldn't an invalid comparator be a bug. I think it would be better to error in such a case to make it easier to track down the bug.

RedRafe: Updated w/ error message & type checks

grilledham (approved):

PR #1484, opened 2025-01-14
Add DO:Coal Maze

RedRafe:
Adds a new Danger Ores preset: Coal Maze
Uses 2 presets we already have: maze gen (50x50, 3 wide) and coal + patches (DO).
Terraforming module has been disabled for this map.

Hopefully many trains will be used here!
Couple of screenshots:

> Close up
Screenshot from 2025-01-14 21-18-18

> Whole view
Screenshot from 2025-01-14 21-23-12

Biters seem dense but it's just the default enemy curve we have on every DO map. We'll see I guess. SHouldnt be a problem in peaceful mode tho

### DIFF
--- a/map_gen/maps/danger_ores/changelog.lua
+++ b/map_gen/maps/danger_ores/changelog.lua
@@ -87,5 +87,8 @@ return [[
 
   2025-01-13:
       - [DO] Added quality settings to rocket launched goals
-      - [DO_Permanence] Increased manual mining speed to +900%
+      - [DO:Permanence] Increased manual mining speed to +900%
+
+  2025-01-14:
+      - [DO:CoalMaze] Added Coal Maze preset
 ]]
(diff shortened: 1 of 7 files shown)


grilledham (approved):
Looks good. One suggestion, I'd be tempted to center the spawn in one of the maze chunks, looks a bit odd being clipped by the wall. I'll leave it up to you if you want to change that. 

You have my approval to merge either way.

RedRafe:
It does depend on how the random maze ends up being, but changing to square and translating of 2,2 kinds fixes it. If I translate it more the starting cross will be offset a lot since the spawn area does not have coordinates x,y to translate it.

I should change the map builder and have whole map accept additional x,y params to shift it whole, but maybe can be done in another PR

Screenshot from 2025-01-14 21-45-25
Screenshot from 2025-01-14 21-46-15

grilledham (approved):

PR #1485, opened 2025-01-24
Loop handlers backwards

RedRafe:
# Changes
- [x] event handlers are prepended to the handlers table
- [x] handlers are looped backwards on events

# Issue
Removing functions at runtime is not remove-safe when looping the handlers "straight", while looping them backwards, would allow any handler to remove itself at runtime, without the event_core skipping any functions.

Example: `F1` and `F2` are added at runtime and each of them removes itself.
When the event triggers, `F1` removes itself, then `F2` is effectively skipped because the handler has now moved to next index, which is nil now.

By looping them backwards (and so, registering them backwards to maintain the load order of all modules), `F1` and `F2` can each remove themselves without the event module skipping any handler anymore.

Test code:

```lua
/c
local Event = _G.package.loaded['__level__/utils/event.lua']
storage.ftick = 123 
local f1, f2
f1 = function(event)     game.print("f1")     local Event = _G.package.loaded['__level__/utils/event.lua']     Event.remove_removable_nth_tick_function(storage.ftick, f1) end
f2 = function(event)     game.print("f2")     local Event = _G.package.loaded['__level__/utils/event.lua']     Event.remove_removable_nth_tick_function(storage.ftick, f2) end

Event.add_removable_nth_tick_function(storage.ftick, f1)
Event.add_removable_nth_tick_function(storage.ftick, f2)
```

I'll leave it to you Grilled if you think this is an edge case worth fixing or if it's something we don't care much.

### DIFF
--- a/utils/event_core.lua
+++ b/utils/event_core.lua
@@ -22,14 +22,14 @@ local script_on_nth_tick = script.on_nth_tick
 local call_handlers
 if _DEBUG then
     function call_handlers(handlers, event)
-        for i = 1, #handlers do
+        for i = #handlers, 1, -1 do
             local handler = handlers[i]
             handler(event)
         end
     end
 else
     function call_handlers(handlers, event)
-        for i = 1, #handlers do
+        for i = #handlers, 1, -1 do
             local handler = handlers[i]
             local success, error = pcall(handler, event)
             if not success then
@@ -89,7 +89,7 @@ function Public.add(event_name, handler)
         event_handlers[event_name] = {handler}
         script_on_event(event_name, on_event)
     else
-        table.insert(handlers, handler)
+        table.insert(handlers, 1, handler)
         if #handlers == 1 then
             script_on_event(event_name, on_event)
         end
@@ -103,7 +103,7 @@ function Public.on_init(handler)
         event_handlers[init_event_name] = {handler}
         script.on_init(on_init)
     else
-        table.insert(handlers, handler)
+        table.insert(handlers, 1, handler)
         if #handlers == 1 then
             script.on_init(on_init)
         end
@@ -117,7 +117,7 @@ function Public.on_load(handler)
         event_handlers[load_event_name] = {handler}
         script.on_load(on_load)
     else
-        table.insert(handlers, handler)
+        table.insert(handlers, 1, handler)
         if #handlers == 1 then
             script.on_load(on_load)
         end
@@ -131,7 +131,7 @@ function Public.on_configuration_changed(handler)
         event_handlers[configuration_changed_name] = {handler}
         script.on_configuration_changed(configuration_changed)
     else
-        table.insert(handlers, handler)
+        table.insert(handlers, 1, handler)
         if #handlers == 1 then
             script.on_configuration_changed(configuration_changed)
         end
@@ -145,7 +145,7 @@ function Public.on_nth_tick(tick, handler)
         on_nth_tick_event_handlers[tick] = {handler}
         script_on_nth_tick(tick, on_nth_tick_event)
     else
-        table.insert(handlers, handler)
+        table.insert(handlers, 1, handler)
         if #handlers == 1 then
             script_on_nth_tick(tick, on_nth_tick_event)
         end
--- a/utils/event.lua
+++ b/utils/event.lua
@@ -48,27 +48,6 @@
 -- It's not an error to register the same token multiple times to the same event, however when
 -- removing only the first occurrence is removed.
 --
--- ** Event.add_removable_function(event_name, func) **
---
--- Only use this function if you can't use Event.add_removable. i.e you are registering the handler at the console.
--- The same restrictions that apply to Event.add_removable also apply to Event.add_removable_function.
--- func cannot be a closure in this case, as there is no safe way to store closures in the global table.
--- A closure is a function that uses a local variable not defined in the function.
---
--- @usage
--- local Event = require 'utils.event'
---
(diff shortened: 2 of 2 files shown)


utils/event_core.lua
```diff
@@ -89,7 +89,7 @@ function Public.add(event_name, handler)
         event_handlers[event_name] = {handler}
         script_on_event(event_name, on_event)
     else
-        table.insert(handlers, handler)
+        table.insert(handlers, 1, handler)
```
grilledham, line 92: I don't think it's worth adding the handlers in reverse order. I'm guessing this is to technically keep the behaviour the same, but in practice we shouldn't be relying on the order events are added.

Part of me doesn’t like this because it means adding events goes from O(1) to O(n), though in practice we don’t add enough events for this to matter.

grilledham (commented):
If you agree and action my comment, I'm happy for you to merge.

grilledham:
> I'll leave it to you Grilled if you think this is an edge case worth fixing or if it's something we don't care much.

I think it's worth fixing this edge case, good catch in spotting it.

utils/event_core.lua
RedRafe: True, and at first I just implemented the backward looping thinking the same.

However, maps wouldn't even load and break instantly without the reverse registering. I think there is some dependency in the core modules (storage, map loader, tokens...) that has to happen in a certain order `on_init` at least (maybe on_load as well?). For other events it shouldn't matter instead.

I can dig deeper on what modules have these issues during the next days. I'm in no rush.
Module has been fine for years and we're not adding any handlers past on init at the moment on any maps

RedRafe:
The issue with handling order is that they reflect the files dependency:

If scenario requires featureA and featureB, they will be loaded/required first, and so will their handlers.
I'm not sure it is currently possible to convert the whole module into a load-order-independant system: as pretty much all core features (event, global, redmew_surface) need to happen before the map code.

For example, a DO map has an init handler which references RS, RS will be null since it also reference Global:

https://github.com/Refactorio/RedMew/blob/151c3cd10a8bccbbf15a204d5b93ef07597a993f/map_gen/shared/redmew_surface.lua#L285-L287

RedRafe:
My guess is that event handlers should not rely on call order but be independent, which is probably true 99% of the time (`on_chunk_generated` being the outlier).

> edit: I checked, preserving the registration order in call order for init and events works, preserving for on_init but not for other events works too, not preserving for init and all events doesnt work.

And, for core modules that need to have priority, it is best practice that they generate their own event to be called next, and Generate module is actually the best example for this case: Generate module registers for `on_chunk_generated` and raise a copy of the event with new name, after they've take action. And all maps would register to the Generate's event name instead of `defines.events.on_chunk_generated`, resulting n that event having 1 handler only from Generate, and priority is always guaranteed even without the `table.insert(tbl, 1, element)` hack.

This is not possible for `on_init`, as init doesn't raise any new events until it's finished. We could have a custom event at tick 1 that works as "redmew_init" where all the non-core modules/maps can register to do their stuff, regardless of call order, just after the core modules have been primed. Dunno if it's worth the refactor tho, as it may raise new issues...

grilledham:
It’s unfortunate that we depend on the order of event handlers, but it would seem like a lot of work/risky to change that.

I wonder if instead we could solve this in another way. What if, when a handler is removed, instead of removing it we add another special handler to the end of the array. We also track the handler that needs to be removed in `needs_removing table`. That special handler would then remove all the handlers from the `needs_removing` table. The complication is what if another event handler is added after a remove. So I think the special handler would have to check to see if it’s the last handler in the array and if not add itself to the end of the array and do nothing. I would also have the special handler add itself to the `needs_removing` table. Then when the special handler runs for the second time it would remove all the handlers that need to be removed.

Alternatively we just add a comment saying event handlers can’t remove themselves. Are there situations where that would present a problem?

RedRafe:
The issue only raises when a function needs to remove a handler registered to the same event that is triggering, removing handlers of any other event is no problem. And I think it's only common that a handler would remove itself when an event is fired or remove handlers for other different events, but almost never remove another handler of the same event (in which case, you'd have 2 handlers registered on the same event on the same context, so why not just having 1 function from that context?).

But this is simply massive speculation of the use of Event with token functions and removable functions. I don't think we use it anywhere currently, other than a couple of token functions that add/remove themselves.

Bonus: the use of add/remove removable_function is currently broken because it's not possible to save functions in the storage table anymore. So the game would not allow to save

RedRafe:
> the use of add/remove removable_function is currently broken because it's not possible to save functions in the storage table anymore. So the game would not allow to save

I've shared some ideas about the refactor of this on https://github.com/Refactorio/RedMew/pull/1486, but I'd be more inclined to generally keep things as they are: acknowledge some dependency are in place, and if we want to keep `add/remove _removable_function`, to pass a string function + name ID to current endpoints.

grilledham:
> the use of add/remove removable_function is currently broken because it's not possible to save functions in the storage table anymore. So the game would not allow to save

I take it no one is using it as it's broken? Then I think we just remove it.

RedRafe:
Removed logic for those but kept the endpoint with error message (just in case).

LMK if that's okay for you and if you still think it's worth to prepend handlers and loop them backwars on events (I think it's safer to cover 1 more possible use and should not impact performance). But up to you. It's not a big deal anyways, very edgecase

grilledham (approved):
LGTM, I think looping the handlers backwards and prepending the arrays is the most sensible compromise.

PR #1489, opened 2025-03-13
Rework map_selection.sample.lua

maritaria:
- Include a list of available maps
- Rewrite instructions

### DIFF
--- a/map_selection.sample.lua
+++ b/map_selection.sample.lua
@@ -16,7 +16,9 @@ return require 'map_gen.maps.default'
     return require 'map_gen.maps.danger_ores.presets.danger_ore_gradient'
     return require 'map_gen.maps.crash_site.presets.normal'
     return require 'map_gen.maps.crash_site.presets.desert'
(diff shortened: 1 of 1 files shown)


maritaria (commented):
Note: I've not yet gone through all available options and verify that each works. I basically made a list of all generators in paths `map_gen/maps/*.lua` and `map_gen/maps/*/preset/*.lua`. Some areas are not covered like the april-fools one.

I made this PR because I think this format is a bit more noob-friendly.
Perhaps having the list of options in a giant comment for them to copy/paste would be easier for them to use than to uncomment lines, but I'll leave that up to you.

maritaria:
(closed, will reopen with similar idea).

Turns out there is something wrong with having code after top-level return.

EDIT: Lua complains about statements after top-level return. So this approach would only work if user only ever uncomments exactly one `return require` line. I'll rework to more closely resemble old setup but with full list of all options in the comment area.

PR #1493, opened 2025-05-19
Update Diggy scenario

RedRafe:
## List of changes:

- added more items to starting inventory
- added Mk2 shields to Market
- increased Experience's character mining speed progression from .05 to .20
- increased Experience's max character mining speed from 20x to 30x
- added crafting speed, running speed, character reach to Exp bonuses
- enabled landfill back (now a minable tile, not subject to griefing)
- fixed a bug when bot mining entities in `market_chest.lua`
- fixed that player would get stuck inside spawned enemies
- removed redundant logic of `big-rock` (introduced with the 2.0 port replacing strings)


grilledham (approved):

grilledham (approved):

RedRafe:
Added crafting speed, running speed, character reach to Exp bonuses. These in conjuction with the 4x buff on manual mining speed should help speed up the scenario, as QoL, because right now it takes like 60h to get to purple/yellow science if playing solo

A better system of bonuses may be implemented as RPG, taking in account the # of players playing. Just an idea for the future

PR #1500, opened 2025-09-14
Add calculator/technology feature

RedRafe:
Adds a calculator for technologies to compute total cost of a research, prerequisites breakdown, and navigation of tech tree from a command line.
(bonus) also adds an admin command to open the UI for another player

- back/forward navigation withing the player's history of searches
- left click to select target technology (or use internal name from command)
- on open/close, history is maintained (reopening the window with nil will have latest search as default)
- right click on tech icon to clear the search history
- recursive navigation of technologies by clicking the icon of target tech ion its dependencies breakdown

> Demo

https://github.com/user-attachments/assets/c6133913-6adc-449a-b675-30cd74650577

### DIFF
--- a/features/gui/calculator/technology.lua
+++ b/features/gui/calculator/technology.lua
@@ -36,7 +36,6 @@ local function set_history(player_index, technology_name)
     elseif prototypes.technology[technology_name] ~= nil then
         h:add(technology_name)
     end
-    history[player_index] = h
 end
 
 local function dict_to_array(dict)
@@ -421,6 +420,7 @@ end)
 Command.add('calculator-technology', {
     description = 'Computes the cost in science packs to research target technology',
     arguments = { 'technology' },
+    default_values = { technology = '' },
     allowed_by_server = false,
     required_rank = Ranks.guest,
     capture_excess_arguments = true,
--- a/features/gui/calculator/technology.lua
+++ b/features/gui/calculator/technology.lua
@@ -1,4 +1,4 @@
--- This feature adds a command "/technology-cost" that shows a small popup
+-- This feature adds a command "/calculator-technology" that shows a small popup
 -- with the breakdown cost to research target technology.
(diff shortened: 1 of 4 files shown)


features/gui/calculator/technology.lua
```diff
@@ -0,0 +1,456 @@
...
+end
+
+local function set_history(player_index, technology_name)
+    local h = get_history(player_index)
+    if technology_name == nil then
+        h:clear()
+    elseif prototypes.technology[technology_name] ~= nil then
+        h:add(technology_name)
+    end
+    history[player_index] = h
```
grilledham: `history[player_index] = h` probably isn't needed in `set_history` as `get_history` sets this.

features/gui/calculator/technology.lua
```diff
@@ -0,0 +1,456 @@
...
+    end
+    set_history(event.player_index, event.element.tags.name)
+    draw(event.player)
+end)
+
+-- == COMMANDS ================================================================
+
+Command.add('calculator-technology', {
+    description = 'Computes the cost in science packs to research target technology',
+    arguments = { 'technology' },
```
grilledham, line 419: Maybe this argument should be optional.
```suggestion
    arguments = { 'technology' },
    default_values = { technology = '' },
```

grilledham (approved):
Looks good, a few suggestions.

features/gui/calculator/technology.lua
RedRafe: Ah yes, ty. I was wondering what the correct syntax was for optional params in commands definitions, I completely missed in the module's docs

RedRafe:
I think opt param is okay for "calculator-technology" since it should open the UI anyways to pick the tech, while for "c-t-for-player" it will be okay if no defaults are set since it shouldnt activate if the tech or player are invalid

PR #1504, opened 2025-09-22
Improve mod compatibility

RedRafe:


grilledham (approved):

RedRafe:
heh I opened a ticket with SE devs for this and they said they were gonna fix it, still no updates on it. SO Im reverting-the-revert and will update in the future if they will change it for real

PR #1505, opened 2025-10-10
Add simple rpg to CS

RedRafe:
Add a simple rpg system to CS (similar to Diggy's experience) with military ranks a buffs.

50 levels, each level a small random buff is granted, and each 5 levels a bigger one preset is granted. Players accrue points by killing, capturing and upgrading outposts themselves (no XPs from turrets, only melee, vehicles, landmines, capsules).

UI on top left has rank level, rank name, buffs granted history and hint at non-standard buffs (standard are in the bonuses tab, non standard are listed as aura/regen/whataver)
Screenshot from 2025-10-10 11-22-56

Possible to testi it out in DEBUG mode with `level-up` and `level-up-all` cmds

### DIFF
--- a/map_gen/maps/crash_site/outpost_builder.lua
+++ b/map_gen/maps/crash_site/outpost_builder.lua
@@ -10,6 +10,7 @@ local Donator = require 'features.donator'
 local RS = require 'map_gen.shared.redmew_surface'
 local Server = require 'features.server'
 local CrashSiteToast = require 'map_gen.maps.crash_site.crash_site_toast'
+local RPG = require 'features.gui.rpg_military'
 
 local table = require 'utils.table'
 --local next = next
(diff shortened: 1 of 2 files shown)


map_gen/maps/crash_site/outpost_builder.lua
```diff
@@ -1090,6 +1092,18 @@ local function do_capture_outpost(outpost_data)
...
+    local nearests = RS.get_surface().find_entities_filtered{
+        position = { x = (area.bottom_right.x - area.top_left.x) / 2, y = (area.bottom_right.y - area.top_left.y) / 2 },
+        type = 'character',
+        name = 'character',
+        force = 'player',
+        limit = 1,
+        radius = 32 * 10,
+    }
+    if #nearests == 1 then
+        game.print('Awarded outpost to '..nearests[1].player.name)
```
grilledham: `'Awarded outpost to '..nearests[1].player.name` I wonder if it would make more sense to include this in the toast, rather than having an extra game.print for it. Just a suggestion/thought. Could also include it in the `to_discord_bold` too.

grilledham (approved):

map_gen/maps/crash_site/outpost_builder.lua
RedRafe: Mmh yeah, maybe I could merge the two and change it to "Outpost X has been captured by Y"

PR #1511, opened 2026-01-06
Update Discord link to as redmew.com domain is down

knot2006:


RedRafe:
Thanks, this has already been fixed for next release along with some more changes to enhance the user experience!

(btw the domain has simply changed from redmew.com to refactorio.de)

PR #1512, opened 2026-01-07
Moderation update

RedRafe:
This PR implements 3 major changes:

1. Added `Moderator` rank
2. Reworked commands/utils to accomodate the new rank
3. GUIs update & extension

Detailed explanation of each bullet point below.

---
### 1. Addition of Moderator rank
- [X] Added `moderator` rank to `Ranks`. Value = 25, between admin(30) and regular(25)
- [X] Added a color and localization for mod rank (yellow)
- [X] Decreased the required rank for some commands from admin to mod (only the ones needed to manage the game)

### 2. Rework of core libs to accomodate the new rank
- [X] Changed core utils to allow moderators to be considered as valid actors (not just admin-only)
- [X] Added commands: moderator-add/remove, mute/unmute, purge, trust-all, kick (admins & mods)
- [X] Added new custom event for the admin panel core module to allow children to react to its events
- [X] Added new APIs for admin panel core module for children to easily access those properties

### 3.A GUIs update
- [X] Updated RedMew's URL from redmew.com to refactorio.de in game (Wiki excluded)
- [X] Uniformed commands format
- [X] Updated RedMew Info panel (`features/gui/info`) with all features available: ranks, market, autofill, polls, tasks, shortcuts, paint brush, production HUD, train station teleport, score
- [X] Changed sprite for Tags feature from text to bookmark
- [X] Restyled map restart config window for admins with a grid
- [X] Uniformed topbar's tooltips for all features
- [X] Renamed "softmods" to "Features" in RedMew Info
- [X] Removed old localization entries from non-EN locales when wrong/unused

### 3.B GUIs extension
- [X] Added localization for all commands (debug-only excluded)
- [X] Added tooltip localization for all commands (debug-only excluded) & for admin panel usage
- [X] Added "**Moderation Guide**" section in Admin panel to provide support to all mods/admins
- [X] Added **Guide/Ranks** page: breakdown of each rank's permissions & tools
- [X] Added **Guide/Moderation** page: a wiki-like overview of all admin panel features
- [X] Added **Guide/Commands** page: full list, breakdown & search of in-game commands (100+)
- [X] Added **Guide/Server** page: list of useful console commands/procedures for admins to manage the server
- [X] Added **Guide/Resources** page: link to the github's wiki/moderation guide & our mission statement

---
## Future works

1. I'm marking this PR as a draft while I finish implementing: Guide/Moderation and Guide/Server (mainly wiki-style).
2. Couple old modules have massive code repetitions in them that could be (maybe) simplified in the future (or not, certainly not in this PR).
3. I'm considering if the Guide ("in-game wiki") page should be made available for everyone in game (or keep it mods/admins, probably not in this PR, but support for it will be in place)
4. I will rework the style of the Guide to allow to open and keep on the side each page individually for cross reference instead of opening everything in a long scroll pane like it is presented now (**before this PR gets merged**).

Let me know if anything from the core functionalities/new functionalities needs better explanation or rework


features/admin_commands.lua
```diff
@@ -158,6 +206,95 @@ local function probation_remove(args, player)
...
+
+    if not maybe_target_player and not know_player_or_rerun(target_name, actor, 'moderator') then
+        return
+    end
+
+    -- Prevent promoting anyone who is below regular (guest/probation)
+    if Rank.less_than(target_name, Ranks.regular) then
+        Game.player_print({'admin_commands.moderator_add_fail_low_rank'}, Color.fail, player)
+        return
+    end
```
grilledham: Why, doesn't this just create more work for admins to have to first promote someone to regular then to moderator?

features/gui/admin_panel/moderation_pages.lua
```diff
@@ -0,0 +1,473 @@
...
+    path(parent, {'guide', 'General actions', 'Cheat mode'})
+    path(parent, {'guide', 'General actions', 'Show reports'})
+    path(parent, {'guide', 'General actions', 'Blueprints ON/OFF'})
+end
+
+-- == COMMANDS ================================================================
+
+local search_field_name = Gui.uid_name()
+
+local base_commands = {
```
grilledham: Would it be possible to generate theses dynamically? Just thinking about maintenance effort to keep this update to date when adding new commands. We should at least be able to get the redmew commands dynamically given we have our own custom module for registering commands, and those are the ones that are most likely to change in the future.

features/gui/admin_panel/functions.lua
```diff
@@ -72,6 +72,18 @@ function Actions.ban_player(target_name, reason, admin)
   Report.ban_player(player, reason, admin)
 end
 
+---@param target_name string
+---@param reason? string
+---@param admin? LuaPlayer
+function Actions.kick_player(target_name, reason, admin)
```
grilledham, line 78: I probably should have noticed this before, but using `function Actions.*` along side `local Actions = require 'features.admin_commands'` slightly worries me. If we created a function here that had a name clashed with the the keys in the `'features.admin_commands'` table, wouldn't it override it? Which is probably not what we want given it's a shared object. I'd be happier if we created a copy of the table when setting `Actions` given we plan to mutate it.

grilledham (approved):
Looked through the code and looks great. Really appreciate the effort you have put into this. I'd be happy for it to merge as is, but I do I have a few comments/thoughts though.

I also didn't test it. Perhaps you can get Tigress to help test it?

features/admin_commands.lua
RedRafe: True and false: this should have been a `< Guest` condition so it doesnt promote someone in jail/probation to mod by mistake

features/gui/admin_panel/moderation_pages.lua
RedRafe: For the base ones - not really, for two reasons:
1. it is not known if a command is admin only or anyone can run it other than reading it from the official Wiki/testing it out in-game (I did both)
2. the command-name/command-help are formatted differently

For RedMew's: yes, it could be posssible to keep it saved in the `utils.command` module & then have the list automatically populated from registered commands (only needs a few extra formatting steps but really doable)

features/gui/admin_panel/functions.lua
RedRafe: Agreed

features/admin_commands.lua
RedRafe: Changed to guest (I do like the additional check to prevent misclicks, especially since I do most stuff from Admin UI and not console command where I just click buttons instead of typing it all out (which could be prone to more human errors) :)))

grilledham: For guest makes sense. I can see the logic in preventing probation to moderator.

features/gui/admin_panel/moderation_pages.lua
RedRafe: Took a fair bit of formatting among all the localizations but I managed to autobuild the redmew command list from `utils.command`.

The only change is that only commands that effectively loaded are now visible, while before "all" commands (even from optional modules not enabled) were listed. This is better for playability, and a bit worse for developing. As the user are the final target for this guide/wiki, I'm happy with the changes and the tradeoff.

RedRafe:
Nevermind, the commands registered by all the modules loaded after admin gui are not listed. May need to revert last 2 commits or find a way to cache commands after `control.lua` finishes/first load of the page

RedRafe:
I think I'm happy with the result. This is how it now works: Admin panel > Guide > Secondary pages (that can be kept open simultaneosly and closed individually). When closing the main window (with 'X' or ESC) all other secondary pages are closed too. Opening toggle button works too.

Screenshot from 2026-01-09 17-45-49

---

Future works:
- [ ] expand on the wiki with all the buttons info
- [ ] maybe spawn Admin panel a bit more on the left so there's more space, but whatever

grilledham (approved):
Amazing stuff. I had a play around with it and it looks good to me.

PR #1519, opened 2026-01-11
Reduce save size

RedRafe:
# Changes
- [X] Removes player data on: kick/ban/AFK > 7
- [X] Add `on_player_removed` handler for all modules that store data in global storage

### DIFF
--- a/features/save_manager.lua
+++ b/features/save_manager.lua
@@ -1,7 +1,7 @@
 local Event = require 'utils.event'
 local config = require 'config'.save_manager
 
-local DEFAULT_INACTIVE_INTERVAL = 60 * 60 * 60 * 24 * 7 -- a week
+local DEFAULT_INACTIVE_INTERVAL = 60 * 60 * 60 * 24 * 30 -- a month
 
 local function remove_player_data(event)
     local player = game.get_player(event.player_index)
--- a/features/save_manager.lua
+++ b/features/save_manager.lua
@@ -13,7 +13,6 @@ local function remove_player_data(event)
     game.remove_offline_players({ player.index })
 end
 
-Event.add(defines.events.on_player_kicked, remove_player_data)
 Event.add(defines.events.on_player_banned, remove_player_data)
 
 Event.on_nth_tick(config.inactive_interval or DEFAULT_INACTIVE_INTERVAL, function()
--- a/features/save_manager.lua
+++ b/features/save_manager.lua
@@ -1,0 +1,33 @@
+local Event = require 'utils.event'
+local config = require 'config'.save_manager
+
+local DEFAULT_INACTIVE_INTERVAL = 60 * 60 * 60 * 24 * 7 -- a week
+
+local function remove_player_data(event)
+    local player = game.get_player(event.player_index)
+    if not (player and player.valid) then
+        return
+    end
+
+    game.print({ 'save_manager.remove_player', player.name })
+    game.remove_offline_players({ player.index })
+end
+
+Event.add(defines.events.on_player_kicked, remove_player_data)
+Event.add(defines.events.on_player_banned, remove_player_data)
+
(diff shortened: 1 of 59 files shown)


features/save_manager.lua
```diff
@@ -0,0 +1,33 @@
...
+    local player = game.get_player(event.player_index)
+    if not (player and player.valid) then
+        return
+    end
+
+    game.print({ 'save_manager.remove_player', player.name })
+    game.remove_offline_players({ player.index })
+end
+
+Event.add(defines.events.on_player_kicked, remove_player_data)
```
grilledham: Doing this on kick might be a bit harsh.

grilledham (commented):
Adding `on_player_removed` handlers for the modules is great, we should do this in case a scenario feature or mod removes player data.

The part about removing player data for afk players I need more convincing about.
1. Do we have evidence that this makes a meaningful different to save file size? I would think that only really the player's blueprints would take up a noticeable amount of space. And I wonder if we can just remove those for afk players?
2. I'm nervous about having this on by default. It will be difficult to know what the right default timeout should be. I could see some long term maps where a player might only play once per weekend. And one day they might play Saturday then the next week they play Sunday. Should those players have their data removed every time they play?
3. When this is on do banned players still drop their inventories into a corpse? A common reason to ban players is they are stealing high value items. We wouldn't want such high value items to be lost on ban. Also for danger ore maps destroying items could be seen as cheating.
4. Does this play nicely with the afk corpse feature? I wonder if this should only be on by default if we also have the afk corpses on too. That way items aren't lost.

RedRafe:
>     1. Do we have evidence that this makes a meaningful different to save file size? I would think that only really the player's blueprints would take up a noticeable amount of space. And I wonder if we can just remove those for afk players?

Yes, indeed, BPs are the biggest concern. I've already opened a ticked on the forum to expose the remove BP api to servers (rn it is only available to "humans" and it's 1 or everybody).
[Factorio API requests](https://forums.factorio.com/viewtopic.php?t=132357)
Goal is to remove BPs, but removing players is the only way to do it if the devs dont expose the API to script.

>     2. I'm nervous about having this on by default. It will be difficult to know what the right default timeout should be. I could see some long term maps where a player might only play once per weekend. And one day they might play Saturday then the next week they play Sunday. Should those players have their data removed every time they play?

Goal is not to harm players. This is a cleanup service that should reduce filesize without interferring w/ players. BP storage size is only a concern for eternal maps and "long" maps where hundreds of player join in its lifespan. This does not happen on weekly/daily maps. I'm happy to increase "AFK" time to 1 month, as in the long run, it'll achieve its goal of pruning unused BP storages.

>     3. When this is on do banned players still drop their inventories into a corpse? A common reason to ban players is they are stealing high value items. We wouldn't want such high value items to be lost on ban. Also for danger ore maps destroying items could be seen as cheating.

Yes, player corpse module has been correctly integrated and instantly calls the drop inventory content *before* the player is removed (thats why in some module it is used `on_PRE_player_removed` while other only do `on_player_removed`, to still access the player safely.

>     4. Does this play nicely with the afk corpse feature? I wonder if this should only be on by default if we also have the afk corpses on too. That way items aren't lost.

Ditto 3.

---

Additionally, we could incorporate this with rank system and maybe only remove players that are below `Regular` (so we can still tell script to "save" players by promoting them. I would still drop auto-trusted in this case.

grilledham:
> I'm happy to increase "AFK" time to 1 month

If the default was 1 month, then most of my concerns disappear and I would be happy with this being included.

> I've already opened a ticked on the forum to expose the remove BP api to servers...

Nice, hopefully this will get added, as this would allows us to be more aggressive with reducing map size.

features/save_manager.lua
RedRafe: We **rarely** use /kick, most of the times we either jail or ban directly. IDK how useful it is since kicked players can still rejoin the save the next second they're kicked, so I guess I could not remove player data when kicked

PR #1520, opened 2026-01-11
Fix looting alert

RedRafe:


grilledham (approved):
Haven't tested, but code looks good. 👍

PR #1521, opened 2026-01-11
Refactor Tags ui

### LINKED ISSUE #1517: Tags update
- [x] refactor ui
- [x] add AFK to default tags list

RedRafe:
Closes #1517 

Screenshot from 2026-01-11 15-26-03

### DIFF
--- a/features/gui/tag_group.lua
+++ b/features/gui/tag_group.lua
@@ -98,7 +98,7 @@ local function get_size(players, show_offline)
     local size = 0
 
     if not players then
-        return ''
+        return size
     end
 
     if show_offline then
@@ -112,11 +112,7 @@ local function get_size(players, show_offline)
         end
     end
 
-    if size == 0 then
-        return ''
-    else
-        return ' (' .. size .. ')'
-    end
+    return size
 end
 
 local main_button_name = Gui.uid_name()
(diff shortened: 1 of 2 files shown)


features/gui/tag_group.lua
```diff
@@ -38,6 +38,12 @@ local function notify_players(message)
     end
 end
 
+local function compare_tag(player, tag_name)
+    local old_tag = player.tag
+    local new_tag = '[' .. tag_name .. ']'
+    return (old_tag == new_tag) or (old_tag == tag_name)
```
grilledham, line 44: `a == b or b == a`? I don't thin the order you do the equality check matters?

features/gui/tag_group.lua
```diff
@@ -112,11 +118,7 @@ local function get_size(players, show_offline)
         end
     end
 
-    if size == 0 then
-        return ''
-    else
-        return ' (' .. size .. ')'
-    end
+    return size
```
grilledham, line 121: For the online case we could probably just use `table.size(game.connected_players)` or something like that.

grilledham (approved):
This is fine, but I do have a few suggestions

features/gui/tag_group.lua
RedRafe: It's `(a == b) or (a == c)` ?

grilledham: May bad 😅

features/gui/tag_group.lua
RedRafe: Mmh I think the gist of this function is:

```lua
local player_with_X_tag = player_tags[tag]
local count = size(player_with_X_tag)
```

to count how many players (offline/offline) have that tag. I could filter from `game.connected_players` but I think using the cached player idx as it was used it's easier here. I just converted the function from returning string `(count)` to returning number `count` as Im showing the count in the `button.number` instead of concatenating the string beside the tag caption name

grilledham: Makes sense. I should have read this a bit more carefully. No change required.

PR #1522, opened 2026-01-11
Fixes for DO/BobAngels maps

RedRafe:
# Changes
- [X] DO maps now automatically list all the trees and fishes available from game prototypes
- [x] changelog is fetched from server if scenario does not provide one itself
- [X] fixed DO/BA maps not spawning ores, trees & resources correctly

### DIFF
--- a/map_gen/shared/builders.lua
+++ b/map_gen/shared/builders.lua
@@ -2135,45 +2135,9 @@ local water_tiles = {
 
 Builders.water_tiles = water_tiles
 
---- Returns fish only
+--- Returns any of the fishes prototypes that have been added to the game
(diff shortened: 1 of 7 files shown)


map_gen/shared/builders.lua
```diff
@@ -2163,6 +2172,44 @@ function Builders.fish(shape, spawn_rate)
     end
 end
 
+--- Returns any of the fishes prototypes that have been added to the game
+function Builders.fishes(shape, spawn_rate)
```
grilledham: I'd be happy for this to just be the default and only implementation of `Builders.fish`

grilledham (approved):

PR #1532, opened 2026-01-23
diggy: fixed error on open popup on placing solar panel

Sascha-Kuehl:
There is a popup, that should be shown, when a player places a solar panel in diggy scenario, so he get informed, that it will not work. But the code raises an error instead of showning the popup, because the required statement was used directly in the event function (what is not working). So extracting the required statement to the top of the file and use it in event function does the trick. So Popup is shown again.


grilledham (approved):
LGTM, thanks for the fix ❤️

PR #1535, opened 2026-01-26
Gui updates

RedRafe:
Just a touch of restyling for our main GUIs of toolbar + file formatting to give every gui the same style/theme.
No logic changes.

---

## Changes
- [X] unified frame styles with inner frames in all guis
- [X] replaced draggable frames with non_draggable_frames for center guis
- [X] added subheaders where needed
- [X] improved tooltips overall (localized tips, colors, fonts, shortcuts) 
- [X] restyled buttons to make them easier to read
- [X] restyled some GUIs to be a copy of vanilla's
- [X] added some input sanitization where easily supported

---

## Before/After of the changes

> Before
Screenshot from 2026-01-26 17-31-42

> After
Screenshot from 2026-01-26 15-47-44

### DIFF
--- a/features/gui/autofill.lua
+++ b/features/gui/autofill.lua
@@ -82,7 +82,7 @@ local function toggle_main_frame(event)
 
         local grid = enabled_ammos_flow
             .add { type = 'scroll-pane', style = 'deep_slots_scroll_pane' }
-            .add {type = 'table', column_count = 5, style = 'filter_slot_table'}
+            .add { type = 'table', column_count = 5, style = 'filter_slot_table' }
 
         for name, enabled in pairs(Autofill.get_player_ammos(player_index)) do
             local button =
--- a/features/gui/autofill.lua
+++ b/features/gui/autofill.lua
@@ -52,33 +52,45 @@ local function toggle_main_frame(event)
     else
         frame = Gui.add_left_element(player, { type = 'frame', name = main_frame_name, caption = {'autofill.frame_name'}, direction = 'vertical' })
 
+        local inner = frame.add { type = 'frame', style = 'inside_shallow_frame', direction = 'vertical' }
+        Gui.set_style(inner, { padding = 8 })
+
         local enabled_checkbox =
-            frame.add {
+            inner.add {
             type = 'checkbox',
             name = enabled_checkbox_name,
             caption = {'autofill.enable'},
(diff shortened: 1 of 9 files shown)


grilledham (approved):
Looks good. Thanks for cleaning up the code and updating the style etc.

Out of curiosity you mention in the PR that you "added some input sanitization where easily supported" but I couldn't spot what these changes were?

features/gui/autofill.lua
```diff
@@ -50,37 +50,43 @@ local function toggle_main_frame(event)
...
+        local ammo_count_label = ammo_count_flow.add { type = 'label', caption = { 'autofill.ammo_count' } }
+        Gui.set_style(ammo_count_label, { minimal_width = 140 })
+        local ammo_count_textfield = ammo_count_flow.add {
             type = 'textfield',
             name = ammo_count_name,
-            text = tostring(Autofill.get_ammo_count(player_index))
+            text = tostring(Autofill.get_ammo_count(player_index)),
+            numeric = true,
+            allow_decimal = false,
+            allow_negative = false,
```
RedRafe, line 75: Added native input sanitization for ammo count (the parsing logis was kept still, but this should give the user even more feedbacks as letters/commas cannot be used at all).
I havent checked if RedMew settings could use more of those but I havent heavily refactored that module (yet, dunno if it is needed) so maybe I'll come back to it in the future.
Next will be poll/player_list

features/gui/tasklist.lua
```diff
@@ -136,55 +162,64 @@ local function get_task_updated_by_message()
...
+    element.tooltip = last_edit_message
+end
+
+local function get_task_label_caption(task)
+    local caption = task.name
+    if #caption > label_max_length then
+        caption = caption:sub(1, label_max_length) .. ' [...]'
+    end
+    return table.concat { '#', task.task_id, ' ', caption }
+end
```
RedRafe, line 196: Tasklist are capped at ~100 chars in the Tasks display list to prevent text flooding thee scroll pane.
The full text, editors, time of creation is now displayed for all tasks in their tooltip instead (which is cleaner)

RedRafe (commented):
Left couple comments for ya @grilledham

grilledham:
> Left couple comments for ya @grilledham

Thanks, looks good.

PR #1537, opened 2026-01-31
Update DO:Expanse

RedRafe:
Update DangerOres Expanse map for 2.0 after testing it

### DIFF
--- a/utils/logistic_point.lua
+++ b/utils/logistic_point.lua
@@ -21,11 +21,11 @@ local function parse_item_stack(item_stack)
       value = {
         type = item_stack.value.type or 'item',
         name = item_stack.value.name,
-        quality = item_stack.value.quality,
+        quality = item_stack.value.quality or 'normal',
         comparator = item_stack.value.comparator,
       },
-      min = item_stack.min or item_stack.min,
-      max = item_stack.max or item_stack.max,
+      min = item_stack.min,
+      max = item_stack.max,
       minimum_delivery_count = item_stack.minimum_delivery_count,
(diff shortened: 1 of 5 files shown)


RedRafe:
Ugh... Local git displayed the changes differently than GitHub. Here's the only 2 relevant function changes that were modified effectively (if it can help), everything else is just autoformatter that was hidden in my git changes

Screenshot from 2026-01-31 23-18-46

utils/logistic_point.lua
```diff
@@ -8,11 +8,11 @@ local function parse_item_stack(item_stack)
       value = {
         type = item_stack.type or 'item',
         name = item_stack.name,
-        quality = item_stack.quality,
+        quality = item_stack.quality or 'normal',
         comparator = item_stack.comparator,
       },
       min = item_stack.min or item_stack.count,
-      max = item_stack.max or item_stack.count,
+      max = item_stack.max,
```
grilledham, line 15: Just checking removing `or item_stack.count` was intentional. You didn't remove it below on line 28.

grilledham (approved):
Hide whitespace makes the diff easier to look at.

Image

utils/logistic_point.lua
RedRafe: Thanks, I had to remove it from below too.

Basically, only `min` is a required field, and acts as the request count. `max` is only used to set the upper limit (i.e. have in player logistic 200-400 belts). But, if max is not passed, I think it simply looks "worse" to visualize 200-200 (which is the exact count instead). So, this should make it more visually appealing and correct overall since it will request 200 and NOT set the upper limit exactly as `min`, if `max` is not passed

grilledham (approved):

Issue #1536, opened 2026-01-27
Autodeconstruction callback


RedRafe:
Add async callback for Autodeconstruct to prevent mining from holding last piece of ore when depleting the resource patch (+ check for decon condition & remove it when a new ore tile is added and miner is marked for decon/updated)

PR #1539, opened 2026-02-02
Increase autodeconstruct task timer

RedRafe:
Drills that can output stacked items dont have time to drop the ore inside them before being marked for decon, so increasing time from 5 ticks to 5s should give them enough space to output the last bit of ore


RedRafe:
Fixes #1536

PR #1542, opened 2026-02-05
Update DO:Safety

RedRafe:
# Changes
- Updates DO:Safety to 2.0, added to poll list
- Refactor map to new DO presets format
- DO.map_config module now has enable/disable as well (default: `true`)
- DO.allowed_entities module now only registers the events when calling `register`)

I'm evaluating if `do/modules/allowed_entities` should have some form of alert when calling `.register()` in order to not register the events twice by mistake. Although, that would be true even for each other module that has something like this as well (I believe?), so I'm not sure if needs to be added to be exxxxtra sure.

I know some of the old modules have those type of checks for runtime stage...

```lua
return function(config)
    ...
    Event.add(event_name_1, handler_1)
    Event.add(event_name_2, handler_2)
end


grilledham (approved):
Looks good.

Regarding `allowed_entities.register()`, it probably should have a check to prevent registering twice if we are going to move the event adding code inside `register`. The `merge_dictionary` code implies to me that it's setup such that multiple calls to `register` would be allowed, but I wonder if that is a good idea or if any maps actually make use of that. Either way as I don't think register gets called more than once per map, we don't have to solve this now.

RedRafe:
I used a local var `registered True|False` to track event handler registration, similar to what restart command does. LMK if its good

grilledham (approved):
The `registered` check in allowed_entities looks good to me.

PR #1543, opened 2026-02-08
Update QoL module

RedRafe:
# Changes

Refactored QoL module entirely:
- removed a lot of code repetitions for loaders gui
- improved loaders snapping functionality
- all features are safe to add/remove at runtime now
- added support for redmew data loaders
- fixed getbar/setbar calls

### DIFF
--- a/features/redmew_qol.lua
+++ b/features/redmew_qol.lua
@@ -201,7 +201,7 @@ local function preserve_bot(event)
         return
     end
     local entity = player.selected
-    if not (entity and entity.valid and entity.name == 'construction-robot')then
+    if not (entity and entity.valid and entity.name == 'construction-robot') then
         return
     end
     local logistic_network = entity.logistic_network
@@ -393,7 +393,11 @@ end
 -- Module's public setters/getters
 for _, f in pairs(features) do
     Public['set_' .. f.name] = function(enable)
-        return enable and register_feature(f) or unregister_feature(f)
+        if enable then
+            return register_feature(f)
+        else
+            return unregister_feature(f)
+        end
     end
     Public['get_' .. f.name] = function()
         return enabled[f.name] or false
--- a/features/redmew_qol.lua
+++ b/features/redmew_qol.lua
@@ -1,106 +1,68 @@
--- Assorted quality of life improvements that are restricted in scope. Similar to redmew_commands but event-based rather than command-based.
--- Dependencies
-local Token = require 'utils.token'
 local Event = require 'utils.event'
-local Utils = require 'utils.core'
 local Global = require 'utils.global'
+local Gui = require 'utils.gui'
+local Rank = require 'features.rank_system'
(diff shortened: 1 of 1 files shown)


features/redmew_qol.lua
```diff
@@ -109,644 +71,394 @@ local loaders_technology_map = {
...
-            count = 1
-        end
-    elseif button == defines.mouse_button_type.right then
-        count = 5
-    else
-        return
+-- Module's public setters/getters
+for _, f in pairs(features) do
+    Public['set_' .. f.name] = function(enable)
+        return enable and register_feature(f) or unregister_feature(f)
```
grilledham: You have to be a bit careful about using the `a and b or c` pattern. It's not the same as `a ? b : c` is in other languages. Notably if `a` is `true` and `b` is `false` they do different things. E.g.
`print(true and false or 'cake')`
and
`print(false and false or 'cake')`
both print cake. Whereas, if it existed in lua, `print(true ? false : 'cake')` would print false.

I didn't flag this in previous code reviews because the expression you previously used for `b` couldn't be falsely, but in this case it can be.

Personally I find in cases like this, where I want a ternary operator, it's just simpler to use good old if/else.

grilledham (commented):
Looks like a decent amount of clean up. In all honestly I haven't gone through it all, but I assume you have tested at least the commonly used features: loaders, inserter_drops_pickup, restrict_chest, preserve_bot.

features/redmew_qol.lua
RedRafe: Fair, thanks for pointing that out. Chenged to your suggestion so it should be easier to read too

RedRafe:
> Looks like a decent amount of clean up. In all honestly I haven't gone through it all, but I assume you have tested at least the commonly used features: loaders, inserter_drops_pickup, restrict_chest, preserve_bot.

Yes, I did test them (except the preserve bot, which I only formatted a bit but left unchanged), thats how I found out that the chest limiter module was actually broken in 2.0.

I mostly wanted to refactor this file to update the loaders module since I'm adding data support for those, and code for them was... improvable.

Loaders have new Gui, all handlers are added only if needed (not extra overhead) and Gui code is simpler to maintain (just need to update entity list rather than create new ones for each tier). Game-wise, loaders now will snap to a bunch more entity types + all their ghosts variants which should improve usability.

Altho, this becomes secondary sometimes as the data patch from RedMew-Data mod introduces 1x1 redmew loaders fully supported in-game (hence why the loaders Gui is now only drawn if no `redmew-loader` is resent), to offer a bunch of upgrades to this hack:

1. 1x1 loaders will automatically snap to belts in addition to containers
2. 1x1 loaders are not hidden, so they will be visible for upgrade planners and use regular techs/recipes
3. Better graphics(!) :)

grilledham (approved):
Cool, all looks good.

PR #1547, opened 2026-02-18
Add DO/Joker, Fox, and 3-way railgrid maps

RedRafe:
## Changes:
- Renamed `railgrid` map into `rail_grid_4_ways`
- Updated railgrid maps allowed entities to 2.0
- Updated railgrid alignments to 2.0
- Added `rail_grid_3_ways` map
- Added `fox` map from picture
- Added `crash-site` and `remnants` entity list to resources
- Added `DO/Joker` preset to DangerOre maps

## DangerOres changelog:
- [DO] Turn OFF always day mode
- [DO] Removed Memory unit mod (web)
- [DO] Added `enemy_starting_radius`, `tile_builder`, `spawn_builder` settings to map config
- [DO] Added `PictureBuilder` module
- [DO] Added `Nightfall` module (biters attack in the dark, OFF)
- [DO] Added `Radioactivity` module (walking on ore poison the player, ON)
- [DO] Fixed `Water` module not aligning deepwater to water threshold automatically

## RedMew Data changelog:
- Added RedMew 1x1 loaders
- Added reskins for advanced factory
- Added technology milestones
- Added vibrant colors to entities with player color masks
- [DO] Added Deep-house chest
- [DO] Rebalanced solar energy production and rocket recipes
- [DO] Rebalanced increased copper consumption early game
- [DO] Removed ore stack changes
- [DO] Removed robots cost multipliers
- [DO:JK] Added biters drop resources

## Summary
This PR looks like a lot of changes but it is most relevant to the addition of DO:Joker map.
DO:Joker features a dense fox picture as main starting area and a lean 3-way railgrid outside, hence, `rail_grid_3_ways`, `rail_grid_4_ways`, and `Fox` maps have been added/updated in the process.

Overall DO changelog is split between mod & scenario to rebalance a few mechanics and game progression
- [X] **scenario**: no relevant changes are made on scenario part other than allowing a few settings of map builders to have default values/optional values in order to provide more flexibility when creating new maps.
- [X] **mod**: major code changes involve reskinning all "vanilla-like" ptototypes with the addition of a new reskin lib & assets, code cleanup for shared prototypes (entity/recipe/technology) across different maps, code cleanup for scenario packs info, settings simplification, addition of  aforementioned features directly supported by the engine.

---

For all the mod changes, please refer to the git history of the public repo: [github.com/RedRafe/redmew-data/commits](https://github.com/RedRafe/redmew-data/commits/main/):
- Add biter loot
- Add vibrant colors
- Add ore assets
- Fix steam engine graphics
- Add inserter stacking behavior
- Add milestones
- Add recipe rebalance
- Add deep house
- Add loaders
- Add entities, recipes and technologies to ptototypes (+ assets)

---

Here's a preview of the different maps added, DO:Joker has been playtested after 1st draft and 2nd draft is queued for 2nd round of fine tuning regarding ore density balance and progression.

> DO:Joker map
Screenshot from 2026-02-12 12-48-24

> Fox map
Screenshot from 2026-02-18 10-37-23

> 4-ways rail grid
Screenshot from 2026-02-18 10-36-29

> 3 ways rail grid
Screenshot from 2026-02-18 10-35-51


map_gen/data/presets/fox.lua
```diff
@@ -0,0 +1,1061 @@
+return {
```
grilledham, line 1: Suggestion: Adding these images/maps to 
https://github.com/Refactorio/RedMew/tree/develop/map_gen/data/.source_images
and
https://github.com/Refactorio/RedMew/tree/develop/map_gen/data/.map_previews
Would be good. It's nice to have the reference images kept somewhere in case we want to regenerate the data.

map_gen/shared/builders.lua
```diff
@@ -329,8 +329,31 @@ local tile_map = {
     [30] = 'sand-3',
     [31] = 'stone-path',
     [32] = 'water-green',
-    [33] = 'water'
+    [33] = 'water',
+    --[[
```
grilledham: Did you use https://github.com/grilledham/FactorioMapConverter to create the image data? If so I'd be happy for you to add these tile IDs.

resources/crash-site.lua
```diff
@@ -0,0 +1,15 @@
+return {
```
grilledham, line 1: What is `resources/crash-site.lua` and `resources/remnants.lua` used for?

map_gen/maps/danger_ores/configuration.lua
```diff
@@ -84,7 +87,7 @@ return {
     draw_resource_selection = false,
     technology_price_multiplier = 25,
     manual_mining_speed_modifier = 1,
-    always_day = true,
+    always_day = false,
```
grilledham, line 90: Does this change all the danger ore maps to have nighttime?

map_gen/maps/danger_ores/modules/picture_builder.lua
```diff
@@ -0,0 +1,127 @@
...
+
+        if entity.amount and entity.amount < 1 then
+            entity.amount = 1
+        end
+
+        return entity
+    end
+end
+
+local function picture(pic)
```
grilledham: I'd be happy for this variation of `b.picture` to be added to the builders module. Or the existing one to be changed to accommodate these extra features.

map_gen/maps/danger_ores/modules/nightfall.lua
```diff
@@ -0,0 +1,262 @@
+-- Biters attack in the dark
```
grilledham, line 1: I don't see this module `map_gen/maps/danger_ores/modules/nightfall.lua` being used anywhere?

grilledham (commented):
A quick skim through and some thoughts.

map_gen/data/presets/fox.lua
RedRafe: Yes I can add the images to the src. 512/1024 is good for for file size? Seems to be around the definition of other images too, but the 4 suits of poker can be scaled down to 64px since they're pretty basic if we want to reduce file size

grilledham: Whatever size the images were that you used to generate the data. I'm not too worried about the size. It's only for the repo, the scenarios don't have the images.

map_gen/shared/builders.lua
RedRafe: I knew that was the one used to generate the other img data, but since I'm on Linux, I didn't recompile the c# project and instead re-implemented a similar script with python to do basically the same I guess (except I get to pick the tile-color pairs, idk if the original did that based on prevalent tile color). I was wondering why the tile list didn't include all tiles, of if there was a reason behind it (like excluding colored concrete is maybe fine since they're not usually available to player, even if they provide bright color for images, as I used the green/orange for the Fox's eye and fur)

resources/crash-site.lua
RedRafe: They're a list of prpptotypes/chests that can be useful to have around as per-computed data. I use them to scatter debriefs around the map, at RNG positions, to give a bit of scenery (see'`artifacts` function at do/joker)

map_gen/maps/danger_ores/configuration.lua
RedRafe: Yes. This turns off always day mode for all DO maps

map_gen/maps/danger_ores/modules/picture_builder.lua
RedRafe: I can add this variation to builders (dunno if extending/replacing original is needed since no other maps other than DO would need a Pic filler included in the process)

map_gen/maps/danger_ores/modules/nightfall.lua
RedRafe: Correct, although I playtested this feature locally, it didn't make the cut with latest DO additions. Since the DO maps are default on peaceful, when biters are spawn, they do not attack the player or it's structures by default, unless triggered.
So, even if the feature does it's job correctly, it does not play well with peaceful mode. And it came down between turning peaceful mode off, or postponing this feature to a new DO map where maybe peaceful mode can be turned off just for that one.

map_gen/shared/builders.lua
grilledham: > I knew that was the one used to generate the other img data, but since I'm on Linux, I didn't recompile the c# project and instead re-implemented a similar script with python

I should convert it to [Avalonia ](https://github.com/AvaloniaUI) then it would be cross platform. Or we add it to the web interface.

> I was wondering why the tile list didn't include all tiles, of if there was a reason behind it

I think it's because the map converter code was written before some of these tiles existed

map_gen/maps/danger_ores/configuration.lua
grilledham: I'm ok with testing this and getting feedback from users. But my personal take is that always daytime makes senses for danger ore. It makes solar stronger which is weak anyway because of the space issues and burning more coal is good. Also nice to not have to deal with night time when building.

resources/crash-site.lua
grilledham: Ok, I guess I didn't search very well as I didn't see remnants being used.

map_gen/maps/danger_ores/modules/nightfall.lua
grilledham: Ok, happy to keep the file. Just checking that you hadn't forgot to use it and thought you were.

map_gen/maps/danger_ores/modules/picture_builder.lua
grilledham: I'll leave it up to your judgment.

map_gen/shared/builders.lua
RedRafe: Ah, makes sense. Up to you. I have an equivalent script to do it too. I can extend the tiles (I didn't officially because I guess if I were to rearrange all tiles alphabetically it would mess up all other image data, so I appended the extra ones left out alphabetically below, but they're not used anywhere)

The builders change tile can be used to expand the old tile list to new tiles if needed

resources/crash-site.lua
RedRafe: Np. Remnants are uded in DO/joker, crash site not (because CS chests cannot bee marked for decon and I though it would be too tedious to leave wrecks scattered around that can only be handmined, so a custom list of 1x1 wrecks is used to compensate instead in the scenario code).

But since I already made the list, didn't felt too out of place to include it in resources if needed for other scenarios to add/remove stuff (since base game already uses that list for the CS starter area)

map_gen/maps/danger_ores/configuration.lua
RedRafe: The rationale behind this change, other than testing out new stuff, is that even with always mode, no scenario ever uses solar power.

The 2 main reasons are:
- coal is plenty on all maps to power the base and prevent brownouts. Memlry storage mod used to drain power to store infinite amount of resources. Even when that became unsustainable, base simply browns out or disconnects, but solar in never built to act as emergency power.
- game doesn't last long enough to go to the "endgame" where solar is better for UPS, so for late game, mk2 entities or nuke (or both) are usually used.

If we were to incentivize solar, we would need to have always day + a solar multiplier for surface (which is LuaSurface::solar_power_multiplier) and set it to 2x, 4x, 8x... This would have 0 performance impact honestly.

But, nobody still uses solar because it takes up too much space and it's tedious to setup. I think if we had a power multiplier it would make solar Op as it's a lot of power for very little ore input, and players would still not use tto but more coal and invest more resourves/space elsewhere.

The reason why I turned it back on is because I rebalanved solar from mod part: solar & accus are 2x powerful and require about 2x resources to craft. Satellites require 50 each instead of 100 each to compensate for the double-density. This is still in testing so we'll found out in the next weeks, but a solar/accus array now occupies about double the space off a coal plant. Which is kinda fair for the resource needed and basically endless power.

Additionally, the map DO:Collapses uses a mechanic where every some Million ore mined, all ores on map double time to mine.
This mechanic results in coal becoming more expensive to mine than what it yields in fuel power, so solar becomes mandatory. The 2x density of new solar seems to fit a bit better within the scenario, which was lost the first time it got played instead

RedRafe: On the other hand, always day makes it easier to see at nighttime. This is the only drawback I too feel bad because squinting eyes at the monitor is not always cool, but fortunately there are lamps so a bit more use of scenery entities in the base is now required for QoLs

RedRafe:
I've added the 2.0 tiles at the back of the default tiles list and extended the decompress/picture builders function to allow entities within them.
Maybe this will make the shape function a bit slower (as it will make 2 extra checks) but, chunk gen is still capped by the generate module so I dont expect users to notice any performance issue.

I've tested the changes with `maps/fox` (regular pic-> tile shape gen) and `maps/do:joker` (pic -> tile + entity gen) and everything seems right.

`b.decompress` will not assign any default tile map, `b.picture` will assign a default/fallback table for both tiles and entities (empty table) so it should be safe enough as decompress will usually cascade into picture function anyway. LMK if this works or if I should make additional changes

grilledham (approved):
Looks great

PR #1548, opened 2026-02-19
Diggy: add blueprint tools

RedRafe:
This PR adds Blueprint tools for Diggy scenario.

## Context & feature explanation
Support beams (tiles and special entities) cannot be marked for deconstruction by default (antigrief/safety feature)
Regulars and above will have access to this new feature to turn On/Off the safety measure however they prefer.
The feature will automatically popup whenever the player is about to perform a "blueprint action", which is:
- using a blueprint record
- using a blueprint book
- setting up a blueprint from map
- holding a blueprintable item-entity in hand

Players can:
- turn On/Off the safety measure all together
- turn On/Off tiles in bulk
- Show/Hide the popup window (from this window or RedMew "Settings > Show blueprint tools")
- turn On/Off individual entities

Here's a quick demo that goes through the various settings/cases:

Demo video

# Changes:
- moved support beams definition from presets to template
- moved `on_marked_for_deconstruction` handler from `diggy_cave_collapse` to BP Tool module
- BP Tool is now responsible for the antigrief feature all together

Arguably, `diggy_cave_collapse` module could force require this module or I could have included this logic in `diggy_cave_collapse` to begin with, but I thought if would have been cleaner to detach the antigrief/tool logic on its own module as it handles a lot of Gui events rather than game actions.

### DIFF
--- a/map_gen/maps/diggy/feature/blueprint_tools.lua
+++ b/map_gen/maps/diggy/feature/blueprint_tools.lua
@@ -277,7 +277,7 @@ local function get_admin_main_frame(_, player)
         Gui.set_style(online.parent, { horizontally_stretchable = true })
 
         local grid = inner
-            .add { type = 'scroll-pane' }
+            .add { type = 'scroll-pane', style = 'naked_scroll_pane' }
             .add { type = 'table', style = 'table_with_selection', column_count = 4 }
 
         local function make_rank(name)
@@ -305,6 +305,7 @@ local function get_admin_main_frame(_, player)
             local list = grid
                 .add { type = 'scroll-pane', style = 'naked_scroll_pane' }
                 .add { type = 'table', column_count = 10, style = 'filter_slot_table' }
+            list.parent.horizontal_scroll_policy = 'never'
 
             for name, status in pairs(ps.entities) do
                 list.add{ type = 'flow' }.add{
--- a/map_gen/maps/diggy/feature/blueprint_tools.lua
+++ b/map_gen/maps/diggy/feature/blueprint_tools.lua
@@ -16,6 +16,7 @@
     - turn On/Off individual entities
 ]]
 
+local Command = require 'utils.command'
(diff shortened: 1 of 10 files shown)


map_gen/maps/diggy/feature/blueprint_tools.lua
```diff
@@ -0,0 +1,363 @@
+--[[
+    Blueprint tools for Diggy
+
+    Support beams (tiles and special entities) cannot be marked for deconstruction by default (antigrief/safety feature).
+    Regulars and above will have access to this tool to turn On/Off the safety measure however they prefer.
```
grilledham, line 5: Just thinking about how this works. Is there an edge case where a players is a regular then enable deconstruction of walls, then get demoted to guest or probation and then can't change the settings?

RedRafe: It's a valid edge case.

I did account for the UI that it will automatically show or not based on rank, however, I didn't account for it in marked for decon event.

I guess when a player is demoted I could react to the event and reset to default all the values. Or, option B, on marked for decon, skip the controls if player rank is not met and cancel decon.

I would prefer option b among the 2 as it seems more resilient

grilledham: I prefer option b as well.

RedRafe:
I solved the edge case of demotion with player still being able to still decon stuff, fixed it and tested it again.

I also implemented a quick admin GUI `/blueprint-tools` that lists all the players' settings and can act as a master toggle for them (per-entity mask is included as well, On/Off switch should be enough to quiet any doubt) .

This command is available to Moderators+ as per new rank system

Screenshot from 2026-02-19 23-57-40

> From MP game (obviously everything is On since nobody hasnt played with it yet)
Screenshot from 2026-02-20 10-05-03

grilledham (approved):
Looks good.

PR #1550, opened 2026-02-20
Address exploit of container dump module

RedRafe:
### Issue: 

Container dump module of Danger Ores can lead to endgame exploit to generate infinite resources from nothing by:
- killing a full chest of copper cables (i.e. running over it with tank)
- (module turns into coal)
- coal is mined with Prod3s
- coal is exchanged with copper-ore at 2:1 ratio
- copper is smelted + crafted into cables with double prod3s bonus
- fill chest back and repeat

Screenshot from 2026-02-20 11-55-12

### Solution:

1. Destroying a chest to clear space and void items now also kills the player, making the automation of infinite loop resources very tedious to automate as chests would need to be killed 1 by 1 with respawn cooldowns in between
2. Market offers decreased from 2:1 to 5:1 ratio

### DIFF
--- a/map_gen/maps/danger_ores/modules/container_dump.lua
+++ b/map_gen/maps/danger_ores/modules/container_dump.lua
@@ -1,11 +1,14 @@
 local Event = require 'utils.event'
 local Global = require 'utils.global'
 local math = require 'utils.math'
+local Token = require 'utils.token'
+local Task = require 'utils.task'
 
 local floor = math.floor
 local max = math.max
 
 local chest = defines.inventory.chest
+local stun_sticker_duration = 3 -- seconds
 
 local died_entities = {}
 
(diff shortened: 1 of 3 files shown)


map_gen/maps/danger_ores/modules/container_dump.lua
```diff
@@ -76,6 +101,25 @@ Public.register = function(config)
...
+                return
+            end
+
+            local character = actor.character
+            if not (character and character.valid) then
+                return
+            end
+
+            actor.print('The ore fights back!', { color = { 255, 128, 0 } })
+            character.die(game.forces.neutral)
```
grilledham: I wonder if the dying part should be configurable. This seems more like a diggy danger ore issue than a regular danger ore issue. Part of me feels that killing the player is quite harsh, but the only alternative I can think of is spilling the items instead. Though I suppose it's odd for players to accidentally destroy containers, outside of using explosives to kill biters and the container getting caught in the crossfire.

RedRafe: Experience of the last X months is: in regular DO almost no-one ever kills a container (mainly because ppl know it would spawn "useless" coal to mine, and also because there's no real need to get rid of stuff to make room). So feature is pretty much irrelevant.

For Diggy, the infinite resource loop is a more relevant concern. We had in DO the powered-warehouses, and they would start spilling content on ground when left unpowered. That made everyone angry at the mechanic cuz it punished the whole party rather than the single griefer/offender.

IMO, neither killing nor spawning ore on a 1x1 or NxN area is an harsh punishment, as respawning is a matter of 10s and mining out the ore tiles is more an annoyance than a real issue.

It only acts as a "warning" after 1st attempt, since it wastes a bit of player time and cannot be automated, players wont do it again ever after.



I dont think this feature is triggered enough to cause any issues among regular players. Maybe griefers would be the only ones experiencing a few deaths from time to time, but it's still unusual to get kills in DO or to see players blow up chest at all. A game can easily finish with 0 deaths total (no trains, peaceful,...).

If so, we can still revert the change again and turn it down a notch.

grilledham: > Experience of the last X months is: in regular DO almost no-one ever kills a container (mainly because ppl know it would spawn "useless" coal to mine, and also because there's no real need to get rid of stuff to make room). So feature is pretty much irrelevant.

As in the feature works in it's current form and doesn't need changing for DO?

> IMO, neither killing nor spawning ore on a 1x1 or NxN area is an harsh punishment, as respawning is a matter of 10s and mining out the ore tiles is more an annoyance than a real issue.

For large maps it can be more than a 10s inconvenience as it's also the time to get back to your corpse. Plus it would interrupt what you are doing if you are in the middle of building something while clearing chests. Maybe I am over reacting about killing the player. But it sets the tone for the map, particularly for new players. I could imagine a new player to DO would naturally try to destroy containers to take advantage of the DO game mechanics.  Staying alive makes it easier to see/notice the spawned ore. I'd be more accepting of killing players in diggy as that is inherently a more dangerous map (DO normally has bitters to peaceful).

Though thinking about this again, there are some other solutions we could explore.
1. Don't use the container module for diggy DO. The mechanic of the map are different for diggy around clearing space and resource scarcity. So maybe this feature just isn't needed.
2. For diggy DO lower the amount of coal spawned to make the exploit not viable. Again as the mechanics for diggy are different a fraction of the ore spawn is still inconvenient without being able to exploit freeing space as that happens mostly through mining rocks.
3. Change the market ratio to make the expolit not viable.
4. Combination of 2 and 3.

RedRafe: For once, I find myself in disagreement with the concerns raised.

I'll try to explain the issue so a better solution can be found:

1. Fo regular DO, this feature in any of it's implementations has little to no impact as no chests are ever killed in a DO run. Neither to make space, nor to exchange resources to coal, nor by accident.

2. For DiggyDO, Diggy used a separate DO logic, but for the sake of reusing and maintaining, it was switched to use DO's files & implementation around when both scenarios got heavily refractored at the end of 1.1 and 2.0 porting.

The issue with Diggy is that long lasting maps generate absurd resource imbalance and players started to blow up chests of whatever was overflowing to keep the base running. To address this mechanic, 3 features were added back in few versions ago:
- disabling mining prod (from DO)
- container dump (from DO)
- market chest (new diggy feature)

The combination of those 3 was meant to incentivize players into using all resources as best as possible; and then if really needed, trade excess at bad ratios instead of voiding it in chests (instead of storing 420M ores in bot network).

Players later realized that without mining prod, diggy needed a lot of expansion and manual work to feed the factory. So the exploit to kill chest of cables, mine coal, exchange for copper &brevraft in cable loop was found to generate infinite stream of resources.

For latest Diggy version, exchange rates have been raised from 2:1 to 5:1 to offset the prod3s chain already - however - since constantly expanding manually was deemed too tedious for large SPM goals - mining prod has been introduced back. Now no value of exchange rate can offset the productivity gained from all the steps involved. So I nerfed the infinite loop by basically debouncing the loop by killing poster when it destroys a chest.

Lore-wise, I personally found it very fitting with the danger of ores theme, and penalty was not received as too harsh neither in discussion with players nor in tests with public server.

However, if a better solution can be found, I'm open to it. I haven't quite found a fix to address all these issues so far

grilledham: 1. Are you saying the container dump module isn't needed for regular DO, or that this change to the module will have no impact on DO runs?

Either way I think we will just have to agree to disagree. This doesn't feel like the right way to solve the problem to me. But I'm open to trying it and seeing what players say.

grilledham (approved):

map_gen/maps/danger_ores/modules/container_dump.lua
RedRafe: > 1. Are you saying the container dump module isn't needed for regular DO, or that this change to the module will have no impact on DO runs?

Both - DO doesnt *need* container dump as antigrief/anticheat system as players already behave for that matter, but I find it very fitting to the lore/design of the map so I wouldn't necessarily remove it even if it's almost unused.
Given that it is not a relevant feature, changes to it (like harsh penalty of character death) would not have relevant impact as feature is not triggered enough for players to notice.

Even tho we might be in disagreement I dont want to push changes that are not shared. I dont need this to go through as is just for the sake of it.
A new Diggy run has already restarted and it is testing these changes + some some other little tweaks, so I'm in no rush :)

If we think death is too harsh a penalty, we can maybe just stun the player. Looking at base protos, there's landmines, discharge defense and (with SA) tesla turrets that have this "stun mechanic":

```lua
action = {
    type = "create-sticker",
    sticker = "stun-sticker"
}
```

(A sticker is some kind of entity - enemy green spits are "sticker" that get attached to player and fade with time just to give context).
Maybe I can fiddle around with that and see if I can attach stun stickers around target actors (will have to check how it works with vehicles but spit do attach to tanks/cars so it should be possible). This would be sufficient I believe as debouncing mechanism to prevent players from running over a row of chests and convert it to coal for endless mining.

Bonus: if it's not enough to stop it, then I can additionally kill just the vehicle and maybe also spawn some enemies as well as coal so the feature is less predictable from player side

grilledham: > Both - DO doesnt need container dump as antigrief/anticheat system as players already behave for that matter

I don't think this is correct. I added the module originally because we did have a problem with this.

> Even tho we might be in disagreement I dont want to push changes that are not shared.

I appreciate that 😊

> If we think death is too harsh a penalty, we can maybe just stun the player.

I would prefer stunning the player over killing them, so if you can make that work, that works for me.

RedRafe: > I added the module originally because we did have a problem with this.

Sure! And I think nowadays the combination of regulars knowing there's a penalty to killing chests of ore and them understanding game economy a Lil better has lead to no chests being blown up anymore.

DO is very chill so the only "precious" resource is time. Extracting resources takes time an nobody wants to mine twice a tile because someone blew a chest. Even if not triggered anymore, I think if the feature was disabled ppl would start blowing up chest again

RedRafe:
I've managed to replace the kill with stun.
Tho it was not possible to stun a vehicle (only slowdown stickers), so I've resumed to kill the vehicle instead (the driver is automatically ejected from it but not killed, just like when you crash your car into a rock).

Module is configurable with params:
```lua
---@field entity_name? string, resource to be spawned (default: coal)
---@field time_penalty? number, time lost by the player when misbehaving (default: 18s, 0 to apply none)
---@field spare_vehicle? boolean, saves the involved vehicle, if any (default: false)
```

grilledham:
> I've managed to replace the kill with stun. Tho it was not possible to stun a vehicle (only slowdown stickers), so I've resumed to kill the vehicle instead (the driver is automatically ejected from it but not killed, just like when you crash your car into a rock).
> 
> Module is configurable with params:
> 
> ```lua
> ---@field entity_name? string, resource to be spawned (default: coal)
> ---@field time_penalty? number, time lost by the player when misbehaving (default: 18s, 0 to apply none)
> ---@field spare_vehicle? boolean, saves the involved vehicle, if any (default: false)
> ```

I'm happy with this.

grilledham (approved):

PR #1552, opened 2026-02-23
Update CrashSite outposts

RedRafe:
# Changes

### A. Add outposts maximum level

`outpost_data` for outpost builder will now also accept 2 optional parameters:
- `maximum_level_count`: number
- `maximum_level_formula`: token id to function callback

Both will try to set `maximum_level` and if set/reached, it wont be possible to upgrade the outpost further.
I have decided to simply disable the upgrade item instead of removing it completely as this option keeps open the possibility to develop further the feature in the future, without adding/removing market item every time.

Screenshot from 2026-02-23 12-14-23

> Following Tigress and CrashSite veteran players feedbacks, Spawn Market in CS has been capped at Lvl.6 for multiplayer events

### C. Slight GUI style update to `Retailer`

### DIFF
--- a/map_gen/maps/crash_site/scenario.lua
+++ b/map_gen/maps/crash_site/scenario.lua
@@ -802,6 +802,10 @@ local function init(config)
 
     map = b.apply_entity(map, ore_grid)
 
+    local spawn_maximum_level_formula = Token.register(function()
+        return game.is_multiplayer() and 6 or nil
+    end)
+
     local market = {
         callback = spawn_callback,
         data = {
@@ -809,6 +813,8 @@ local function init(config)
             upgrade_rate = 0.5,
             upgrade_base_cost = 500,
             upgrade_cost_base = 2,
+            maximum_level_count = nil,
+            maximum_level_formula = spawn_maximum_level_formula,
             {
                 price = 0,
                 stack_limit = 1,
@@ -862,7 +868,6 @@ local function init(config)
                 name_label = {'command_description.crash_site_barrage_radius_name_label', 1},
                 sprite = 'virtual-signal/signal-B',
                 description = {'command_description.crash_site_barrage_radius', 1, 0, 25}
-
             },
             {
                 price = 1000,
--- a/features/retailer.lua
+++ b/features/retailer.lua
@@ -382,11 +382,9 @@ local function draw_market_frame(player, group_name)
         Gui.set_data(button, frame)
     end
 
-    local inner = frame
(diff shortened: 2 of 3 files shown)


grilledham (approved):
Looks good.

map_gen/maps/crash_site/scenario.lua
```diff
@@ -802,13 +802,19 @@ local function init(config)
...
+
     local market = {
         callback = spawn_callback,
         data = {
             market_name = 'Spawn',
             upgrade_rate = 0.5,
             upgrade_base_cost = 500,
             upgrade_cost_base = 2,
+            maximum_level_count = nil,
+            maximum_level_formula = spawn_maximum_level_formula,
```
grilledham, line 817: Possibly a bit odd that you could set both `maximum_level_count` and `maximum_level_formula`. For cases like this I favour having a single field but allowed it to be a number or token. Which is a bit unfortunate in this case as a token is also a number. We could use a table with a single value field for the token case. Or a table with a  type field.

This is all just a suggestion.

RedRafe: I took inspiration from technologies prototypes, tech can either specify a count or formula for its cost. 

This field is not directly used, but rather, it is only read-used when the builder parses it the first time, and then saves the result of either in ` maximum_level`. That is the key read at true runtime.

We could also add a string field for the type, technology also has flag for infinite true/false iirc

grilledham: I was thinking about a case where someone configures this wrong by setting both fields then gets an unexpected result. If there is only one field such errors are harder to make. But this is minor.

RedRafe: Okay, just checking we were talking about the same issue.

True, count is not used but I thought for regular outposts, having a simple max would have been simpler instead of setting a callback for it, so I went ahead and implemented the double behavior

There's no official docs on what fields are necessary for outposts data, but I could improve comments on that and add an assert/alert in case someone tries to set both (like we have in some other places of code where an error is thrown if some calls are misplaced or repeated)

grilledham: An assert would be another way to protect against bugs.

RedRafe: Aight, I've also added some Lua annotations on top for a couple data used in the outpost builder so it should be easier to understand if someone needs to refactor things around in the future

PR #1554, opened 2026-03-01
Add inventory command

RedRafe:
Add a Moderator+ command to show all/target player inventory content with `/inventory <player>`

Screenshot from 2026-03-01 10-04-02

### DIFF
--- a/features/admin_commands.lua
+++ b/features/admin_commands.lua
@@ -4,6 +4,7 @@ local DiscordChannels = require 'resources.discord'.channel_names
 local Event = require 'utils.event'
 local Game = require 'utils.game'
 local Global = require 'utils.global'
+local PlayerInventory = require 'features.gui.player_inventory'
 local Rank = require 'features.rank_system'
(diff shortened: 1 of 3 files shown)


grilledham (approved):
Looks good.

Based on what you said in the mod-general channel.
>  I think this is especially useful for Admins/Mods to quickly check general players contents to scout for any griefer and then proceed with /open <player>

Would it be possible to add an open button to this ui that open's the player's inventory?

RedRafe:
> Would it be possible to add an open button to this ui that open's the player's inventory?

I checked this earlier with API, I couldnt find any way to override built-in commands or allow non-admins to run them if they're admin-only. I can only react to them and possibly do something in addition, however, which is as good as drawing this Gui.

For admins, this is mainly useful to get an overview of all invs then investigate with open.
For mods, this is also good for general overview and acts as a lite `/open player` since it can display contents but not interact with it

grilledham:
Ok, not as simple as I thought it'd be.

PR #1557, opened 2026-03-09
Update error handler

RedRafe:
Removes the double log and replaces it with full stack trace of the error

### DIFF
--- a/utils/error_logging.lua
+++ b/utils/error_logging.lua
@@ -10,7 +10,8 @@ local floor = math.floor
 local format = string.format
 local insert = table.insert
 local concat = table.concat
-local pcall = pcall
+local xpcall = xpcall
+local trace = debug.traceback
 
 -- Local constants
 local minutes_to_ticks = 60 * 60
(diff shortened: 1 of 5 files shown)


utils/error_logging.lua
```diff
@@ -85,10 +86,14 @@ end
...
-    local success, err = pcall(try_generate_report, str)
+    local success, result = xpcall(try_generate_report, Public.error_handler, str)
     if not success then
-        log(err)
+        log(result)
     end
 end
 
+function Public.error_handler(err)
+    return trace(err)
```
grilledham, line 96: What does trace do in factorio? Does it go into the logs? If so do we need both trace and log(result) in event_core and others?

RedRafe: [Lua 5.2 debug::traceback](https://www.lua.org/manual/5.2/manual.html#pdf-debug.traceback) it only returns a string with a traceback of the call stack, which is then logged by the log function

grilledham: I see that makes sense.

grilledham (approved):

PR #1558, opened 2026-04-13
hail hydra bugfix

actinoide:
fixed empty unit_group creation, leading to lag on large maps.

### DIFF
--- a/map_gen/shared/hail_hydra.lua
+++ b/map_gen/shared/hail_hydra.lua
@@ -151,6 +151,9 @@ local on_died =
             if group and group.valid and #group.members > 0 then
                 group.set_command(command)
             end
+            if group and #group.members == 0 then
+                group.destroy()
+            end
         end
     end
 )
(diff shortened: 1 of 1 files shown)


map_gen/shared/hail_hydra.lua
```diff
@@ -151,6 +151,9 @@ local on_died =
             if group and group.valid and #group.members > 0 then
                 group.set_command(command)
             end
+            if group and #group.members == 0 then
```
grilledham:
```suggestion
            if group and group.valid and #group.members == 0 then
```
If the above check is doing this, then we might need to do that here also. Or both the valid checks are not needed. Probably safer to just always call valid.

grilledham (commented):
Thanks for the fix. I have one suggestion though. Do you want to make the change?

map_gen/shared/hail_hydra.lua
actinoide: I thought it wouldnt be necesary to do since the group is being marked for deletion, probably doesnt hurt though.

grilledham: Thanks, I'll merge this.

PR #1559, opened 2026-04-15
fixed landfill deletion

actinoide:
fixed landfill deletion upon mining of covering tile.

(sorry for the repeated PRs with tiny fixes, let me know if there is a more elegant way to do this in case i find any more)

### DIFF
--- a/map_gen/maps/diggy/feature/diggy_hole.lua
+++ b/map_gen/maps/diggy/feature/diggy_hole.lua
@@ -134,7 +134,7 @@ local function on_mined_tile(surface, tiles)
     for _, tile in pairs(tiles) do
         if (artificial_tiles[tile.old_tile.name]) then
             count = count + 1
-            new_tiles[count] = {name = 'dirt-' .. random(1, 7), position = tile.position}
+            new_tiles[count] = { tile.old_tile.hidden_tile, position = tile.position }
         end
     end
 
--- a/map_gen/maps/diggy/feature/diggy_hole.lua
+++ b/map_gen/maps/diggy/feature/diggy_hole.lua
@@ -134,7 +134,7 @@ local function on_mined_tile(surface, tiles)
     for _, tile in pairs(tiles) do
         if (artificial_tiles[tile.old_tile.name]) then
             count = count + 1
-            new_tiles[count] = { tile.old_tile.hidden_tile, position = tile.position }
+            new_tiles[count] = { name = tile.old_tile.hidden_tile or ('dirt-' .. random(1, 7)), position = tile.position }
         end
     end
 


map_gen/maps/diggy/feature/diggy_hole.lua
```diff
@@ -134,7 +134,7 @@ local function on_mined_tile(surface, tiles)
     for _, tile in pairs(tiles) do
         if (artificial_tiles[tile.old_tile.name]) then
             count = count + 1
-            new_tiles[count] = {name = 'dirt-' .. random(1, 7), position = tile.position}
+            new_tiles[count] = { tile.old_tile.hidden_tile, position = tile.position }
```
grilledham:
```suggestion
            new_tiles[count] = { name = tile.old_tile.hidden_tile or ('dirt-' .. random(1, 7)), position = tile.position }
```
1. You seemed to have dropped the `name` key?
2. I wonder if diggy does this because the `artificial_tiles` it spawns don't have hidden tiles set? But I haven't checked to confirm that. It's just otherwise this code is a bit odd as is this not the default factorio behaviour for mining tiles?

grilledham (commented):
Thanks for the PR, I have a comment though. Also have you tested this?

> (sorry for the repeated PRs with tiny fixes, let me know if there is a more elegant way to do this in case i find any more)

I think what you are doing is correct. Generally if the issues are unrelated it's favourable to have separate PRs, even if they are small. I personally don't mind small or large PRs, but in my experience some people tend to be more against larger PRs than smaller ones.

map_gen/maps/diggy/feature/diggy_hole.lua
actinoide: i have done some very basic testing (placing landfill on water followed by bricks and then mining said bricks and gotten the desired behaviour (the landfill being uncovered instead of replaced), i also placed bricks on normal dirt and removed it to check that normal functionality wasnt impacted, the two versions are different in that the old one always randomises the underlying dirt upon removal of the bricks, while my version places the same kind of dirt, i believe that these are equivalent enough for practical considerations though.).

1. yes, i did forget the name key, though as mentioned the code seems to work fine without it.
2. yes im trying to imitate default behviour, i dont feel confident removing the function though, due to the call at the end, which to me looks like it might trigger the updates to cave stability (which is impacted by placed tiles) and therefore that functionality might be impacted if the function was entirely removes).

Im not very familiar with Lua, im assuming the string will only evaluate to false if it is empty ? if so your failsafe seems sensible, even if i dont know where it would come into play.

grilledham: 1. We should probably add the name key back, the docs imply it's needed https://lua-api.factorio.com/latest/concepts/Tile.html
2. As I said I wonder if this is needed because diggy might spawn a `artificial_tiles` without correctly setting the hidden tile. But I don't know diggy well enough to know if that is the case.

> Im not very familiar with Lua, im assuming the string will only evaluate to false if it is empty ?

Empty strings are truthy in lua, however I believe hidden tile is nil (which is falsy) if there is not a hidden tile set. At least that is what the docs imply (string vs string?) https://lua-api.factorio.com/latest/classes/LuaTile.html#hidden_tile

You can easily test lua behaviour in game using `/c print("" or "empty strings are truthy")` or something like https://play.luau.org/

grilledham (approved):

map_gen/maps/diggy/feature/diggy_hole.lua
actinoide: 1. i absolutely agree, i didnt realise it was missing and since it worked i didnt think to confirm.
2. i dont know if the default tile placement behviour is overriden but for player placed tiles (wheter manual or bot) the hidden tile is set correctly. The only place i observed an artificial tile without a hidden tile was under the stone bricks generated on map creation, those are not mineable as far as im aware though so i dont think that will cause issues.

thanks once again for checking over my code and merging, could i bother you to update the code running on s13 and running "/c game.forces["enemy"].kill_all_units()" or another command that removes all enemy unit groups as we have a quite significant (over 30%) performance loss due to them ?

grilledham: `/c game.forces["enemy"].kill_all_units()` might not trigger deletion of the groups. You would probably need a command that looks for all groups and deletes them, if you find one that you think works correctly I or another admin can run that for you.

Normally the scenario updates when the map is restarted. I guess that's more problematic for s13 as it's longer running. It is possible to hot patch them, but that is more effort.

Hot patching it and running the kill all units command I don't think would fix the existing empty group issue, so really we need to find a command that will do that first.

actinoide: ive run the command in a singleplayer save of the s13 (this was a few days ago, shouldnt matter though) and just did it again (unit groups are displayed by turning on show_unit_group_info and setting force to enemy). in both cases the performance improvement (in the show_time_usage graph the "commander" item is of interest here) was consistent with all groups being removed (dropping to near 0) and the groups disapear from display.
Attached one image showing the empty groups and a second after the command was run (i did not change my F4 settings in any way betweeen these screenshots).

Screenshot 2026-04-15 211545
Screenshot 2026-04-15 211603

PR #1561, opened 2026-05-21
crash-site: Strike and Barrage Updates

misterburgers:
This PR contains several changes to strikes and barrages for crash-site. This includes some functional, visual, and limited balance changes. I've tried to separate the changes cleanly into commits for review, and I anticipate these will be squashed if committed.

This change may represent some of my own personal wishes, so I'm open to feedback and refinement as needed. I'm also happy to explain any changes here or on discord (furgers).

I have tested these changes locally directly, but not across a full playthrough.

Major changes:
- Support the strike and barrage remote deconstruction planners being used from a player's blueprint library 
- Improve targeting for the barrage command
  - Prioritize lower health nests followed by nests closest to the center
  - When nests require 2 shells to kill, the logic now prefers to kill nests rather than hit every nest once
  - Now ignores nests which will already die due to incoming damage (usually other barrage shells)
- Balance: Introduce a multiplicative 7% cost reduction for every level past the first when upgrading
  - Results in about a 30% discount at level 6

Minor changes:
- Double rate of fire for poison strike capsules
- Change poison strikes to cover a circle instead of a square 
  - This reduces the effective area, could consider increasing radius
- Change reveal logic to reveal the entire radius of a strike or barrage
- Fix issue with reveal timing of barrages 
- Fix issue with reveal timing of poison strikes
- Fix description for rocket barrage damage upgrade


grilledham (approved):
This looks really good, thanks for making the PR!

I haven't tested it manually, but the code looks good and I couldn't spot any issues. I'm happy to merge it and see what happens 😅. The worst case scenario is we revert while we fix any issues, though it's more likely we would patch them if discovered.

PR #1572, opened 2026-07-07
Add Danger Ores - City Blocks map

lex:
Danger Ores – City Blocks

## Danger Ores – City Blocks

It's always been a little sad how little use trains get on danger ores maps — belts rule everything. Maybe this map changes that.

A grid of isolated one-ore rooms (4×4 chunks; stone rooms rare) separated by paved rail corridors carrying a **ready-made double-track network**: continuous lines ring every room and meet at a **signalled roundabout on every corner**. The rails are ordinary and minable — a head start, not a constraint. Only rail infrastructure (rails, signals, train stops, power poles) and trains can be built on the corridors, and belts cannot span them: **trains are the only inter-room logistics**.

### Details
- Spawn room carries all four ores (one per quadrant); rail techs pre-researched with a starter train kit (rails, stops, signals, locomotive, wagons, fuel).
- Corridors are paved (gravel rail beds, concrete curbs/medians); anything non-rail built there is refunded with a message.
- `modules/city_blocks.lua` generates everything chunk-locally (no cross-chunk writes, generation-order independent); room ores assigned once on init (weighted, BFS-guaranteed all four ores adjacent to spawn).
- Roundabout geometry decoded from an in-game blueprint (legacy rail pieces — functional in 2.0 and already on the danger-ores allowed list).
- Registered in `map_poll.lua` (`danger_ore_normal`) and the danger ores changelog.

### Play it
`return require 'map_gen.maps.danger_ores.presets.danger_ore_city_blocks'`

### Testing
Layout harness (ore coverage/weighting/spawn guarantee) + playtested in-game through several iterations: corridors, roundabouts, signals, train routing, build restrictions.


map_gen/maps/danger_ores/modules/city_blocks.lua
```diff
@@ -0,0 +1,376 @@
...
+        if in_bounds(ni, nj) then
+            list[#list + 1] = { ni, nj }
+        end
+    end
+    return list
+end
+
+-- room ore weights: iron, copper, coal, stone -- stone rooms are deliberately rare
+local ORE_WEIGHTS = { 3, 3, 3, 1 }
+local ORE_WEIGHT_TOTAL = 10
```
RedRafe: Shouldn't theses be computed from `danger_ores/config/*_ores.lua`?

I think it would be better for reusability if this module could accept ore config as input to be reused in other presets

map_gen/maps/danger_ores/modules/city_blocks.lua
```diff
@@ -0,0 +1,376 @@
...
+    if y_wall then
+        lay_h_rails(surface, area)
+    else
+        lay_v_rails(surface, area)
+    end
+end
+
+-- === strip placement rule ==================================================
+
+local ALLOWED_ON_STRIP = {
```
RedRafe, line 350: for allowed entities, I think it would be better for mod compatibility if it was whitelisted by LuaEntity::type and not by LuaEntity::name.
Also, you could look at entity_placement_restriction module that already implements entity destroying/saving, just need to pass the validation function to it which you already have

map_gen/maps/danger_ores/config/city_blocks_ores.lua
```diff
@@ -0,0 +1,57 @@
+local b = require 'map_gen.shared.builders'
+local start_value = b.euclidean_value(0, 0.35)
+local value = b.exponential_value(0, 0.07, 1.45)
+
+-- Order is significant: it maps to room ore indices 1..4 from the city_blocks module.
+-- Heavier dominant ratio (85/15) than the vanilla sectors: each room is meant to be
+-- one resource, so hauling between rooms by train stays essential.
+return {
```
RedRafe: Adding this comment here but I think it covers last 5 maps: each maps doesnt necessarily need a new ore config, especially if it's vanilla.
Having 4x files with same vanilla ore distribution (iron>copper>coal>stone) does not carry any benefit, especially if we decide to tweak ore percentages, having 1 file with all the ore numbers is easier to maintain.

Evaluate if it's possible to remove the latest ore configs in favor of using [vanilla_ores.lua](https://github.com/Refactorio/RedMew/blob/develop/map_gen/maps/danger_ores/config/vanilla_ores.lua) (yes, stone is "missing" but stone is not really used aside from SpaceAge

RedRafe (commented):
I've left a few suggestions. Additionally, have you tested this map by saving, closing, & reloading the savefile?

map_gen/maps/danger_ores/modules/city_blocks.lua
```diff
@@ -0,0 +1,376 @@
...
+
+-- === public ================================================================
+
+function Public.register()
+    Event.on_init(function()
+        if data.generated then
+            return
+        end
+        data.generated = true
+        assign_ores()
```
RedRafe: Not sure what the usage of `data.generated` is since it's only read and assigned once here and nowhere else. `on_init` functions will only run once on_init so no need to check that, can simply pass `Event.on_init(assign_ores)`

lex:
Thanks for the review! All four points addressed in the new commit:

- **Ore config as input**: `register(config)` now derives the ore count from the passed `main_ores` (any N works), with optional `room_ore_weights`; the spawn-quadrant split and the BFS all-ores-near-spawn guarantee adapt to whatever config is used.
- **vanilla_ores**: dropped `city_blocks_ores.lua` entirely — the preset now uses `config/vanilla_ores.lua`. Rooms are iron/copper/coal and stone stays in the mixed ratios, which plays fine here. One heads-up for the "last 5 maps" part: **Tetrominoes and Voronoi can't move to a 3-ore config** — their no-two-touching-regions-share-an-ore guarantee is a graph 4-colouring, and 4 colours is the mathematical minimum (with 3, some adjacent regions must share). Consolidating their two identical 4-ore configs into one shared file would work though — happy to do that as a small follow-up PR.
- **entity_placement_restriction**: corridor rule now runs through the shared module with a keep-alive callback (ghosts/refunds/destruction handled there). The whitelist was already keyed by `LuaEntity.type` rather than name, and stays that way.
- **`data.generated`**: gone — `Event.on_init(assign_ores)` directly.

Save/close/reload tested on the local scenario (both before and after these changes) — layout, room ores and corridor rules all persist correctly; the layout lives in Global so nothing regenerates.

Also refreshed the preview image to match the current map.

RedRafe:
Thanks for applying the changes! Looks good

> **Tetrominoes and Voronoi can't move to a 3-ore config** — their no-two-touching-regions-share-an-ore guarantee is a graph 4-colouring, and 4 colours is the mathematical minimum (with 3, some adjacent regions must share). Consolidating their two identical 4-ore configs into one shared file would work though — happy to do that as a small follow-up PR.

Happy to accept follow-up PRs if you want to. I see two issues hanging:

1. Would be cool if Tetris & Voronoi could also accept ore configs as inputs to make the module moddable with other ores. If there's a minimum number of ores to be inputed I would suggest enforcing in each module the limit and actively checking on map creation that certain rules have bene followed.

2. Regarding the shared files for ores config, I see "many" vanilla ores have similar ore configs but are split each one in its own file (even older maps). This was only recently spotted. We could probably reduce the number of vanilla ore configs by collapsing presets into a few shared files (or a more dynamic module that provides ore configs) - but it needs some rework and understanding of how different maps work. I understand if this is something you dont want to invest time so we can leave it for the future

lex:
The follow-up is up as #1573 — it went a bit further than the two duplicates: a shared main-ores factory now holds the canonical tile sets and ratio mixes once, with nine configs derived from it (all byte-identical to before, one commit per config), plus the 4-ore minimum asserted at map creation in the Tetrominoes/Voronoi layouts and a test suite over all the ore configs.

map_gen/maps/danger_ores/modules/city_blocks.lua
```diff
@@ -0,0 +1,383 @@
...
+    { name = 'legacy-straight-rail', x = 7, y = 9, direction = 14 },
+    { name = 'rail-signal', x = -4.5, y = -15.5, direction = 0 },
+    { name = 'rail-signal', x = 4.5, y = -15.5, direction = 8 },
+    { name = 'rail-signal', x = -15.5, y = -4.5, direction = 4 },
+    { name = 'rail-signal', x = 15.5, y = -4.5, direction = 4 },
+    { name = 'rail-signal', x = -15.5, y = 4.5, direction = 12 },
+    { name = 'rail-signal', x = 15.5, y = 4.5, direction = 12 },
+    { name = 'rail-signal', x = -4.5, y = 15.5, direction = 0 },
+    { name = 'rail-signal', x = 4.5, y = 15.5, direction = 8 },
+}
```
RedRafe: I would replace all legacy entities to Factorio 2.0 entities `straight-rail`, `curved-rail-a`, `curved-rail-b`.
Although entities are migrated correctly from 1.1 to 2.0/2.1, there's no guarantee support will remain in place with new experimental updates and since this PR comes already with 2.0, it's better to be up-to-date

RedRafe:
Good for me, I havent tested the map but seems to be working. Passing the PR into @grilledham 's hands for an accurate revision/approval after the last comment is addressed.

Thanks @lex :D

lex:
Note: once #1573 (the ore-config factory) merges, this branch will be refreshed against develop and CI re-run. No changes expected here — City Blocks consumes `config.vanilla_ores` by require and carries no ore numbers of its own, so it picks up the factory-backed config automatically.

map_gen/maps/danger_ores/modules/city_blocks.lua
lex: Done — the roundabout is rebuilt from a native 2.0 blueprint: `curved-rail-a`/`curved-rail-b`/`half-diagonal-rail`, no legacy pieces anywhere. The 2.0 ring is larger (40x40 vs 31x31) and overhangs into the adjacent corridor chunks, so stamping is now partitioned chunk-locally (each chunk places only the pieces inside itself, so generation order can never matter) and the corridor trunk stops at ±23 where the ring's lane-end curves take over. Approach lanes still sit at center ±3, so nothing else moved. Playtested in-game: trains through all four approaches, far-from-spawn corners, and the trunk joints all behave.

RedRafe (approved):
Good to go, thanks! :)

grilledham (approved):
LGTM. Thanks for the contribution.

PR #1574, opened 2026-07-10
Make loader crafting GUI persistent to avoid multiplayer latency delay

lex:
## Backstory

Playing on a slow and unreliable internet connection, we noticed the loader panel next to the crafting menu took seconds to appear, shifting the crafting menu over whenever it finally loaded.

## Root cause

The vanilla crafting menu is latency-hidden, so the client renders it immediately. The loader frame however was created inside `on_gui_opened` and destroyed in `on_gui_closed` — and script events only run once the open action has made the round trip through the server. On a bad connection that round trip is seconds, so the frame popped in late and re-laid-out the crafting menu when it arrived.

## Fix

- The controller-anchored frame is now created once per player (on create, join, force change and research finished) and left in place. Anchored relative frames show and hide with their target GUI, so it now appears in the same frame as the crafting menu with no server round trip. The `on_gui_opened` redraw is kept as a self-healing refresh.
- The per-entity assembling machine frame keeps its create/destroy cycle and gets its own element name so the two variants no longer collide on lookup.
- New `on_register`/`on_unregister` hooks in the QOL feature framework so runtime toggling via `set_loaders` draws and removes the persistent frames instead of leaking them; `on_configuration_changed` cleanup covers both frames.
- Fixed a pre-existing nil dereference in `draw_loader_frame` (`player.force` was read before the nil check) and nil-guarded the recipe lookups in the click handlers.

## Tests

Adds `features/redmew_qol_tests.lua` for the in-game test runner: frame drawn/removed on feature toggle, frame persists after the crafting menu closes, frame removed when no loader recipe is enabled, one button per enabled recipe, and the machine frame staying separate and destroyed on close. All five pass in-game. Note: run the suite with the redmew-data mod disabled — the loaders feature intentionally goes inert when that mod provides its own loaders.

### DIFF
--- a/features/redmew_qol.lua
+++ b/features/redmew_qol.lua
@@ -34,8 +34,13 @@ local function pick_name()
 end
 
 local loader_frame_name = Gui.uid_name()
+local loader_machine_frame_name = Gui.uid_name()
 local loader_button_player = Gui.uid_name()
 local loader_button_machine = Gui.uid_name()
+
+-- Exposed for tests
+Public._loader_frame_name = loader_frame_name
+Public._loader_machine_frame_name = loader_machine_frame_name
 
 local loaders = {
     ['loader'] = true,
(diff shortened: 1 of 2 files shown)


grilledham (approved):
Makes sense. Decent minor qol improvement.

PR #1575, opened 2026-07-10
Fix debug test suite for Factorio 2.0

lex:
Fixes two long-standing failures in the in-game test runner (`_DEBUG` + `/test-runner`), both fallout from the Factorio 2.0 engine update.

## Gui toggle test: `13 ~= 14 - after close count should be equal to before count`

The top-button toggle test counted every child of the gui roots (top/left/center/screen). GUIs like the production hud hide their screen frame on close to preserve its dragged position instead of destroying it, so the count stayed one high after closing. The test now counts only visible elements, which keeps the "open shows something / close removes it" contract while allowing the hide-on-close pattern.

## `helper.lua: bad argument #1 to 'rawset' (table expected, got userdata)`

`Helper.modify_lua_object` monkey-patched fields directly onto LuaObjects (`player.print`, `game.get_player`, ...) with `rawset`. That worked in Factorio 1.1 where LuaObjects were Lua tables, but they are userdata since 2.0, so every test using it errored. This PR adds:

- `Helper.fake_lua_object(object, overrides)` — a proxy table that reads and writes through to the wrapped LuaObject except for the overridden fields.
- `Helper.modify_global(context, name, value)` — replaces a global (e.g. `game`) with such a proxy until test teardown.

The death corpse tags, restart command and landfill remover tests are migrated to the new helpers; `modify_lua_object` now raises a clear error when given a LuaObject. This is safe for these tests because the features under test only read through the player object (`print`, `admin`, `force`, ...) and never pass it back into engine APIs.

Verified in-game on a vanilla-mod map: the Gui toggle tests and death corpse tags test now pass. The restart command and landfill remover suites are migrated mechanically but were not exercised, as those features are not loaded on the default map.

### DIFF
--- a/features/death_corpse_tags_tests.lua
+++ b/features/death_corpse_tags_tests.lua
@@ -69,13 +69,17 @@ Declare.module({'features', 'death_corpse_tags'}, function()
 
         local actual_text
 
-        Helper.modify_lua_object(context, player, 'print', function(text)
-            actual_text = text
-        end)
+        local fake_player = Helper.fake_lua_object(player, {
+            print = function(text)
+                actual_text = text
+            end
+        })
 
-        Helper.modify_lua_object(context, game, 'get_player', function()
-            return player
-        end)
+        Helper.modify_global(context, 'game', Helper.fake_lua_object(game, {
+            get_player = function()
+                return fake_player
+            end
+        }))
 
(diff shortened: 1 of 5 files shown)


grilledham (approved):
Thanks for fixing the tests.

PR #1576, opened 2026-07-10
Show pass counts in test runner module rows

lex:
Module rows in the in-game test viewer showed only the total test count, so a module with a single failure read the same as one that failed entirely — e.g. a red `Gui (17)` where 12 of the 17 actually passed.

After a run, module rows now show `passed/total`, e.g. `Gui (12/17)`. Modules with no results yet keep the plain total, so an unrun module doesn't misleadingly read as `0/17`.

Implementation: `mark_module_for_passed` in the runner already walks the module tree after every run to set the pass/fail colouring, so it now also tallies passed and ran tests per module in the same pass. The viewer uses the tally when present.

## Before

Screenshot 2026-07-10 at 23 07 49

## After

Screenshot 2026-07-10 at 23 08 27

### DIFF
--- a/utils/test/module_store.lua
+++ b/utils/test/module_store.lua
@@ -19,6 +19,7 @@ local function new_module(module_name)
         is_open = false,
         depth = nil,
         count = nil,
+        pass_count = nil,
         passed = nil
     }
 end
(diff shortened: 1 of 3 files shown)


grilledham (approved):
Nice improvment.

PR #1577, opened 2026-07-11
Reduce ore density on the maze danger ore maps

lex:
Playtest tuning for the maze danger ore variants. The core problem: the ore field is so rich that groups never need to push more than a few corridors out — the maze layout ends up having no say in the gameplay. Lower density means more mining outward, more corridors explored, and the maze actually shaping the run.

Player feedback from the current maps:

> Some feedback about the new scrap maze. Giant lakes are hard to fill with the tiny bits of stone we get.

> on coal maze the patches could be a little bit less dense ... some of them take really long to mine

> I feel lower the better, we have unlimited amount, and can always keep going for more dense

## Changes

- **Scrapworld Maze / Omnimatter Maze**: main ore density scaled to 25%.
- **Scrapworld Maze**: occasional dense pure-stone patches added, so groups can set up dedicated stone mining for landfill. Scrap itself only yields stone at a 5% chance per mined unit, which also has to cover rails and bricks — hence the lake complaint above. (These can go away again if the scrap resource ever gets customizable yield ratios.)
- **Coal Maze**: dense iron/copper/stone patches halved in richness and doubled in size — fewer, chunkier patches that don't take forever to mine out; the coal background is unchanged.

## Implementation

First commit extends the fluent main-ores factory so the presets can express all of this relative to their shared configs instead of restating curve numbers:

- `:scale_richness(multiplier[, ore_name])` multiplies the start and richness curves of a main-ores config.
- `Factory.patches()` — a fluent builder for resource patches configs with the same contract as the main-ores API: `add_patch` appends a pure-resource perlin patch, `:scale{richness=, size=}` adjusts existing configs (wrapping shared plain configs without modifying them).
- The Omnimatter ore config moves to the factory (it duplicated the scrap config).

Covered by new cases in the standalone `ore_configs.test.lua` harness (all passing), and verified in-game on new maps for all three presets.


grilledham (approved):
LGTM.

